PR #6, opened 2019-04-02
Allow to run multiple replicas

SchSeba:
Copy the manager code from controller runtime and change it.
With the new manager the controller will became ready only when it became leader.

most of the code is the same as the regular manager I just change the leader election function.
When a controller take the leader he also create a file under `/tmp/healthy` this way the pod will became ready.
When the pod became ready it will be added as an endpoint for the admission webhook service

### DIFF
--- a/pkg/manager/internal.go
+++ b/pkg/manager/internal.go
@@ -1,0 +1,321 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+ Same file as vendor/sigs.k8s.io/controller-runtime/pkg/manager/internal.go
+ The change is on the startLeaderElection function.
+ when we get the election lock we create a file.
+ This file is watch by the kubernetes readinessProbe.
+ When the pod because ready it will be added to the service for the webhook.
+ This way we have only one mutating webhook active.
+*/
+
+package manager
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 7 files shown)


phoracek:
@SchSeba please elaborate why is this better and why we want to have it. It adds a lot of code, so I'd like to have a good reason.

Also, could you explicitly say which part of the code has been copied and what are our changes? I wonder whether the regular manager supports extensions, so we would not need to duplicate the whole thing.

SchSeba:
the change is on the `startLeaderElection` function

this code:
https://github.com/K8sNetworkPlumbingWG/kubemacpool/pull/6/files#diff-77faf6b20512574869434402d5c5b6a2R284

original code:
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/internal.go#L297

The manager doesn't support for any extension on the leader election callback functions.

The reason I make this change is because the leader election is to make sure only one controller is up **but** the webhooks are **always** up unrelated to the leader.

With our implementation of the macpool we use the allocation mac in memory and only init the maps when the controller become a leader.

So with my change the non leader pod will not be in ready state and not be part of the webhook service endpoints.

pkg/manager/internal.go
```diff
@@ -0,0 +1,320 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+ Same file as vendor/sigs.k8s.io/controller-runtime/pkg/manager
```
phoracek: manager is not a file, shouldn't this be manager/internal.go?

pkg/manager/internal.go
```diff
@@ -0,0 +1,320 @@
...
+func (cm *controllerManager) startLeaderElection() (err error) {
+	l, err := leaderelection.NewLeaderElector(leaderelection.LeaderElectionConfig{
+		Lock: cm.resourceLock,
+		// Values taken from: https://github.com/kubernetes/apiserver/blob/master/pkg/apis/config/v1alpha1/defaults.go
+		// TODO(joelspeed): These timings should be configurable
+		LeaseDuration: 15 * time.Second,
+		RenewDeadline: 10 * time.Second,
+		RetryPeriod:   2 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(_ context.Context) {
```
phoracek: Please mark the part that is changed compared to the original with a comment

phoracek:
So we end up with a single webhook? Sounds limiting, but I guess we don't have other choice at this point.

pkg/manager/internal.go
SchSeba: done

pkg/manager/internal.go
SchSeba: done

PR #8, opened 2019-04-03
Add owner reference for the mutating webhook

SchSeba:
This way when we remove the statefulset the webhook will be removed as well.

This PR also change the name of the hook from `mutating-webhook-configuration` to `kubemacpool`

This should fix https://github.com/kubevirt/cluster-network-addons-operator/issues/33

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -111,6 +111,11 @@ func main() {
 	startPoolRange := loadMacAddressFromEnvVar(poolmanager.StartPoolRangeEnv)
 	endPoolRange := loadMacAddressFromEnvVar(poolmanager.EndPoolRangeEnv)
 
+	// create a owner ref on the mutating webhook
+	// this way when we remove the statefulset of the manager the webhook will be also removed from the cluster
+	err = webhook.CreateOwnerRefForMutatingWebhook(clientset)
+	ExitIfError(err, "unable to create owner reference for mutating webhook object")
+
 	isKubevirtInstalled := checkForKubevirt(clientset)
 	poolManager, err := poolmanager.NewPoolManager(clientset, startPoolRange, endPoolRange, isKubevirtInstalled)
 	ExitIfError(err, "unable to create pool manager")
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -17,13 +17,19 @@ limitations under the License.
 package webhook
 
 import (
+	"fmt"
+
+	admissionregistration "k8s.io/api/admissionregistration/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apitypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	runtimewebhook "sigs.k8s.io/controller-runtime/pkg/webhook"
-
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/K8sNetworkPlumbingWG/kubemacpool/pkg/pool-manager"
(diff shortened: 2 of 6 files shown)


cmd/manager/main.go
```diff
@@ -111,6 +111,9 @@ func main() {
 	startPoolRange := loadMacAddressFromEnvVar(poolmanager.StartPoolRangeEnv)
 	endPoolRange := loadMacAddressFromEnvVar(poolmanager.EndPoolRangeEnv)
 
+	err = webhook.CreateOwnerRefForMutatingWebhook(clientset)
```
phoracek: please add inline documentation saying why we need that

pkg/webhook/webhook.go
```diff
@@ -74,6 +82,35 @@ func AddToManager(mgr manager.Manager, poolManager *pool_manager.PoolManager) er
 	if err != nil {
 		return err
 	}
-
 	return nil
 }
+
+func CreateOwnerRefForMutatingWebhook(kubeClient *kubernetes.Clientset) error {
+	managerStatefulSet, err := kubeClient.AppsV1().StatefulSets("kubemacpool-system").Get("kubemacpool-mac-controller-manager", metav1.GetOptions{})
```
phoracek: Isn't it possible to set the reference while the MutatingWebhookConfiguration is being created? Or is the code hidden somewhere?

pkg/webhook/webhook.go
```diff
@@ -74,6 +82,35 @@ func AddToManager(mgr manager.Manager, poolManager *pool_manager.PoolManager) er
...
+	if err != nil {
+		return err
+	}
+	ownerRefList := []metav1.OwnerReference{{Name: managerStatefulSet.Name, Kind: "StatefulSet", APIVersion: "apps/v1", UID: managerStatefulSet.UID}}
+
+	mutatingWebHookObject, err := kubeClient.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get("kubemacpool", metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
```
phoracek: why do we create it ourselves? it is already being created, isn't it? this looks like a duplicate.

SchSeba:
The problem is the service is not created until the manager is not running and I want this to run before the manager start.

`manager.start()` is a block command

pkg/webhook/webhook.go
SchSeba: see the comment below

phoracek:
Wait, we create the service dynamically too?

SchSeba:
yes this function creates both the service and secret dynamically

https://github.com/K8sNetworkPlumbingWG/kubemacpool/blob/master/pkg/webhook/webhook.go#L42

pkg/webhook/webhook.go
phoracek: Referencing Service was one thing. In this comment I was talking about the fact, that MutatingWebhookConfiguration is dynamically created by kubemacpool even without this PR (am I right?). I don't see you removing that code in this PR, so we have a duplicate.

I wonder whether we could edit the code that creates mentioned MutatingWebhookConfiguration, so it would set reference before creating.

cmd/manager/main.go
SchSeba: done

phoracek:
@SchSeba I wonder whether we can set it up in https://github.com/K8sNetworkPlumbingWG/kubemacpool/blob/master/pkg/webhook/webhook.go#L42. I don't think so :(

SchSeba:
@phoracek no there is no way to extend it :(

pkg/webhook/webhook.go
SchSeba: the only way to do it is move the code from the vendor folder and make it our code but I don't like this solution (bad for upgrade the packages).

There is no problem on creating the webhook first is the same as if the manager restart itself.
The code check the related spec variables he need but I doesn't change the `ObjectMeta`

PR #9, opened 2019-04-05
Fix multiple issues

SchSeba:
This PR fix multiple issues found in the manager.

* add more debug logs
* fix vmi object
* remove the isLeader function on every request
* fix pod allocation problem of pod the are part of statefulsets deplyment and vms


SchSeba:
@phoracek I think is ready for review.

this first commit from taken from here https://github.com/K8sNetworkPlumbingWG/kubemacpool/pull/10

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -143,6 +144,68 @@ func (p *PoolManager) allocateRequestedVirtualMachineInterfaceMac(requestedMac s
 	return nil
 }
 
+func (p *PoolManager) initVirtualMachineMap() error {
```
phoracek: I don't see this function called. It should be used in the same commit you introduce it. You can merge it with the following commit.

pkg/pool-manager/pool.go
```diff
@@ -99,19 +99,6 @@ func (p *PoolManager) getFreeMac() (net.HardwareAddr, error) {
 	return nil, fmt.Errorf("the range is full")
 }
 
-func (p *PoolManager) setAsLeader() {
```
phoracek: should be removed in the same commit you remove calls to this function

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -219,13 +219,24 @@ func (p *PoolManager) IsKubevirtEnabled() bool {
...
+func (p *PoolManager) isRelatedToKubevirt(pod *corev1.Pod) bool {
 	if pod.ObjectMeta.OwnerReferences == nil {
 		return false
 	}
 
 	for _, ref := range pod.OwnerReferences {
 		if ref.Kind == kubevirt.VirtualMachineInstanceGroupVersionKind.Kind {
+			requestUrl := fmt.Sprintf("apis/kubevirt.io/v1alpha3/namespaces/%s/virtualmachines/%s", pod.Namespace, ref.Name)
+			log.V(1).Info("test", "requestURI", requestUrl)
+			result := p.kubeClient.ExtensionsV1beta1().RESTClient().Get().RequestURI(requestUrl).Do()
```
phoracek: why do you use RESTClient instead of kubeClient VirtualMachines?

pkg/pool-manager/pool.go
SchSeba: the `setAsLeader` was removed In a different PR I forgot to remove the function there

pkg/pool-manager/virtualmachine_pool.go
SchSeba: because to import the kubecli from kubevirt I need to add multiple packages like `gorilla/websocket` and others and I don't want to do it just to check if the object exist.

I import only the api package from kubevirt.

phoracek: Sounds good

pkg/pool-manager/virtualmachine_pool.go
SchSeba: Done

pkg/pool-manager/pod_pool.go
```diff
@@ -152,6 +153,69 @@ func (p *PoolManager) allocatePodFromPool(pod *corev1.Pod) (string, error) {
 	return macAddr.String(), nil
 }
 
+func (p *PoolManager) initPodMap() error {
+	log.V(1).Info("start InitMaps to reserve existing mac addresses before allocation new ones")
+	pods, err := p.kubeClient.CoreV1().Pods("").List(metav1.ListOptions{})
```
phoracek: AFAIK this will iterate only over pods in the working namespace, not all namespaces

pkg/pool-manager/pod_pool.go
```diff
@@ -152,6 +153,69 @@ func (p *PoolManager) allocatePodFromPool(pod *corev1.Pod) (string, error) {
...
+					"mac", network.MacRequest)
+				continue
+			}
+		}
+	}
+
+	return nil
+}
+
+func (p *PoolManager) allocatedToCurrentPod(podname string, network *multus.NetworkSelectionElement) bool {
```
phoracek: I don't see this used anywhere

pkg/pool-manager/pod_pool.go
```diff
@@ -123,7 +124,7 @@ func (p *PoolManager) allocatePodRequestedMac(requestedMac string, pod *corev1.P
 		return err
 	}
 
-	if _, exist := p.macPoolMap[requestedMac]; exist {
+	if isAllocatedToPod, exist := p.macPoolMap[requestedMac]; exist && isAllocatedToPod && !p.allocatedToCurrentPod(podNamespaced(pod), network) {
```
phoracek: why is `isAllocatedToPod` needed? `macPoolMap` is a set with `true` values. I haven't noticed a code assigning false to it. And if there is any, please update documentation of this attribute in its structure

pkg/controller/pod/pod_controller.go
```diff
@@ -89,5 +89,15 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 		return reconcile.Result{}, err
 	}
 
+	// allocate only when the pod is ready
```
phoracek: how is this change related?

pkg/pool-manager/pool.go
```diff
@@ -26,24 +26,25 @@ import (
...
-	vmToMacPoolMap  map[string][]string  // cap for namespace/vmname and a list of allocated mac addresses
-	poolMutex       sync.Mutex           // mutex for allocation an release
-	isLeader        bool                 // leader boolean
-	isKubevirt      bool                 // bool if kubevirt virtualmachine crd exist in the cluster
+	kubeClient      kubernetes.Interface         // kubernetes client
+	startRange      net.HardwareAddr             // fist mac in range
+	endRange        net.HardwareAddr             // last mac in range
+	currentMac      net.HardwareAddr             // last given mac
+	macPoolMap      map[string]bool              // allocated mac map
+	podToMacPoolMap map[string]map[string]string // map for namespace/podname and a mac of network name as key and mac address as value
```
phoracek: i don't understand the comment.

maybe 'map allocated mac address by networkname and namespace/podname: {"namespace/podname: {"network name": "mac address"}}'

pkg/pool-manager/pod_pool.go
```diff
@@ -42,6 +42,11 @@ func (p *PoolManager) AllocatePodMac(pod *corev1.Pod) error {
 		return nil
 	}
 
+	// allocate only when the network status is no exist
```
phoracek: could you explain why you do it?

"if the network status annotation is already set, it means *???*"

pkg/pool-manager/pod_pool.go
```diff
@@ -131,11 +137,18 @@ func (p *PoolManager) allocatePodRequestedMac(requestedMac string, pod *corev1.P
 		return err
 	}
 
+	if pod.Name == "" {
+		// we are going to create the podToMacPoolMap in the controller call
+		// because we don't have the pod name in the webhook
+		p.macPoolMap[requestedMac] = false
```
phoracek: why don't you drop it instead? if you are using that true/false as a communication channel between webhook and controller, the attribubute needs to be renamed and properly documented. for me it's really hard to understand this right now

pkg/pool-manager/pod_pool.go
SchSeba: I test it now and I get all the pod in all namespaces with the `Pods("")`

phoracek: Oh, then it's good! Thanks

pkg/pool-manager/pod_pool.go
SchSeba: in the if statement `if isAllocatedToPod, exist := p.macPoolMap[requestedMac]; exist && isAllocatedToPod && !p.allocatedToCurrentPod(podNamespaced(pod), network)`

phoracek: Should be in the commit that uses it, but let's leave it this time

pkg/pool-manager/pod_pool.go
SchSeba: the pod namespace is already exist because multus runs and create the network status label in the pod

pkg/controller/pod/pod_controller.go
SchSeba: we need this change.
run the allocation after the pod is created (have a name and not only a generateName object) this will update the value on the mac address map to true and create the pod entry to in `podToMacPoolMap`

pkg/pool-manager/pod_pool.go
SchSeba: added a properly documentation about the variable.

`allocated mac map if the value is false the mac was allocated but the pod is not running yet if it is true there is a pod running. for vm is always true`

This append when the user create a pod with the generateName and not with name. we get the wehook without the pod name and then the kube-controller will create the pod with a name this is why we need also need the allocatemac in the reconcile function for pods

pkg/pool-manager/pool.go
SchSeba: done

pkg/pool-manager/pod_pool.go
SchSeba: done

pkg/pool-manager/pod_pool.go
```diff
@@ -119,23 +127,34 @@ func (p *PoolManager) ReleasePodMac(podName string) error {
...
 
-func (p *PoolManager) allocatePodRequestedMac(requestedMac string, pod *corev1.Pod) error {
+func (p *PoolManager) allocatePodRequestedMac(network *multus.NetworkSelectionElement, pod *corev1.Pod) error {
+	requestedMac := network.MacRequest
 	if _, err := net.ParseMAC(requestedMac); err != nil {
 		return err
 	}
 
-	if isAllocatedToPod, exist := p.macPoolMap[requestedMac]; exist && isAllocatedToPod && !p.allocatedToCurrentPod(podNamespaced(pod), network) {
+	if isAllocatedToPod, exist := p.macPoolMap[requestedMac]; exist &&
```
phoracek: s/isAllocatedToPod/macAllocationStatus/

pkg/pool-manager/pool.go
```diff
@@ -26,26 +26,34 @@ import (
...
+	poolMutex       sync.Mutex                   // mutex for allocation an release
+	isLeader        bool                         // leader boolean
+	isKubevirt      bool                         // bool if kubevirt virtualmachine crd exist in the cluster
 }
 
+type AllocationStatus string
+
+const (
+	AllocationStatusAllocated     AllocationStatus = "Allocated"
+	AllocationStatusWaitingForPod AllocationStatus = "WaitingForPod"
```
phoracek: I see it being assigned, but never changed to "Allocated". Shouldn't we set it to "Allocated" somewhere?

phoracek:
Just a small remark, you can leave it if you want.

Next time please let's try to do one logical change in a single PR so it is easier to review and we can do faster iterations. Also moving code from one commit to another (if there are more commits) will be simpler.

pkg/pool-manager/pool.go
SchSeba: https://github.com/K8sNetworkPlumbingWG/kubemacpool/pull/9/files/c219fd40a10bbfb5c0c0810e721e3e1981a8cc6b..789c062254a1fccc8b8e6211e12b642f26eab7e3#diff-f178743f06898ca1ae8f344f45065660R179

here is the change

phoracek: I see, all good then

pkg/pool-manager/pod_pool.go
SchSeba: done

SchSeba:
> Next time please let's try to do one logical change in a single PR so it is easier to review and we can do faster iterations. Also moving code from one commit to another (if there are more commits) will be simpler.

Thanks for the comment I will try to implement it on my next PRs

phoracek:
LGTM

PR #11, opened 2019-04-11
Update resource limit and request

SchSeba:
For large cluster the kubemacpool gets kill by OOM.

### DIFF
--- a/config/default/manager/manager.yaml
+++ b/config/default/manager/manager.yaml
@@ -67,11 +67,11 @@ spec:
                 key: END_POOL_RANGE
         resources:
           limits:
-            cpu: 100m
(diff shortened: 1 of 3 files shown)


config/default/manager/manager.yaml
```diff
@@ -67,11 +67,11 @@ spec:
                 key: END_POOL_RANGE
         resources:
           limits:
-            cpu: 100m
-            memory: 30Mi
+            cpu: 800m
```
phoracek: do we need to raise CPU? the problem is in memory

SchSeba: yes I also saw a CPU issue when we run the initMap function on a large cluster

PR #13, opened 2019-04-14
Fix init function for pod to skip kubevirt vm objects

SchSeba:
This PR fix an issue when initializing the mac address maps.

The problem was that when we init the pod mac we didn't check if the pod is related to kubevirt virtual machine object.
If it is we skip it and let the vm initialization function to handle it.

### DIFF
--- a/pkg/pool-manager/pod_pool.go
+++ b/pkg/pool-manager/pod_pool.go
@@ -210,7 +210,18 @@ func (p *PoolManager) initPodMap() error {
 		if err != nil {
 			continue
 		}
+
 		log.V(1).Info("pod meta data", "podMetaData", pod.ObjectMeta)
+		if len(networks) == 0 {
+			continue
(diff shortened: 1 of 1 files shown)


SchSeba:
Fixed: https://github.com/K8sNetworkPlumbingWG/kubemacpool/issues/14

pkg/pool-manager/pod_pool.go
```diff
@@ -210,7 +210,18 @@ func (p *PoolManager) initPodMap() error {
 		if err != nil {
 			continue
 		}
+
```
phoracek: commit message:

This PR **fixes an** issue.

I don't understand the second paragraph. Please explain what was the problem and how did you fix it.

SchSeba: done can you take a look please

PR #19, opened 2019-04-15
Revert mac allocation

SchSeba:
This PR allow to revert allocations if there was a problem allocating one of the interfaces.

The allocation could fail if there is not enough free mac addresses in the pool for all the interfaces, if the user request a mac that is already allocated or invalid.

### DIFF
--- a/pkg/pool-manager/virtualmachine_pool.go
+++ b/pkg/pool-manager/virtualmachine_pool.go
@@ -56,24 +56,32 @@ func (p *PoolManager) AllocateVirtualMachineMac(virtualMachine *kubevirt.Virtual
 		"namespace", virtualMachine.Namespace,
 		"interfaces", virtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces)
 
-	for idx, iface := range virtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces {
+	copyVM := virtualMachine.DeepCopy()
+	allocations := []string{}
+	for idx, iface := range copyVM.Spec.Template.Spec.Domain.Devices.Interfaces {
 		if iface.Masquerade == nil && iface.Slirp == nil && networks[iface.Name].Multus == nil {
 			log.Info("mac address can be set only for interface of type masquerade and slirp on the pod network")
 			continue
 		}
 
 		if iface.MacAddress != "" {
-			if err := p.allocateRequestedVirtualMachineInterfaceMac(iface.MacAddress, virtualMachine); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -56,24 +56,32 @@ func (p *PoolManager) AllocateVirtualMachineMac(virtualMachine *kubevirt.Virtual
 		"namespace", virtualMachine.Namespace,
 		"interfaces", virtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces)
 
-	for idx, iface := range virtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces {
+	copyVM := virtualMachine.DeepCopy()
```
phoracek: Please comment in the commit message when the allocation could fail

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -56,24 +56,32 @@ func (p *PoolManager) AllocateVirtualMachineMac(virtualMachine *kubevirt.Virtual
...
+				p.revertAllocationOnVm(vmNamespaced(copyVM), allocations)
 				return err
 			}
-			virtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces[idx].MacAddress = macAddr
+			copyVM.Spec.Template.Spec.Domain.Devices.Interfaces[idx].MacAddress = macAddr
+			allocations = append(allocations, macAddr)
 		}
 	}
+
+	virtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces = copyVM.Spec.Template.Spec.Domain.Devices.Interfaces
```
phoracek: I'm little tired, could you please explain to me why do we need the copy?

pkg/pool-manager/virtualmachine_pool.go
SchSeba: done

pkg/pool-manager/virtualmachine_pool.go
SchSeba: because we add the mac addresses allocated into the copy object so if we need to revert it will be easy (we don't change the original object)

if we get to this line the allocation succeeded so we copy the interfaces from the copy object (the one with the macAddress field configured and validated by the manager) to the original one.

SchSeba:
@phoracek I also rebase base this PR

pkg/pool-manager/virtualmachine_pool.go
phoracek: why don't we do the same with pods?

SchSeba: in pods we marshal the networks again into the annotation :)

PR #28, opened 2019-04-28
Add namespace to vm object if not exist

SchSeba:
If the vm object doesn't have a namespace variable we take the namespace from the request and add it to the vm object.


SchSeba:
@booxter @mmirecki Can you review this

SchSeba:
Hi @booxter thanks for the comment.

About the commit message I save the number for the first line of the comment then I leave a empty line and then a more detailed message

PR #32, opened 2019-05-07
Fix pod annotation clear

SchSeba:
This PR fix an issue with requested networks.

Before this PR networks with requested mac where not added into the generated annotation.


phoracek:
LGTM

PR #33, opened 2019-06-02
Bump kubevirt to version v0.17.0

SchSeba:
update unit test to use MultusNetwork insted of CniNetwork struct


SchSeba:
@phoracek we don't break here because the yaml struct didn't change only the struct name.
we can continue with the v0.15.0 without any issue and it will work with kubevirt version v0.17.0

PR #34, opened 2019-06-03
Fix multiple yaml apply

SchSeba:
Add virtual machine update webhook
    
Add update handler for virtual machines objects.

This PR fix an issue when the user try to apply the same vm again.

Before this PR if the issue apply the yaml file again the mac addresses allocated by the kubemacpool will disapear from the vm object.

We support now to apply the same yaml again and all the allocated addresses will be reserved but also add and remove interfaces.

### DIFF
--- a/pkg/pool-manager/virtualmachine_pool.go
+++ b/pkg/pool-manager/virtualmachine_pool.go
@@ -117,6 +117,88 @@ func (p *PoolManager) ReleaseVirtualMachineMac(virtualMachineName string) error 
 	return nil
 }
 
+func (p *PoolManager) UpdateMacAddressesForVirtualMachine(virtualMachine *kubevirt.VirtualMachine) error {
+	p.poolMutex.Lock()
+
+	log.V(1).Info("UpdateMacAddressesForVirtualMachine: data",
+		"macmap", p.macPoolMap,
+		"podmap", p.podToMacPoolMap,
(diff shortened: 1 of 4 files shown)


pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -117,25 +117,108 @@ func (p *PoolManager) ReleaseVirtualMachineMac(virtualMachineName string) error
...
+
+	log.V(1).Info("UpdateMacAddressesForVirtualMachine: data",
+		"macmap", p.macPoolMap,
+		"podmap", p.podToMacPoolMap,
+		"vmmap", p.vmToMacPoolMap,
+		"currentMac", p.currentMac.String())
+
+	existInterfacesMap, isVirtualMachineExist := p.vmToMacPoolMap[vmNamespaced(virtualMachine)]
+	if !isVirtualMachineExist {
+		p.poolMutex.Unlock()
```
qinqon: We can remove this unlock moving the defer before the if statement.

SchSeba: I think it will be a deadlock because the `p.AllocateVirtualMachineMac(virtualMachine)` try to lock the mutex.

correct me but the defer calls after the function finish no if I call another function right?

qinqon: If this is the case then this is alright.

Issue #37, opened 2019-06-26
After applying invalid yaml mac addresses are not released


nikita133:
Once you apply invalid yaml mac addresses are not released.
Scenario:
Step 1: Apply invalid yaml:
**oc apply -f temp_a**
The  "" is invalid: spec.template.spec.networks: every network must be mapped to an interface

Step 2: Check logs
oc logs kubemacpool-mac-controller-manager-844cd8fbf8-rw6vk -n kubemacpool-system
{"level":"info","ts":1561535723.8990824,"logger":"PoolManager","msg":"mac from pool was allocated for virtual machine","allocatedMac":"02:aa:bf:00:00:01","virtualMachineName":"vma","virtualMachineNamespace":"myproject"}
{"level":"info","ts":1561535723.8990977,"logger":"PoolManager","msg":"mac from pool was allocated for virtual machine","allocatedMac":"02:aa:bf:00:00:02","virtualMachineName":"vma","virtualMachineNamespace":"myproject"}
{"level":"info","ts":1561535723.8991125,"logger":"PoolManager","msg":"mac from pool was allocated for virtual machine","allocatedMac":"02:aa:bf:00:00:03","virtualMachineName":"vma","virtualMachineNamespace":"myproject"}

Step 3: Reapply invalid yaml
**oc apply -f temp_a**

Step 4: Check logs:
oc logs kubemacpool-mac-controller-manager-844cd8fbf8-rw6vk -n kubemacpool-system
{"level":"info","ts":1561535798.2916842,"logger":"PoolManager","msg":"mac from pool was allocated for virtual machine","allocatedMac":"02:aa:bf:00:00:04","virtualMachineName":"vma","virtualMachineNamespace":"myproject"}
{"level":"info","ts":1561535798.2917054,"logger":"PoolManager","msg":"mac from pool was allocated for virtual machine","allocatedMac":"02:aa:bf:00:00:05","virtualMachineName":"vma","virtualMachineNamespace":"myproject"}
{"level":"info","ts":1561535798.2917206,"logger":"PoolManager","msg":"mac from pool was allocated for virtual machine","allocatedMac":"02:aa:bf:00:00:06","virtualMachineName":"vma","virtualMachineNamespace":"myproject"}
[...]

PR #40, opened 2019-07-01
Add a cleanup loop for failed vm

SchSeba:
Before this PR we have an issue if the vm creation fails because there
is an validation error from the virt-api we didn't clean the allocation.

This PR introduce a cleanup loop into the system.
When we hit the create vm mutating webhook we create the regular
allocation but we also configure the vm in a wait map.
We have a waiting look the will check if the object is removed from the
map and if is not after 3-6 second we assume the object wasn't saved
into the etcd (no controller event) so we release the object.

### DIFF
--- a/pkg/controller/virtualmachine/virtualmachine_controller.go
+++ b/pkg/controller/virtualmachine/virtualmachine_controller.go
@@ -96,5 +96,6 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 		return reconcile.Result{}, err
 	}
 
+	r.poolManager.MarkVMAsReady(fmt.Sprintf("%s/%s", request.Namespace, request.Name))
 	return reconcile.Result{}, nil
 }
--- a/pkg/pool-manager/virtualmachine_pool.go
+++ b/pkg/pool-manager/virtualmachine_pool.go
@@ -18,10 +18,11 @@ package pool_manager
 
 import (
 	"fmt"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"net"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	kubevirt "kubevirt.io/kubevirt/pkg/api/v1"
 )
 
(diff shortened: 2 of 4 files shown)


SchSeba:
should fix #37

SchSeba:
we should use a persistent storage for this because the kubemacpool can stop after the mutating webhook but before the controller ack on the request.

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -343,11 +347,53 @@ func (p *PoolManager) releaseMacAddressesFromInterfaceMap(allocations map[string
...
+// This function remove the virtual machine from the waiting list
+// this mean that we got an event about a virtual machine in vm controller loop
+func (p *PoolManager) MarkVMAsReady(vmName string) {
+	p.vmMutex.Lock()
+	defer p.vmMutex.Unlock()
+
+	delete(p.vmCreationWaiting, vmName)
+}
+
+// This function check if there are virtual machines that hits the create
```
phoracek: Nice documentation :+1: !

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -96,5 +96,6 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 		return reconcile.Result{}, err
 	}
 
+	r.poolManager.MarkVMAsReady(fmt.Sprintf("%s/%s", request.Namespace, request.Name))
```
phoracek: maybe it would be nicer to pass namespace and name as arguments and join them inside the function

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -82,6 +83,7 @@ func (p *PoolManager) AllocateVirtualMachineMac(virtualMachine *kubevirt.Virtual
 	}
 
 	virtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces = copyVM.Spec.Template.Spec.Domain.Devices.Interfaces
+	p.vmCreationWaiting[vmNamespaced(virtualMachine)] = 0
```
phoracek: maybe map[string]bool would look better, so you do p.vmCreationWaiting[...] = true, makes more sense IMO. But even better would be to have a helper function

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -113,6 +115,7 @@ func (p *PoolManager) ReleaseVirtualMachineMac(virtualMachineName string) error
 	}
 
 	delete(p.vmToMacPoolMap, virtualMachineName)
+	delete(p.vmCreationWaiting, virtualMachineName)
```
phoracek: can't you use MarkVMAsReady function?

SchSeba:
replaced by https://github.com/K8sNetworkPlumbingWG/kubemacpool/pull/59

PR #30, opened 2019-04-30
Func tests

alonSadan:
functional tests for the macpool:(currently not working)
1. test that a mac address is automatically assigned to a new created VM
2.test that when a client treis to assign the same mac address for two VM, an error will be sent.
3...

PR #41, opened 2019-07-01
func test small nit

SchSeba:
move a comment for test ID

related to https://github.com/K8sNetworkPlumbingWG/kubemacpool/pull/30#discussion_r297519047

### DIFF
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -62,8 +62,8 @@ var _ = Describe("Virtual Machines", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 		})
-		//2166
 		Context("When the client tries to assign the same MAC address for two different vm. Within Range and out of range", func() {
+			//2166
 			Context("When the MAC address is within range", func() {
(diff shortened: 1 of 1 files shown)


SchSeba:
@nikita133 please take a look

tests/virtual_machines_test.go
```diff
@@ -62,8 +62,8 @@ var _ = Describe("Virtual Machines", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 		})
-		//2166
 		Context("When the client tries to assign the same MAC address for two different vm. Within Range and out of range", func() {
+			//2166
```
phoracek: In kubevirt, the ID is part of context description, don't we want to follow the same schema? Shoudn't we add some more text to the ID, I get we don't want to reference CNV too much from this project, but 4 digits on their own would make no sense at all to random reader.

SchSeba: I just move this as @nikita133 requested.

I will like if he can open a new PR and move from comments to the context itself the test ID.

PR #42, opened 2019-07-01
Change Cert to file system mode

SchSeba:
This PR change the cert creation mode from secret to file system.

This way the pod will not make a restart in the first time.

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -38,12 +38,5 @@ resources:
   # manager_prometheus_metrics_patch.yaml should be enabled.
 #- manager_prometheus_metrics_patch.yaml
 
-vars:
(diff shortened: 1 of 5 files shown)


phoracek:
@SchSeba could you please explain to me what has changed here (in manifests)? I don't see any added/edited code, just removed.

SchSeba:
@phoracek I remove this section from the manager start config
```golang
Secret: &apitypes.NamespacedName{	
Namespace: "kubemacpool-system",	
Name:      "kubemacpool-webhook-secret",	
},
```

This change the default mode from secret to file system.
We only have one pod for the kubemacpool so this is fine.

if a new macpool is created he will update the cabundle inside the webhook config.

PR #43, opened 2019-07-02
Reload instead of restart

SchSeba:
This PR change the manager to reload when kubevirt is found
instead of reseting the pod

Before this PR if we deploy kubemacpool before kubevirt we start
a check loop for the virtual machine CRD and when we found it we
restart the kubemacpool pod to load the new watch controller.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -21,75 +21,30 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"os/signal"
-	"time"
 
-	"github.com/go-logr/logr"
-	"k8s.io/client-go/kubernetes"
-	"sigs.k8s.io/controller-runtime/pkg/client/config"
-	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 
-	"github.com/K8sNetworkPlumbingWG/kubemacpool/pkg/controller"
+	"github.com/K8sNetworkPlumbingWG/kubemacpool/pkg/manager"
 	poolmanager "github.com/K8sNetworkPlumbingWG/kubemacpool/pkg/pool-manager"
-	"github.com/K8sNetworkPlumbingWG/kubemacpool/pkg/webhook"
 )
 
-var log logr.Logger
-var metricsAddr string
-var restartChannel = make(chan struct{})
-
-func loadMacAddressFromEnvVar(envName string) net.HardwareAddr {
+func loadMacAddressFromEnvVar(envName string) (net.HardwareAddr, error) {
 	if value, ok := os.LookupEnv(envName); ok {
 		poolRange, err := net.ParseMAC(value)
 		if err != nil {
-			log.Error(err, "Failed to parse mac address", "mac address", value)
-			os.Exit(1)
+			return nil, err
 		}
 
-		return poolRange
+		return poolRange, nil
 	}
 
-	log.Error(fmt.Errorf("Environment variable %s don't exist", envName), "")
-	os.Exit(1)
-
-	// with out the the compilation failed
-	return nil
-}
-
-func checkForKubevirt(kubeClient *kubernetes.Clientset) bool {
-	result := kubeClient.ExtensionsV1beta1().RESTClient().Get().RequestURI("/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions/virtualmachines.kubevirt.io").Do()
-	if result.Error() == nil {
-		return true
-	}
-
-	return false
-}
-
-// Check for Kubevirt CRD to be available
-func waitForKubevirt(kubeClient *kubernetes.Clientset) {
-	for _ = range time.Tick(5 * time.Second) {
-		kubevirtExist := checkForKubevirt(kubeClient)
-		log.V(1).Info("kubevirt exist in the cluster", "kubevirtExist", kubevirtExist)
-		if kubevirtExist {
-			log.Info("found kubevirt restarting the manager")
-			close(restartChannel)
-			break
-		}
-	}
-}
-
-// wait for the any interrupt to stop the manager and clean the webhook.
-func waitForSignal() {
-	c := make(chan os.Signal, 1)
-	signal.Notify(c, os.Interrupt, os.Kill)
-	<-c
-	close(restartChannel)
+	return nil, fmt.Errorf("Environment variable %s don't exist", envName)
 }
 
 func main() {
+	var logType, metricsAddr string
+
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
-	var logType string
 	flag.StringVar(&logType, "v", "production", "Log type (debug/production).")
 	flag.Parse()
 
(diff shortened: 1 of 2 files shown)


phoracek:
A present for me!

pkg/manager/manager.go
```diff
@@ -0,0 +1,175 @@
...
+
+func NewKubeMacPoolManager() *KubeMacPoolManager {
+	kubemacpoolManager := &KubeMacPoolManager{continueToRunManager: true,
+		restartChannel:           make(chan struct{}),
+		kubevirtInstalledChannel: make(chan struct{}),
+		stopSignalChannel:        make(chan os.Signal, 1)}
+
+	signal.Notify(kubemacpoolManager.stopSignalChannel, os.Interrupt, os.Kill)
+
+	flag.StringVar(&kubemacpoolManager.metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
```
phoracek: Flags should be handled in main(). This function should receive native Go arguments passed from main. That way we split the application from its "front-end". Logger should be initialized in the main too. It depends on user of this application how logging should be handled.

pkg/manager/manager.go
```diff
@@ -0,0 +1,175 @@
...
+		logf.SetLogger(logf.ZapLogger(false))
+	}
+	log = logf.Log.WithName("manager")
+
+	return kubemacpoolManager
+}
+
+func (k *KubeMacPoolManager) Run() {
+	// setup Pool Manager
+	rangeStart := loadMacAddressFromEnvVar(poolmanager.RangeStartEnv)
```
phoracek: With env vars, I would be happier if they were loaded in main, but that is sometimes annoying once you need to pass the value all the way to its user, so I won't fight for this one.

pkg/manager/manager.go
```diff
@@ -0,0 +1,175 @@
...
+			go k.waitForKubevirt()
+		}
+		go k.waitForSignal()
+
+		log.Info("Setting up manager")
+		mgr, err := manager.New(k.config, manager.Options{MetricsBindAddress: k.metricsAddr})
+		ExitIfError(err, "unable to set up manager")
+
+		log.Info("Setting up controller")
+		ExitIfError(controller.AddToManager(mgr, poolManager), "Unable to register controllers to the manager")
```
phoracek: other errors start with lower case

pkg/manager/manager.go
```diff
@@ -0,0 +1,175 @@
...
+
+	close(k.restartChannel)
+}
+
+func loadMacAddressFromEnvVar(envName string) net.HardwareAddr {
+	if value, ok := os.LookupEnv(envName); ok {
+		poolRange, err := net.ParseMAC(value)
+		if err != nil {
+			log.Error(err, "Failed to parse mac address", "mac address", value)
+			os.Exit(1)
```
phoracek: functions in modules should return error, os.Exit should be handled in main only. Same reason as with flags and logging

pkg/manager/manager.go
```diff
@@ -0,0 +1,175 @@
...
+			close(k.kubevirtInstalledChannel)
+			break
+		}
+	}
+}
+
+// wait for the any interrupt to stop the manager and clean the webhook.
+func (k *KubeMacPoolManager) waitForSignal() {
+	select {
+	case <-k.stopSignalChannel:
```
phoracek: could you add inline comments to this function? Explaining what is being stopped and restarted, the whole process, container, only some internal loop?

SchSeba:
@phoracek can you take another look please?

cmd/manager/main.go
```diff
@@ -18,128 +18,45 @@ package main
...
-// wait for the any interrupt to stop the manager and clean the webhook.
-func waitForSignal() {
-	c := make(chan os.Signal, 1)
-	signal.Notify(c, os.Interrupt, os.Kill)
-	<-c
-	close(restartChannel)
-}
-
 func main() {
+	var log logr.Logger
```
phoracek: I saw this on multiple PRs now and I need to understand the motivation. In some cases, we need to initialize `var` before defining it, but this is not the case. Do you initialize it here on purpose? Do you consider it better and why? For me, it is a waste of space and it has no added value, does it? I need to hear that there is a any motivation/intention for it, so I can move over.

pkg/manager/manager.go
```diff
@@ -0,0 +1,161 @@
...
+	clientset                *kubernetes.Clientset
+	config                   *rest.Config
+	metricsAddr              string
+	continueToRunManager     bool
+	restartChannel           chan struct{}
+	kubevirtInstalledChannel chan struct{}
+	stopSignalChannel        chan os.Signal
+}
+
+func NewKubeMacPoolManager(logType, metricsAddr string, debuglevel bool) *KubeMacPoolManager {
```
phoracek: logType is not used in the function.

pkg/manager/manager.go
```diff
@@ -0,0 +1,161 @@
...
+	kubemacpoolManager := &KubeMacPoolManager{continueToRunManager: true,
+		restartChannel:           make(chan struct{}),
+		kubevirtInstalledChannel: make(chan struct{}),
+		stopSignalChannel:        make(chan os.Signal, 1),
+		metricsAddr:              metricsAddr}
+
+	signal.Notify(kubemacpoolManager.stopSignalChannel, os.Interrupt, os.Kill)
+
+	logf.SetLogger(logf.ZapLogger(debuglevel))
```
phoracek: Level to log should be set while initializing logger (ZapLogger in this case), the rest of the application (after main()) should not care about log levels, it should only log.

pkg/manager/manager.go
```diff
@@ -0,0 +1,160 @@
...
+func NewKubeMacPoolManager(metricsAddr string) *KubeMacPoolManager {
+	kubemacpoolManager := &KubeMacPoolManager{continueToRunManager: true,
+		restartChannel:           make(chan struct{}),
+		kubevirtInstalledChannel: make(chan struct{}),
+		stopSignalChannel:        make(chan os.Signal, 1),
+		metricsAddr:              metricsAddr}
+
+	signal.Notify(kubemacpoolManager.stopSignalChannel, os.Interrupt, os.Kill)
+
+	log = logf.Log.WithName("manager")
```
phoracek: shoudn't this be done in init()? NewKubeMacPoolManager is not handling singleton, it's just returning an instance and I can have as many as I want. Each of those would overwrite log.

cmd/manager/main.go
```diff
@@ -18,128 +18,43 @@ package main
...
-
 func main() {
+	var logType, metricsAddr string
+
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
-	var logType string
 	flag.StringVar(&logType, "v", "production", "Log type (debug/production).")
 	flag.Parse()
 
+	debuglevel := false
```
phoracek: can be moved to the else branch, not a big issue, just to make it symmetrical

pkg/manager/manager.go
```diff
@@ -0,0 +1,160 @@
...
+	config                   *rest.Config
+	metricsAddr              string
+	continueToRunManager     bool
+	restartChannel           chan struct{}
+	kubevirtInstalledChannel chan struct{}
+	stopSignalChannel        chan os.Signal
+}
+
+func NewKubeMacPoolManager(metricsAddr string) *KubeMacPoolManager {
+	kubemacpoolManager := &KubeMacPoolManager{continueToRunManager: true,
```
phoracek: This would make me happy

```
	kubemacpoolManager := &KubeMacPoolManager{
		continueToRunManager: true,
		restartChannel:           make(chan struct{}),
		kubevirtInstalledChannel: make(chan struct{}),
		stopSignalChannel:        make(chan os.Signal, 1),
		metricsAddr:              metricsAddr,
	}
```

pkg/manager/manager.go
```diff
@@ -0,0 +1,160 @@
...
+			return nil, err
+		}
+
+		return poolRange, nil
+	}
+
+	return nil, fmt.Errorf("Environment variable %s don't exist", envName)
+}
+
+func ExitIfError(err error, message string) {
```
phoracek: can we just return an error and handle system interface (return code) in main (if error is returned)

pkg/manager/manager.go
```diff
@@ -0,0 +1,160 @@
...
+	// This interrupt occurred when kubevirt is installed in the cluster and will restart the manager code only
+	// The container will not restart in this scenario
+	case <-k.kubevirtInstalledChannel:
+		log.Info("found kubevirt restarting the manager")
+	}
+
+	close(k.restartChannel)
+}
+
+func LoadMacAddressFromEnvVar(envName string) (net.HardwareAddr, error) {
```
phoracek: this function can be moved to main, it is valuable only there (split between interface and insides)

SchSeba:
Done with the requested changes @phoracek can you take another look please?

PR #52, opened 2019-07-29
Allow mutiple replicas for manager

SchSeba:
This PR allow to run the kubemacpool manager with multiple replicas

When a manager become leader it updates is pod label to a leader.
This label is the selector of the webhook service.
This way the service will always point only to the leader manager.

### DIFF
--- a/pkg/manager/leaderelection.go
+++ b/pkg/manager/leaderelection.go
@@ -1,0 +1,64 @@
+package manager
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/leaderelection"
+	manager_leaderelection "sigs.k8s.io/controller-runtime/pkg/leaderelection"
+)
+
+func (k *KubeMacPoolManager) newLeaderElection(config *rest.Config, scheme *runtime.Scheme) error {
+	recorderProvider, err := NewProvider(config, scheme, log.WithName("events"))
+	if err != nil {
+		return err
+	}
+
+	// Create the resource lock to enable leader election)
+	resourceLock, err := manager_leaderelection.NewResourceLock(config, recorderProvider, manager_leaderelection.Options{
+		LeaderElection:          true,
+		LeaderElectionID:        "kubemacpool-election",
+		LeaderElectionNamespace: "kubemacpool-system",
+	})
+	if err != nil {
+		return err
+	}
+
+	k.resourceLock = resourceLock
+
+	return nil
+}
+
+func (k *KubeMacPoolManager) waitToStartLeading() error {
+	messageChan := make(chan error, 1)
+
+	l, err := leaderelection.NewLeaderElector(leaderelection.LeaderElectionConfig{
+		Lock: k.resourceLock,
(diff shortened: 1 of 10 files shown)


SchSeba:
@alonSadan

pkg/manager/manager.go
```diff
@@ -160,6 +190,21 @@ func (k *KubeMacPoolManager) waitForSignal() {
 	close(k.restartChannel)
 }
 
+func (k *KubeMacPoolManager) markPodAsLeader() error {
```
phoracek: Can a race happen here? I'm not too familiar with the leader mechanism. Could it happen that two pods would claim to be leaders at the same moment?

pkg/manager/recorder.go
```diff
@@ -0,0 +1,61 @@
...
+	p.eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: clientSet.CoreV1().Events("")})
+	p.eventBroadcaster.StartEventWatcher(
+		func(e *corev1.Event) {
+			p.logger.V(1).Info(e.Type, "object", e.InvolvedObject, "reason", e.Reason, "message", e.Message)
+		})
+
+	return p, nil
+}
+
+func (p *provider) GetEventRecorderFor(name string) record.EventRecorder {
```
phoracek: Could you document what is a recorder?

pkg/manager/recorder.go
```diff
@@ -0,0 +1,61 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
phoracek: Is this just a copied module or do we have our own changes here?

pkg/manager/leaderelection.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/leaderelection"
+	manager_leaderelection "sigs.k8s.io/controller-runtime/pkg/leaderelection"
```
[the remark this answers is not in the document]
SchSeba: yep I think it's better to use the snake_case here

pkg/manager/manager.go
SchSeba: It should not append this is how we do the leader election in kubevirt and also implemented by kubernetes

pkg/manager/recorder.go
SchSeba: just copy

pkg/manager/recorder.go
SchSeba: because it's copy I will not like to add anything on this file

pkg/manager/leaderelection.go
```diff
@@ -0,0 +1,64 @@
...
+func (k *KubeMacPoolManager) newLeaderElection(config *rest.Config, scheme *runtime.Scheme) error {
+	recorderProvider, err := NewProvider(config, scheme, log.WithName("events"))
+	if err != nil {
+		return err
+	}
+
+	// Create the resource lock to enable leader election)
+	resourceLock, err := manager_leaderelection.NewResourceLock(config, recorderProvider, manager_leaderelection.Options{
+		LeaderElection:          true,
+		LeaderElectionID:        "kubemacpool-mac-election",
```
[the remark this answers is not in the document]
SchSeba: I don't know. my bad.

removed it

PR #54, opened 2019-08-19
test mac release after NIC deletion

alonSadan:
test that when a VM NIC is remmoved , it's
mac address is released and can be used as
a custom mac for another VM configuration.

### DIFF
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -363,7 +363,7 @@ var _ = Describe("Virtual Machines", func() {
 					}
 					return testClient.VirtClient.Update(context.TODO(), updateObject)
 
-				}, timeout, pollingInterval).Should(Not(HaveOccurred()), "failed to update client")
+				}, timeout, pollingInterval).ShouldNot(HaveOccurred(), "failed to update VM")
 
 				for index := range vm1.Spec.Template.Spec.Domain.Devices.Interfaces {
 					Expect(vm1.Spec.Template.Spec.Domain.Devices.Interfaces[index].MacAddress).To(Equal(updateObject.Spec.Template.Spec.Domain.Devices.Interfaces[index].MacAddress))
@@ -486,6 +486,63 @@ var _ = Describe("Virtual Machines", func() {
 				}, timeout, pollingInterval).Should(Not(HaveOccurred()), "failed to apply the new vm object")
 				_, err = net.ParseMAC(anotherVm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
 				Expect(err).ToNot(HaveOccurred())
+			})
+		})
+
+		Context("When a VM's NIC is removed and a new VM is created with the same MAC", func() {
+			It("should successfully release the MAC and the new VM should be created with no errors", func() {
+				err := setRange("02:00:00:00:00:00", "02:00:00:00:00:01")
+				Expect(err).ToNot(HaveOccurred())
+
+				vm := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br1", ""), newInterface("br2", "")},
+					[]kubevirtv1.Network{newNetwork("br1"), newNetwork("br2")})
+				Eventually(func() error {
+					return testClient.VirtClient.Create(context.TODO(), vm)
+
+				}, 50*time.Second, 5*time.Second).ShouldNot(HaveOccurred(), "failed to apply the new vm object")
+				_, err = net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
+				Expect(err).ToNot(HaveOccurred())
+				_, err = net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[1].MacAddress)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("checking that a new VM cannot be created when the range is full")
+				newVM := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br1", "")},
+					[]kubevirtv1.Network{newNetwork("br1")})
+				Eventually(func() error {
+					err = testClient.VirtClient.Create(context.TODO(), newVM)
+					if err != nil && strings.Contains(err.Error(), "Failed to create virtual machine allocation error: the range is full") {
+						return err
+					}
+					return nil
+
+				}, 50*time.Second, 5*time.Second).Should(HaveOccurred(), "should have failed to create the vm because the mac is already taken")
+
+				By("checking that the VM's NIC can be removed")
+				Eventually(func() error {
+					err := testClient.VirtClient.Get(context.TODO(),
+						client.ObjectKey{Namespace: vm.Namespace, Name: vm.Name}, vm)
+					if err != nil {
+						return err
+					}
+
+					vm.Spec.Template.Spec.Domain.Devices.Interfaces = []kubevirtv1.Interface{newInterface("br2", "")}
+					vm.Spec.Template.Spec.Networks = []kubevirtv1.Network{newNetwork("br2")}
+					err = testClient.VirtClient.Update(context.TODO(), vm)
+					if err != nil {
+						return err
+					}
+
+					if len(vm.Spec.Template.Spec.Domain.Devices.Interfaces) != 1 {
+						return fmt.Errorf("failed to delete the VM's interface")
+					}
+
+					return nil
+				}, 50*time.Second, 5*time.Second).ShouldNot(HaveOccurred(), "failed to update VM")
+
+				By("checking that a new VM can be created after the VM's NIC had been removed ")
+				Eventually(func() error {
+					return testClient.VirtClient.Create(context.TODO(), newVM)
+				}, 50*time.Second, 5*time.Second).ShouldNot(HaveOccurred(), "failed to create the new VM")
 			})
 		})
 	})


tests/virtual_machines_test.go
```diff
@@ -450,6 +449,58 @@ var _ = Describe("Virtual Machines", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 		})
+
+		Context("When a VM's NIC is removed and a new VM is created with the same MAC", func() {
+			It("should successfully release the MAC and the new Vm should be created with no errors ", func() {
```
SchSeba: remove the space at the end please

tests/virtual_machines_test.go
```diff
@@ -450,6 +449,63 @@ var _ = Describe("Virtual Machines", func() {
...
+					}
+					return nil
+
+				}, 50*time.Second, 5*time.Second).Should(Not(HaveOccurred()), "failed to update client")
+
+				By("after the VM's NIC has been deleted")
+				Expect(updatedVM.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress).To(Equal(mac2.String()),
+					"kubemacppool should assign the same mac as before the update")
+				Eventually(func() error {
+					return testClient.VirtClient.Create(context.TODO(), newVm)
```
SchSeba: here you need to create a new vm not the one you already use.

alonSadan: but I think the creation of "newVm" should fail, so it is not really used, isn't it?

tests/virtual_machines_test.go
```diff
@@ -450,6 +449,63 @@ var _ = Describe("Virtual Machines", func() {
...
+					if len(updatedVM.Spec.Template.Spec.Domain.Devices.Interfaces) != 1 {
+						return fmt.Errorf("failed to delete the VM's interface")
+					}
+					return nil
+
+				}, 50*time.Second, 5*time.Second).Should(Not(HaveOccurred()), "failed to update client")
+
+				By("after the VM's NIC has been deleted")
+				Expect(updatedVM.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress).To(Equal(mac2.String()),
+					"kubemacppool should assign the same mac as before the update")
```
yossisegev: typo: kubemacppool (you have 1 extra 'p').

alonSadan: Thanks:)

SchSeba:
please rebase this PR

tests/virtual_machines_test.go
```diff
@@ -489,6 +489,63 @@ var _ = Describe("Virtual Machines", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 		})
+
+		Context("When a VM's NIC is removed and a new VM is created with the same MAC", func() {
+			It("should successfully release the MAC and the new Vm should be created with no errors", func() {
```
SchSeba: Vm -> VM

tests/virtual_machines_test.go
```diff
@@ -489,6 +489,63 @@ var _ = Describe("Virtual Machines", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				mac2, err := net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[1].MacAddress)
+				Expect(err).ToNot(HaveOccurred())
+
+				updatedVM := *vm.DeepCopy()
+				By("before the NIC is deleted")
+				newVm := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br1", mac1.String())},
+					[]kubevirtv1.Network{newNetwork("br1")})
+				Eventually(func() error {
+					return testClient.VirtClient.Create(context.TODO(), newVm)
```
SchSeba: please validate the error here. we need to check that the error is related to the mac allocation and not something else

tests/virtual_machines_test.go
```diff
@@ -489,6 +489,63 @@ var _ = Describe("Virtual Machines", func() {
...
+				Eventually(func() error {
+					return testClient.VirtClient.Create(context.TODO(), vm)
+
+				}, 50*time.Second, 5*time.Second).Should(Not(HaveOccurred()), "failed to apply the new vm object")
+				mac1, err := net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
+				Expect(err).ToNot(HaveOccurred())
+				mac2, err := net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[1].MacAddress)
+				Expect(err).ToNot(HaveOccurred())
+
+				updatedVM := *vm.DeepCopy()
```
SchSeba: you don't need to Copy here.

just create the object
```golang
var updatedVM VirtualMachine
```

tests/virtual_machines_test.go
```diff
@@ -489,6 +489,63 @@ var _ = Describe("Virtual Machines", func() {
...
+					[]kubevirtv1.Network{newNetwork("br1")})
+				Eventually(func() error {
+					return testClient.VirtClient.Create(context.TODO(), newVm)
+
+				}, 50*time.Second, 5*time.Second).Should(HaveOccurred(), "should have failed to create the vm because the mac is allready taken")
+
+				By("deleting the VM's NIC")
+				Eventually(func() error {
+					err := testClient.VirtClient.Get(context.TODO(),
+						client.ObjectKey{Namespace: updatedVM.Namespace, Name: updatedVM.Name}, &updatedVM)
```
SchSeba: use here the vm object not the update

tests/virtual_machines_test.go
```diff
@@ -489,6 +489,63 @@ var _ = Describe("Virtual Machines", func() {
...
+					if err != nil {
+						return err
+					}
+					updatedVM.Spec.Template.Spec.Domain.Devices.Interfaces = []kubevirtv1.Interface{newInterface("br2", mac2.String())}
```
SchSeba: add a new line here

tests/virtual_machines_test.go
```diff
@@ -489,6 +489,63 @@ var _ = Describe("Virtual Machines", func() {
...
+					updatedVM.Spec.Template.Spec.Networks = []kubevirtv1.Network{newNetwork("br2")}
+					err = testClient.VirtClient.Update(context.TODO(), &updatedVM)
+					if err != nil {
+						return err
+					}
+
+					if len(updatedVM.Spec.Template.Spec.Domain.Devices.Interfaces) != 1 {
+						return fmt.Errorf("failed to delete the VM's interface")
+					}
+					return nil
```
SchSeba: add new line here

tests/virtual_machines_test.go
```diff
@@ -489,6 +489,63 @@ var _ = Describe("Virtual Machines", func() {
...
+
```
SchSeba: remove this empty line

tests/virtual_machines_test.go
SchSeba: oh you are right miss that we have vm, newVm and updateVM haha

tests/virtual_machines_test.go
```diff
@@ -489,6 +489,63 @@ var _ = Describe("Virtual Machines", func() {
...
+				}, 50*time.Second, 5*time.Second).Should(Not(HaveOccurred()), "failed to update client")
+
+				By("after the VM's NIC has been deleted")
+				Expect(updatedVM.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress).To(Equal(mac2.String()),
+					"kubemacpool should assign the same mac as before the update")
+				Eventually(func() error {
+					return testClient.VirtClient.Create(context.TODO(), newVm)
+
+				}, 50*time.Second, 5*time.Second).ShouldNot(HaveOccurred(), "failed to create the new VM")
+				Expect(newVm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress).To(Equal(mac1.String()), "kubemacppool should assign the freed mac")
```
SchSeba: freed -> released

tests/virtual_machines_test.go
```diff
@@ -489,6 +489,67 @@ var _ = Describe("Virtual Machines", func() {
...
+					if len(vm.Spec.Template.Spec.Domain.Devices.Interfaces) != 1 {
+						return fmt.Errorf("failed to delete the VM's interface")
+					}
+
+					return nil
+				}, 50*time.Second, 5*time.Second).Should(Not(HaveOccurred()), "failed to update client")
+
+				By("after the VM's NIC has been deleted")
+				Expect(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress).To(Equal(mac2.String()),
+					"kubemacpool should assign the same mac as before the update")
```
phoracek: This is not a good test. It assumes that we always use the first available mac from the pool, and that is undocumented and can change any time. I mean, it does the job, but is not really what we want to test. Could we make it, so our pool has only 2 available addresses, if one from removed NIC has been recycled, we should be able to create a VM here and get **a** MAC. Sorry if I'm nitpicking too much.

alonSadan: HI @phoracek .It's all good :)
I didn't understand though whats the point of limiting the range if we are not checking kubemacpool assigns the same mac that was released (the only one that is available).
If we are checking that the new Vm received just "a MAC" than maybe we can leave the range as it is? 
Thanks:)

phoracek: This test schould check "do we release MAC after this and that happens with a VM". What you wrote checks that, but also "we always assign the first MAC address from the pool". The latter check is not documented and irrelevant.

I was thinking how could we check 1. without 2. What came to my mind is, let's limit the pool to two MAC addresses during the deployment. Then if the MAC is not released, we fail to allocate, if we release, we will have a MAC address to assign. Same results, less assumptions.

Does that make sense? Trying to do BDD testing, hehe.

Anyway, right now we don't control the range while testing, do we? It is only configured once during the deployment, isn't it?

SchSeba: we can do it

```golang
err := setRange(rangeStart, rangeEnd)
```

 we can put a range just before any test

alonSadan: > Does that make sense? Trying to do BDD testing, hehe.
OK. I see your point here.
NP I'll change that.
> Anyway, right now we don't control the range while testing, do we? It is only configured once during the deployment, isn't it?
I think we can. The "setRange" function sets the range for the macpool by configuring the range and deleting the kubemacpool pod and waiting for it to run again.

phoracek: I think it would be the right way to do it.

alonSadan:
Hi @phoracek .
what do you think?

tests/virtual_machines_test.go
```diff
@@ -488,6 +488,64 @@ var _ = Describe("Virtual Machines", func() {
...
+				Eventually(func() error {
+					return testClient.VirtClient.Create(context.TODO(), vm)
+
+				}, 50*time.Second, 5*time.Second).ShouldNot(HaveOccurred(), "failed to apply the new vm object")
+				_, err = net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
+				Expect(err).ToNot(HaveOccurred())
+				_, err = net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[1].MacAddress)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("before the NIC is deleted")
```
phoracek: This message won't be too helpful in test output. It should describe the following block in format "By checking that X is/does Y when Z"

tests/virtual_machines_test.go
```diff
@@ -488,6 +488,64 @@ var _ = Describe("Virtual Machines", func() {
...
+					if err != nil {
+						return err
+					}
+
+					if len(vm.Spec.Template.Spec.Domain.Devices.Interfaces) != 1 {
+						return fmt.Errorf("failed to delete the VM's interface")
+					}
+
+					return nil
+				}, 50*time.Second, 5*time.Second).Should(Not(HaveOccurred()), "failed to update client")
```
phoracek: .ShouldNot(HaveOccurred())

phoracek: Update client? It updates VM's spec, what do you mean by "client"?

tests/virtual_machines_test.go
```diff
@@ -488,6 +488,64 @@ var _ = Describe("Virtual Machines", func() {
...
+
+				By("after the VM's NIC has been deleted")
+				Eventually(func() error {
+					return testClient.VirtClient.Create(context.TODO(), newVM)
+
```
phoracek: please drop the empty line

tests/virtual_machines_test.go
```diff
@@ -488,6 +488,64 @@ var _ = Describe("Virtual Machines", func() {
...
+					}
+
+					if len(vm.Spec.Template.Spec.Domain.Devices.Interfaces) != 1 {
+						return fmt.Errorf("failed to delete the VM's interface")
+					}
+
+					return nil
+				}, 50*time.Second, 5*time.Second).Should(Not(HaveOccurred()), "failed to update client")
+
+				By("after the VM's NIC has been deleted")
```
phoracek: ditto, should describe what you are checking

tests/virtual_machines_test.go
alonSadan: oops...

PR #55, opened 2019-08-20
Add debug logs on tests

SchSeba:
This PR add the output from the kubemacpool manager containers
in case of an error in the tests.


alonSadan:
looks good to me

SchSeba:
No need any more we add this changes to the finilizer PR

PR #56, opened 2019-08-21
add affinty to constrain kubemacpool pods

alonSadan:
we added a pod-anti-affinity to the kubemacpool pod spec which constrain
(in a *soft way) the scheduling of the pod so that
no more than one kubemacpool-pod will be scheduled on a single node.

*if there is no other choice, it will bypass the constraint,
and will create more than one kubemacpool pods on a single node.

### DIFF
--- a/config/default/manager/manager.yaml
+++ b/config/default/manager/manager.yaml
@@ -39,6 +39,18 @@ spec:
         control-plane: mac-controller-manager
         controller-tools.k8s.io: "1.0"
     spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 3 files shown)


alonSadan:
@SchSeba

config/default/manager/manager.yaml
```diff
@@ -39,6 +39,19 @@ spec:
...
+          - podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: control-plane
+                  operator: In
+                  values:
+                  - mac-controller-manager
+              topologyKey: kubernetes.io/hostname
+            weight: 1
+
```
SchSeba: remove this empty line

PR #57, opened 2019-08-26
add pod disruption budget to kubemacapool

alonSadan:
add pod disruption budjet to kubemacpool to set
the minimum number of available kubemacpool pods
on one node to 1 (on volentary distruption)

### DIFF
--- a/config/default/manager/manager.yaml
+++ b/config/default/manager/manager.yaml
@@ -91,3 +91,13 @@ spec:
           name: webhook-server
           protocol: TCP
       terminationGracePeriodSeconds: 5
+---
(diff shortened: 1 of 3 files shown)


alonSadan:
@SchSeba

config/default/manager/manager.yaml
```diff
@@ -91,3 +91,13 @@ spec:
           name: webhook-server
           protocol: TCP
       terminationGracePeriodSeconds: 5
+---
+apiVersion: v1
```
SchSeba: This should be `policy/v1beta1`

PR #59, opened 2019-08-27
add a cleanup loop for failed vm using configmap

SchSeba:
Before this PR we have an issue if the vm creation fails because there
is an validation error from the virt-api we didn't clean the allocation.

This PR introduce a cleanup loop into the system.
When we hit the create vm mutating webhook we create the regular
allocation but we also mark the mac address in  a waiting configmap.

We have a waiting look the will check if the object is removed from the
map and if is not after 30 second we assume the object wasn't saved
into the etcd (no controller event) so we release the object.

If we get an event from the controller then we remove the mac address
from the configmap.

This PR are also add/remove some empty lines in the code, also fix the import positions and fix the output when a functional test fails


SchSeba:
@alonSadan please take a look

pkg/manager/manager.go
```diff
@@ -49,18 +50,20 @@ type KubeMacPoolManager struct {
 	stopSignalChannel        chan os.Signal
 	podNamespace             string
 	podName                  string
+	waitingTime              int
```
phoracek: Please add a docstring explaining why we have this attribute

pkg/pool-manager/pool.go
```diff
@@ -31,6 +31,7 @@ const (
 	RuntimeObjectFinalizerName = "k8s.v1.cni.cncf.io/kubeMacPool"
 	networksAnnotation         = "k8s.v1.cni.cncf.io/networks"
 	networksStatusAnnotation   = "k8s.v1.cni.cncf.io/networks-status"
+	vmWaitConfigMapName        = "kubemacpool-vm-configmap"
```
phoracek: what is this CM good for?

pkg/names/names.go
```diff
@@ -0,0 +1,15 @@
+package names
```
phoracek: could you introduce this change in a different commit please?

pkg/pool-manager/pool_test.go
```diff
@@ -100,13 +103,7 @@ var _ = Describe("Pool", func() {
...
 	Describe("Pool Manager General Functions ", func() {
 		It("should create a pool manager", func() {
-			fakeClient := fake.NewSimpleClientset()
-			startPoolRangeEnv, err := net.ParseMAC("02:00:00:00:00:00")
-			Expect(err).ToNot(HaveOccurred())
-			endPoolRangeEnv, err := net.ParseMAC("02:FF:FF:FF:FF:FF")
-			Expect(err).ToNot(HaveOccurred())
-			_, err = NewPoolManager(fakeClient, startPoolRangeEnv, endPoolRangeEnv, false)
-			Expect(err).ToNot(HaveOccurred())
+			createPoolManager("02:00:00:00:00:00", "02:FF:FF:FF:FF:FF")
```
phoracek: could you please do this change in a separate commit?

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -333,6 +355,121 @@ func (p *PoolManager) revertAllocationOnVm(vmName string, allocations map[string
 	p.releaseMacAddressesFromInterfaceMap(allocations)
 }
 
+func (p *PoolManager) getOrCreateVmMacWaitMap() (map[string]string, error) {
```
phoracek: do you think it would make sense to keep these helpers in a separate module. this logic is not trivial and it would be easier to document there

pkg/manager/manager.go
SchSeba: done

pkg/pool-manager/pool.go
SchSeba: This CM saved the allocated mac addresses in waiting state.

We need to use it in case of a race condition that our manager go down and It have that types of allocations so the new leader can continue from the same state.

pkg/pool-manager/virtualmachine_pool.go
SchSeba: I will prefer to document it here as we are going to have the same functions for pods in the future.

I will add comments onto of every function

pkg/names/names.go
SchSeba: done

pkg/pool-manager/pool_test.go
SchSeba: done

pkg/pool-manager/pool.go
phoracek: Sounds dangerous synchronisation-wise, but that's not different from the rest of kubemacpool... :D

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -333,6 +355,124 @@ func (p *PoolManager) revertAllocationOnVm(vmName string, allocations map[string
...
+// so we release the virtual machine
+func (p *PoolManager) vmWaitingCleanupLook(waitTime int) {
+	c := time.Tick(3 * time.Second)
+	log.Info("starting cleanup loop for waiting mac addresses")
+	for _ = range c {
+		p.poolMutex.Lock()
+
+		configMap, err := p.kubeClient.CoreV1().ConfigMaps(names.MANAGER_NAMESPACE).Get(vmWaitConfigMapName, metav1.GetOptions{})
+		if err != nil {
+			// TODO: should we exit here?
```
phoracek: Do we need to resolve this?

SchSeba: we can remove this.

The next loop will try again

SchSeba:
Open an issue to track the implementation https://github.com/K8sNetworkPlumbingWG/kubemacpool/issues/62

alonSadan:
looks good!

PR #63, opened 2019-09-01
add pod distruption budjet to kubemcapool

alonSadan:
add pod disruption budjet to kubemacpool to set
the minimum number of available kubemacpool pods
on one node to 1 (on volentary distruption)


alonSadan:
I made a mistake with the previous PR and not included namespace...

alonSadan:
I closed this PR (which is an override on an original PR to add podDisruptionBudget)  because in the PR I added a namespace metadata  which is not necessary because kustomization already adds the namespace

PR #64, opened 2019-09-05
Bump kubevirt package version to v0.20.4

SchSeba:
This PR bump the kubevirt pakcage we use to v0.20.4

I also update the kubevirt and cluster-network-addons-operator deployments to latest one for the test clusters environments.

/cc @alonSadan


Gopkg.toml
```diff
@@ -36,7 +37,14 @@ version="v1.4.7"
...
+  version = "=v0.20.4"
+
 [[constraint]]
-  name= "kubevirt.io/kubevirt"
-  source = "https://github.com/kubevirt/kubevirt.git"
-  version = "v0.17.0"
+  name = "kubevirt.io/client-go"
+  version = "=v0.20.4"
+
+[[override]]
```
phoracek: why do we need CDI?

pkg/webhook/virtualmachine/virtualmachine.go
```diff
@@ -110,7 +110,21 @@ func (a *virtualMachineAnnotator) mutateCreateVirtualMachinesFn(ctx context.Cont
 		"virtualMachineName", virtualMachine.Name,
 		"virtualMachineNamespace", virtualMachine.Namespace)
 
-	return a.poolManager.AllocateVirtualMachineMac(virtualMachine)
+	previousVirtualMachine := &kubevirt.VirtualMachine{}
```
phoracek: existingVirtualMachine maybe?

pkg/webhook/virtualmachine/virtualmachine.go
```diff
@@ -110,7 +110,21 @@ func (a *virtualMachineAnnotator) mutateCreateVirtualMachinesFn(ctx context.Cont
...
+	err := a.client.Get(context.TODO(), client.ObjectKey{Namespace: virtualMachine.Namespace, Name: virtualMachine.Name}, previousVirtualMachine)
+	if err != nil {
+		// Check if the object already exist.
```
phoracek: It does the opposite. Maybe describe what are you trying to do there. "If the VM does not exist yet, allocate a new MAC address"

Gopkg.toml
SchSeba: This is an hard request by the kubevirt project and the version should be override (it takes the master) this is way is override and not in the request list of the project

PR #66, opened 2019-09-06
Fix mutiple replica code

SchSeba:
This PR fix an issue we have when we create the kubemacpool before
kubevirt was installed and then we try to restart the manager after
we find kubevirt in the cluster.

The manager was stuck waiting to be the leader (but it's already
the leader)

/cc @alonSadan

### DIFF
--- a/pkg/manager/leaderelection.go
+++ b/pkg/manager/leaderelection.go
@@ -2,13 +2,15 @@ package manager
 
 import (
 	"context"
-	"fmt"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection"
 	manager_leaderelection "sigs.k8s.io/controller-runtime/pkg/leaderelection"
+
+	"github.com/K8sNetworkPlumbingWG/kubemacpool/pkg/names"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/manager/leaderelection.go
```diff
@@ -20,45 +22,68 @@ func (k *KubeMacPoolManager) newLeaderElection(config *rest.Config, scheme *runt
...
 		LeaseDuration: 15 * time.Second,
 		RenewDeadline: 10 * time.Second,
 		RetryPeriod:   2 * time.Second,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(_ context.Context) {
-				messageChan <- nil
+				k.leaderElectionChannel <- nil
 			},
 			OnStoppedLeading: func() {
 				// Most implementations of leader election log.Fatal() here.
```
phoracek: The comment is outdated

pkg/manager/manager.go
```diff
@@ -188,26 +188,13 @@ func (k *KubeMacPoolManager) waitForSignal() {
 	// The container will not restart in this scenario
 	case <-k.kubevirtInstalledChannel:
 		log.Info("found kubevirt restarting the manager")
+	case <-k.restartChannel:
```
phoracek: Please add a comment describing what happens here when and why

SchSeba:
@phoracek can you please take another look I think it should be ready now.

Thanks for the comments!

PR #68, opened 2019-09-11
Update waiting time

SchSeba:
We need this PR because it can take time the manager to reload
and the clean up function can remove the mac from the waiting look.

This PR also remove the vm  from the waiting list event if we can't
update the object to add the finilizer

---
This change is [Reviewable](https://reviewable.io/reviews/k8snetworkplumbingwg/kubemacpool/68)

### DIFF
--- a/pkg/controller/virtualmachine/virtualmachine_controller.go
+++ b/pkg/controller/virtualmachine/virtualmachine_controller.go
@@ -114,6 +114,10 @@ func (r *ReconcilePolicy) addFinalizerAndUpdate(virtualMachine *kubevirt.Virtual
 		return nil
 	}
 
+	if err := r.poolManager.MarkVMAsReady(virtualMachine); err != nil {
+		return err
+	}
+
 	log.V(1).Info("The VM does not have a finalizer",
 		"virtualMachineName", request.Name,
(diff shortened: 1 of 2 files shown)


pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -114,6 +114,10 @@ func (r *ReconcilePolicy) addFinalizerAndUpdate(virtualMachine *kubevirt.Virtual
 		return nil
 	}
 
+	if err := r.poolManager.MarkVMAsReady(virtualMachine); err != nil {
```
phoracek: For the commit message. Please explain when does it happen and what are the consequences.

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -314,7 +314,7 @@ func (p *PoolManager) initVirtualMachineMap() error {
 		}
 	}
 
-	return nil
+	return p.UpdateWaitingTimeOnInit(waitingMac)
```
phoracek: Commit message comment:
That is one long sentece. Please exaplain when does it happen, what happens, what is the consequence.

PR #69, opened 2019-09-17
set webhooks failure policy to Ignore

phoracek:
We are still facing many issues with inaccessible webhook blocking
creation of VMs and especially pods. Let's make our failure policy
ignore for now and investigate mentioned issues only in test
environments for now.

### DIFF
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -18,7 +18,7 @@ import (
 	kubevirtv1 "kubevirt.io/client-go/api/v1"
 )
 
-const timeout = 60 * time.Second
+const timeout = 2 * time.Minute
 const pollingInterval = 5 * time.Second
 
 var _ = Describe("Virtual Machines", func() {
--- a/pkg/webhook/pod/pod.go
+++ b/pkg/webhook/pod/pod.go
@@ -42,7 +42,7 @@ func Add(mgr manager.Manager, poolManager *pool_manager.PoolManager, namespaceSe
 
 	wh, err := builder.NewWebhookBuilder().
 		Mutating().
-		FailurePolicy(admissionregistrationv1beta1.Fail).
+		FailurePolicy(admissionregistrationv1beta1.Ignore).
 		Operations(admissionregistrationv1beta1.Create).
 		ForType(&corev1.Pod{}).
 		Handlers(podAnnotator).


dankenigsberg:
I think it would be useful to state in the commit message the recently reported mode of failure: macpool prevented the upgrade of OpenShiftSDN. starting the SDN Pod requires the macpool, but the macpool uses SDN addresses, which are unavailable when the SDN is down.

I suspect we have to add a test for SDN recovery when macpool is enabled.

SchSeba:
can we change the PR to ignore only on the pods?

dankenigsberg:
> can we change the PR to ignore only on the pods?

I clearly recall a promise: "on the next blocker bug we go to Ignore". It was assertive with no questions ;-)

I'm all for testing granular changes, and solving all bugs, but I think that we have waited enough. I vote for Ignore everywhere - but we should file an Issue to track a proper solution.

phoracek:
@dankenigsberg commit message updated, issue opened https://github.com/K8sNetworkPlumbingWG/kubemacpool/issues/70

phoracek:
@dankenigsberg @SchSeba lot of our tests expect failure to happen. We are blocking not only cases where Kubernetes fails to reach the webhook, but also cases where the validation simply fails.

dankenigsberg:
Oh, that's bad. I suspect there is no IgnoreOnlyIfNonresponsive.

I don't see a way around this. Currently we can:
* disable macpool
* ignore VM&Pod failures and skip negative tests
* ignore only Pod failures and skip some tests

Earlier I voted against the third option, but it does allow us to save SOME face. Can we reconsider it?

Either

phoracek:
I will go over the test we'd need to skip tomorrow to see if it makes sense.

SchSeba:
+1 on the "ignore only Pod failures and skip some tests"

dankenigsberg:
@SchSeba would it be possible register the webhook so that it ignores `openshift-*` namespaces?

phoracek:
@dankenigsberg we looked into it. You can (de-)select Namespaces only based on their labels, not names. And AFAIK, openshift-* namespaces don't have any specific label. Worth double-checking though.

phoracek:
@dankenigsberg we can list all namespaces we want to ignore. However, I think we should just set ignore (and revert shortly after release).

SchSeba:
> @dankenigsberg we looked into it. You can (de-)select Namespaces only based on their labels, not names. And AFAIK, openshift-* namespaces don't have any specific label. Worth double-checking though.

We can't do it.

For now we can only specify a label on a namespace that if it exist we skip or we enforce.
But not on the namespace name.

SchSeba:
> @dankenigsberg we can list all namespaces we want to ignore. However, I think we should just set ignore (and revert shortly after release).

We can run on every start on all the namespaces check for openshift-* and update the namespace with a label but we can't control if someone (operator) remove or update that namespace labels

dankenigsberg:
> However, I think we should just set ignore (and revert shortly after release).

Go go go, then.

phoracek:
@dankenigsberg @SchSeba please review

tests/virtual_machines_test.go
```diff
@@ -252,6 +252,8 @@ var _ = Describe("Virtual Machines", func() {
 		//2165
 		Context("when trying to create a VM after a MAC address has just been released duo to a VM deletion", func() {
 			It("should re-use the released MAC address for the creation of the new VM and not return an error", func() {
+				Skip("Failures on Pod reconciliation are merely ignored https://github.com/K8sNetworkPlumbingWG/kubemacpool/pull/69")
```
SchSeba: This test is for VM why we need to skip it if we change the polocy for pods?

phoracek: This is the only one that failed with Ignore policy on pod.

PR #71, opened 2019-10-07
Create a ownerRef for the service

SchSeba:
This PR creates a OwnerRef for the manager service.
When the manager deployment will be removed it will also remove the service.

Before this PR if we remove the deployment it will not remove the service
related to him.

Signed-off-by: Sebastian Sch <sebassch@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/k8snetworkplumbingwg/kubemacpool/71)

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -115,10 +115,15 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 		}
 
 		// create a owner ref on the mutating webhook
-		// this way when we remove the statefulset of the manager the webhook will be also removed from the cluster
+		// this way when we remove the deployment of the manager the webhook will be also removed from the cluster
 		err = webhook.CreateOwnerRefForMutatingWebhook(k.clientset, k.podNamespace)
 		if err != nil {
 			return fmt.Errorf("unable to create owner reference for mutating webhook object error %v", err)
+		}
+
+		err = webhook.CreateOwnerRefForService(k.clientset, k.podNamespace)
+		if err != nil {
+			return fmt.Errorf("unable to create owner reference for service object error %v", err)
 		}
 
 		isKubevirtInstalled := checkForKubevirt(k.clientset)
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -86,6 +86,7 @@ func main() {
 
 	err = kubemacpoolManager.Run(rangeStart, rangeEnd)
 	if err != nil {
+		log.Error(err, "Failed to run the manager")
 		os.Exit(1)
 	}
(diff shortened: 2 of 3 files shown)


SchSeba:
> @SchSeba do we need to create the service dynamically? Would it be possible to have a static manifest for it?

I think it's better to have it in code just like we have the mutating webhook (so we can be consistent).

pkg/manager/manager.go
```diff
@@ -157,6 +168,40 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
...
+				},
+				Spec: corev1.ServiceSpec{
+					Selector: map[string]string{
+						names.LEADER_LABEL: "true",
+					}, Ports: []corev1.ServicePort{{
+						Port: 443,
+						TargetPort: intstr.IntOrString{
+							Type:   intstr.Int,
+							IntVal: 8000,
+						}}}}}
```
phoracek: this passed gofmt?

pkg/manager/manager.go
```diff
@@ -115,12 +121,17 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 		}
 
 		// create a owner ref on the mutating webhook
-		// this way when we remove the statefulset of the manager the webhook will be also removed from the cluster
+		// this way when we remove the deployment of the manager the webhook will be also removed from the cluster
 		err = webhook.CreateOwnerRefForMutatingWebhook(k.clientset, k.podNamespace)
```
phoracek: This function is namespaced, don't we want to keep the new owner ref function in the same place?

pkg/utils/utils.go
```diff
@@ -16,6 +16,23 @@ limitations under the License.
...
 package utils
 
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/K8sNetworkPlumbingWG/kubemacpool/pkg/names"
+)
+
+func CreateDeploymentOwnerRef(kubeClient *kubernetes.Clientset, managerNamespace string) ([]metav1.OwnerReference, error) {
```
phoracek: If you move `createOwnerRefForService` under `webhook`, this could be moved as well. This is weirdly specific function to live in `utils`.

pkg/manager/manager.go
SchSeba: yep

pkg/manager/manager.go
SchSeba: done

SchSeba:
@phoracek can you please take another look on this PR it should be ready now

PR #74, opened 2019-11-05
Skip pods in critical namespaces

SchSeba:
This PR fix the issue when a Critical pod like the sdn provider
need to be recreated.

change the mutating webhook to skip namespaces to contains

1. runlevel=[0,1] for kubernetes environments
2. openshift.io/run-level=[0,1] for openshift environments

With this change we can change the FailurePolicy from ignore to fail.

Signed-off-by: Sebastian Sch <sebassch@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/k8snetworkplumbingwg/kubemacpool/74)

PR #78, opened 2020-01-08
BZ1779513: make webhook ignoring

AlonaKaplan:
With https://github.com/k8snetworkplumbingwg/kubemacpool/pull/74
we set webhook failing policy to failing in belief the kubemacpool
became stable. Unfortunatelly, we still see some certificate issues
as listed in https://bugzilla.redhat.com/1779513 (most probably due to
ca rotation).

The policy is moved back to failing till the issues are fixed.
The fixes will be tracked in - https://bugzilla.redhat.com/1787213

Signed-off-by: Alona Kaplan <alkaplan@redhat.com>

### DIFF
--- a/pkg/webhook/virtualmachine/virtualmachine.go
+++ b/pkg/webhook/virtualmachine/virtualmachine.go
@@ -50,7 +50,7 @@ func Add(mgr manager.Manager, poolManager *pool_manager.PoolManager, namespaceSe
 
 	wh, err := builder.NewWebhookBuilder().
 		Mutating().
-		FailurePolicy(admissionregistrationv1beta1.Fail).
+		FailurePolicy(admissionregistrationv1beta1.Ignore).
 		Operations(admissionregistrationv1beta1.Create, admissionregistrationv1beta1.Update).
 		ForType(&kubevirt.VirtualMachine{}).
 		Handlers(virtualMachineAnnotator).
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -146,14 +146,20 @@ func setRange(rangeStart, rangeEnd string) error {
 		}
 
 		for _, pod := range podsList.Items {
+			if !pod.ObjectMeta.DeletionTimestamp.IsZero() {
+				return fmt.Errorf("old pod %s has not yet been removed\n", pod.Name)
+			}
 			if pod.Status.Phase != corev1.PodRunning {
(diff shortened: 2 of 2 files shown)


phoracek:
I see we already ignore it in https://github.com/k8snetworkplumbingwg/kubemacpool/blob/v0.8.0/pkg/webhook/pod/pod.go#L45.

Let's wait for CI.

phoracek:
Since v0.8.0 the name of the org changed from `K8sNetworkPlumbingWG` to `k8snetworkplumbingwg`. Could you please submit a PR changing the case in the whole project and once we merge it, rebase this?

phoracek:
@AlonaKaplan you need to change some tests. They are e.g. expected to fail when invalid MAC is provided. However, with this change, they don't return failure anymore.

phoracek:
@AlonaKaplan I don't think this is a solution to the issue. There is a race where sometimes we don't assign any MAC. Your changed `createVM` is executed several times, trying to create and check the VM over and over, so it eventually does not get into the race.

I did some debugging. I run Eventually around Get, but as soon as it succeeds, I check the MAC only once, no retries. With webhook, the MAC entry should be there as soon as the VM is available on the API server. In other words, if we can Get it from the API server, it already passed the webhook.

I ran it several times, sometimes it assignes the MAC (I also checked the VM definition via kubectl manually), sometimes it does not.

Then I noticed that kubemacpool pods are removed and recreated every time we run tests. That is caused by the `setRange` function. For every test, it sets up given range. In order to apply the range, the pods need to be restarted. I changed the code, so it waits for the old pods to be removed and new ones successfully started.

Another issue was, that kubemacpool pods don't have readiness probes implemented (and it takes ~30 seconds to start the webhook service after the pod itself turns Running). This explains why you saw the log only once the test was done. The test passed with the previous set of managers (before `setRange`). Once the new set of pods started, it reconciled existing pods, but it was too late. For this change, we will need a new release of CNAO, since readiness probes are part of manifests. Also, CNAO on D/S may have issues with one of the managers being idle (not ready, waiting to take over the leadership).

With these two changes, all the positive tests are consistently passing. The former issue is only in tests. The latter may happen in production (it would be ignored, by we better fix it).

I will send a fix for the function tomorrow. I want to fix it on master first. I pushed a commit with preview of the changes here: https://github.com/k8snetworkplumbingwg/kubemacpool/pull/80

PR #82, opened 2020-01-10
Update the port number in the manager deployment

phoracek:
The port written in our manifests is not matching the one used by webserver. Since the port in service is only for documentation purposes, it should not be a big deal, but let's be on the safe side.


phoracek:
cc @AlonaKaplan

phoracek:
The test failure is unrelated.

PR #83, opened 2020-01-10
tests: wait for new managers to become available

phoracek:
In the current code, setRange and deleteLeaderManager functions
are not waiting for the old pods to disappear and new ones to become
available. Due to that, tests using this functions don't have any
webhook available or worse, they are being served by the old one.

This issue was not exposed due to Eventually we have around all the
Create calls.


phoracek:
cc @AlonaKaplan

PR #84, opened 2020-01-10
add readiness probe

phoracek:
Currently, manager pods turn Ready as soon as their binary starts.
However, it takes ~40 more seconds for the webhook to actually start
serving requests. In order to avoid that, we introduce a readiness
probe that marks the pod as Ready only once the server starts.

Please note that there is no way to do it properly with the current
controller-runtime. We don't have a hook point that would get executed
once the webhook has started. Therefore we have a Sleep there. That
should be dropped once
https://github.com/kubernetes-sigs/cluster-api/issues/1855
is implemented.

### DIFF
--- a/config/default/manager/manager.yaml
+++ b/config/default/manager/manager.yaml
@@ -92,6 +92,13 @@ spec:
         - containerPort: 8000
           name: webhook-server
           protocol: TCP
+        readinessProbe:
+          exec:
+            command:
+            - cat
+            - /tmp/ready
+          initialDelaySeconds: 5
+          periodSeconds: 5
       terminationGracePeriodSeconds: 5
 ---
 apiVersion: policy/v1beta1
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -149,6 +149,17 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 			return fmt.Errorf("unable to register webhooks to the manager error %v", err)
 		}
 
+		go func() {
+			// TODO: We cannot set the probe any closer to the server start.
+			// Therefore, we add this terrible sleep. It should be replaced by
+			// proper implementation once https://github.com/kubernetes-sigs/cluster-api/issues/1855
+			// is finished.
+			time.Sleep(time.Second * 20)
+			log.Info("Manager has hopefully started, marking as ready")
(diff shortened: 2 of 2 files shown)


phoracek:
cc @AlonaKaplan @dankenigsberg

config/default/manager/manager.yaml
```diff
@@ -92,6 +92,13 @@ spec:
         - containerPort: 8000
           name: webhook-server
           protocol: TCP
+        readinessProbe:
+          exec:
+            command:
+            - cat
+            - /tmp/ready
+          initialDelaySeconds: 5
```
dankenigsberg: why do we need any golang code for this hack? Running `cat` of an ever-existing file every 5 seconds feels silly.
Would it not be enough to have
```
   initialDelaySeconds: 20 
```
with a fat comment carrying the same explanation?

phoracek: We need to create the file once the server starts.

I don't think saving 3 `cat` runs is worth an extra line and comment.

We can replace `cat` with HTTP GET once controller-runtime implements support for it. I'm not aware of any option that would allow us to stop querying once it succeeds once. https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes

dankenigsberg: I was speaking about something like
```
       readinessProbe:
          exec:
            command:
            - true
          initialDelaySeconds: 20
          periodSeconds: 99999
```
and dropping the golang changes altogether. But I'm splitting hairs here. The important thing is that you have found the races that @AlonaKaplan has spoken about.

phoracek: That would not help. True can succeed as soon as the pod starts. That would be the same as happens with default readiness.

The point here is, that we write the file only once we go through all the initialization and we start the webhook server. In other words, we are marked Ready only once we are able to respond to HTTPS requests for /mutate-virtualmachines.

phoracek: We still have races in the code itself. Unfortunately, the tests are more resilient than the code, so it is all hidden.

phoracek: The extra 20 seconds wait is on top of all the initialization (collecting existing MACs, getting certificates). Those 20 seconds are waiting until the `Start` hopefully really starts serving.

dankenigsberg: gotcha.

PR #85, opened 2020-01-15
Update webhooks to comply with latest controller-runtime api

RamLavi:
The need for using diverse certification management methods require us to update to the latest controller-runtime, update to Builder docker file to use go version 1.12.+, and to change the way we open webhook server accordingly. 


pkg/manager/manager.go
```diff
@@ -144,6 +143,7 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 			return fmt.Errorf("unable to register controllers to the manager error %v", err)
 		}
 
+		log.Info("ram-sam-sam1")
```
phoracek: !

Makefile
```diff
@@ -41,35 +42,55 @@ generate-test: manifests
 	kustomize build config/test > config/test/kubemacpool.yaml
 
 # Generate manifests e.g. CRD, RBAC etc.
+<<<<<<< HEAD
```
qinqon: Bad merge.

pkg/webhook/pod/pod.go
```diff
@@ -32,28 +31,18 @@ import (
...
 
 type podAnnotator struct {
 	client      client.Client
 	decoder     *admission.Decoder
 	poolManager *pool_manager.PoolManager
 }
 
-func Add(mgr manager.Manager, poolManager *pool_manager.PoolManager) error {
+// Add adds server modifiers to the server, like registering the hook to the webhook server.
+func Add(mgr manager.Manager, s *webhookserver.Server, poolManager *pool_manager.PoolManager) error {
```
qinqon: mgr is not needed

pkg/webhook/webhook.go
```diff
@@ -40,18 +46,20 @@ import (
 // +kubebuilder:rbac:groups="apiextensions.k8s.io",resources=customresourcedefinitions,verbs=get;list
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;create;update;patch;list;watch
 // +kubebuilder:rbac:groups="kubevirt.io",resources=virtualmachines,verbs=get;list;watch;create;update;patch
-var AddToManagerFuncs []func(manager.Manager, *pool_manager.PoolManager) error
+var AddToManagerFuncs []func(manager.Manager, *webhookserver.Server, *pool_manager.PoolManager) error
```
qinqon: You need to have AddToManagerFuncs and AddToWebhookFuncs

RamLavi: Do you mean change the name to AddToWebhookFuncs?

qinqon: Yep, since this ones are just adding it to webhook.

config/default/manager/manager.yaml
```diff
@@ -103,3 +103,71 @@ spec:
...
+  ports:
+    - port: 443
+      targetPort: 8000
+  selector:
+    kubemacpool-leader: "true"
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: mutator
```
qinqon: Add a comment talking about kustomize adding kubemacpool prefix and also tha caBundle is going to be updated by the kube-admission-webhook library.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.11 as builder
+FROM golang:1.12.12 as builder
```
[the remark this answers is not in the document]
RamLavi: I will add it as part of future grooming.

pkg/webhook/pod/pod.go
```diff
@@ -18,95 +18,69 @@ package pod
...
 	pod := &corev1.Pod{}
 
 	err := a.decoder.Decode(req, pod)
 	if err != nil {
-		return admission.ErrorResponse(http.StatusBadRequest, err)
+		return admission.Errored(http.StatusBadRequest, err)
 	}
 
 	if pod.Annotations == nil {
-		admission.PatchResponse(pod, pod)
```
phoracek: This should not be removed. Here we do early return in case there are no annotations (no extra network requested). In your PR you initialize the annotation.

pkg/webhook/pod/pod.go
```diff
@@ -18,95 +18,69 @@ package pod
...
+		pod.Annotations = map[string]string{}
 	}
 
-	copyPod := pod.DeepCopy()
```
phoracek: How comes we don't need to copy now?

pkg/webhook/virtualmachine/virtualmachine.go
```diff
@@ -18,90 +18,76 @@ package virtualmachine
...
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"reflect"
 
-	"k8s.io/apimachinery/pkg/api/errors"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
+	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager"
```
phoracek: Internal imports should be in a separate group after the third-party ones.

pkg/webhook/virtualmachine/virtualmachine.go
```diff
@@ -18,90 +18,76 @@ package virtualmachine
...
 	virtualMachine := &kubevirt.VirtualMachine{}
 
 	err := a.decoder.Decode(req, virtualMachine)
 	if err != nil {
-		return admission.ErrorResponse(http.StatusBadRequest, err)
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	if virtualMachine.Annotations == nil {
+		virtualMachine.Annotations = map[string]string{}
```
phoracek: Same comments as in the Pod Handle apply here

pkg/webhook/webhook.go
```diff
@@ -19,25 +19,24 @@ package webhook
...
-
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-	runtimewebhook "sigs.k8s.io/controller-runtime/pkg/webhook"
-	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
 
-	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
-	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager"
+const (
+	mutatingWebhookConfigurationName = "kubemacpool-mutator"
```
phoracek: We don't need this to be a constant. Neither is the port number

hack/vendor-tools.sh
```diff
@@ -11,3 +11,4 @@ for tool in $tools; do
     go install $tool
 done
 go mod vendor
+
```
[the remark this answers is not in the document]
RamLavi: generally speaking, when is a new line needed and it isn't?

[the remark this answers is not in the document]
phoracek: The end of the file should be always: 'blablabla\n'. Meaning it should end with exactly one newline. No other whitespace

pkg/names/names.go
```diff
@@ -8,7 +8,7 @@ const WEBHOOK_SERVICE = "kubemacpool-service"
 
 const MUTATE_WEBHOOK = "kubemacpool-webhook"
 
-const MUTATE_WEBHOOK_CONFIG = "kubemacpool"
+const MUTATE_WEBHOOK_CONFIG = "kubemacpool-mutator"
```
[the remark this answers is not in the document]
RamLavi: yes. the name define is used in the project ([example](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/599529e12198a7c3e0f7aef5bdb60fc514ad045b/pkg/webhook/webhook.go#L114))

pkg/webhook/pod/pod.go
RamLavi: I didn't see an early return in the master branch ([permalink](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/599529e12198a7c3e0f7aef5bdb60fc514ad045b/pkg/webhook/pod/pod.go#L78))
Nonetheless I will add one.

pkg/webhook/pod/pod.go
RamLavi: sticking to the [example](https://github.com/kubernetes-sigs/controller-runtime/blob/703858ff1f37d7bf08d48653d1821a0a9fba9c77/examples/builtins/mutatingwebhook.go#L51) from controller-runtime project - to my understanding, json.Marshal() returns the jason equivalent of the pod object, which is needed for PatchResponseFromRaw()
in that sense, marshaledPod is a "copy" of pod. 
Do you agree?

pkg/webhook/webhook.go
RamLavi: agree regarding the constant.
Regarding the port number - [webhookserver.New()](https://github.com/qinqon/kube-admission-webhook/blob/48024d9331d572a8335862974211784b76e7ad24/pkg/webhook/server/server.go#L32) function uses default port of 8443.
Shouldn't it be the same as the port of the service, described in [CreateOwnerRefForService](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/599529e12198a7c3e0f7aef5bdb60fc514ad045b/pkg/webhook/webhook.go#L149)()? (this is the reason I changed the port to 8000 here)

pkg/webhook/pod/pod.go
phoracek: Sounds right

pkg/webhook/webhook.go
phoracek: Good that their match I think. Could you use a constant for this since it is shared across multiple places?

pkg/webhook/pod/pod.go
phoracek: Thanks!

RamLavi: Adding early return fails the tests. currently dropping this change and will revisit it in another commit.

pkg/manager/manager.go
```diff
@@ -23,6 +23,8 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	poolmanager "github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager"
+	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/webhook"
```
phoracek: Internal imports should be in the third group (as they were before this change)

PR #86, opened 2020-02-03
update used kubevirtci version

phoracek:
Let's use the same mechanism as other projects using kubevirtci to obtain
the project and interact with it. This should make it easier to maintain
kubevirtci integration. Bumping kubevirtci release and Kubernetes provider
version should be as easy as changing a single variable.


phoracek:
cc @RamLavi

PR #87, opened 2020-02-03
Use go mod

RamLavi:
migrating to go modules dependency management tool


tool.go
```diff
@@ -0,0 +1,8 @@
+// +build tools
```
phoracek: Make sure to run gofmt on this one. Also, please add a newline at the end.

hack/vendor-tools.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -xe
+export GOFLAGS=-mod=vendor
```
phoracek: These two are already exported by the Makefile, you can drop them.

RamLavi:
@phoracek, why did you merge if I didn't fix the comments yet ? was is by mistake?

phoracek:
Yeah, I was merging from my phone and forgot about those. It's just a nit, we can address it later.

PR #88, opened 2020-02-03
add automation scripts

phoracek:
These can be used to run unit/functional tests on an arbitrary host
with docker and rsync installed. They will be later used for CI triggered
by prow.


phoracek:
cc @RamLavi

PR #89, opened 2020-02-03
drop dockerized test

phoracek:
We can run the tests with Go alone, there is no need to containerize them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 sudo: required
 dist: trusty
-go: "1.11"
+go: "1.12"
 
 services:
 - docker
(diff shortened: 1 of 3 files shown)


phoracek:
cc @RamLavi

RamLavi:
if we are not using docker run then the test is performed on the actual environment? if not, where? why is it better?

phoracek:
We still run functional tests (which change environment) on the local cluster. For unit tests, we used to run them in a container, so all the Go dependencies are taken from the image and not needed on your environment. That is however not needed, we can run just fine with Go alone.

RamLavi:
ok, does this also eliminate the use of builder image?

phoracek:
Not yet, there are other users of it. But eventually we'll get rid of it. Especially with https://github.com/k8snetworkplumbingwg/kubemacpool/pull/93

PR #90, opened 2020-02-03
 bump versions of used CNAO and kubevirt 

phoracek:


phoracek:
cc @RamLavi

hack/deploy-test-cluster.sh
```diff
@@ -6,15 +6,27 @@ set -ex
...
-./cluster/dind-cluster/kubectl create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.15.0/namespace.yaml
-./cluster/dind-cluster/kubectl create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.15.0/network-addons-config.crd.yaml
-./cluster/dind-cluster/kubectl create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.15.0/operator.yaml
-./cluster/dind-cluster/kubectl create -f ./hack/cna/cna-cr.yaml
-
-# Deploy Kubevirt
-./cluster/dind-cluster/kubectl create -f https://github.com/kubevirt/kubevirt/releases/download/v0.20.4/kubevirt-operator.yaml
-./cluster/dind-cluster/kubectl create -f https://github.com/kubevirt/kubevirt/releases/download/v0.20.4/kubevirt-cr.yaml
+
+echo 'Deploying Linux bridge CNI and Multus ...'
```
phoracek: Please note that this will go away as soon as we move to prow (we will drop dind)

cluster/up.sh
```diff
@@ -21,37 +21,25 @@ kubevirtci::install
...
+echo 'Deploying Linux bridge CNI and Multus ...'
+./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.25.0/namespace.yaml
+./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.25.0/network-addons-config.crd.yaml
+./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.25.0/operator.yaml
+cat <<EOF | ./cluster/kubectl.sh create -f -
+apiVersion: networkaddonsoperator.network.kubevirt.io/v1alpha1
+kind: NetworkAddonsConfig
+metadata:
+  name: cluster
+spec:
```
RamLavi: In the original hack/cna/cna-cr.yaml we also deploy imagePullPolicy. why did you drop it?

hack/cni-plugins/cni-plugins.yaml
```diff
@@ -1,43 +0,0 @@
----
```
RamLavi: why are we dropping this yaml?

hack/multus/kubernetes-multus.yaml
```diff
@@ -1,25 +0,0 @@
----
-kind: ConfigMap
```
RamLavi: why are we dropping hack/multus yamls?

cluster/up.sh
phoracek: We don't need to pull it Always, we are ok with whatever Multus and bridge CNI is given to us. And OnlyIfNeeded is the default.

hack/cni-plugins/cni-plugins.yaml
phoracek: CNI plugins are shipped by CNAO, this is just a remnant of the past.

hack/multus/kubernetes-multus.yaml
phoracek: Multus shipped by CNAO, this is just a remnant of the past.

cluster/up.sh
```diff
@@ -21,37 +21,25 @@ kubevirtci::install
...
 $(kubevirtci::path)/cluster-up/up.sh
 
-# Deploy CNA
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.15.0/namespace.yaml
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.15.0/network-addons-config.crd.yaml
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.15.0/operator.yaml
-./cluster/kubectl.sh create -f ./hack/cna/cna-cr.yaml
-
-# wait for cluster operator
+echo 'Deploying Linux bridge CNI and Multus ...'
```
RamLavi: why do we need multus and linux bridge deployed? is it so we have "customers" to give MACs to?

cluster/up.sh
```diff
@@ -21,37 +21,25 @@ kubevirtci::install
...
+./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.25.0/namespace.yaml
+./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.25.0/network-addons-config.crd.yaml
+./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/0.25.0/operator.yaml
+cat <<EOF | ./cluster/kubectl.sh create -f -
```
RamLavi: I noticed that you generally change from using yaml files to heredocs, could you explain why you prefer this?

cluster/up.sh
phoracek: Yes. In e2e tests we connect Pods and VMs to secondary networks via Multus and the bridge CNI.

cluster/up.sh
phoracek: Less jumps between files. Once we drop dind, all the dependencies would be here. But if you prefer it being in a separate file, I don't mind that much.

PR #91, opened 2020-02-03
 fix vendoring by pinning to kubevirt 0.25.0 

phoracek:
Without these changes, go mod tidy was failing on an OpenShift dependency.


phoracek:
cc @RamLavi

PR #93, opened 2020-02-03
include go in the project

phoracek:
In order not to get different results on different systems, let's include
specific Go binary under the project with custom GOROOT.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,20 +1,31 @@
-
 # Image URL to use all building/pushing image targets
 REGISTRY ?= quay.io
 IMAGE_TAG ?= latest
+IMG ?= kubevirt/kubemacpool
+
+BIN_DIR = $(CURDIR)/build/_output/bin/
+
 export GOFLAGS=-mod=vendor
 export GO111MODULE=on
-
-IMG ?= kubevirt/kubemacpool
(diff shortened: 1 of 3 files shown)


phoracek:
cc @RamLavi @qinqon

Makefile
```diff
@@ -1,20 +1,31 @@
...
+
+BIN_DIR = $(CURDIR)/build/_output/bin/
+
 export GOFLAGS=-mod=vendor
 export GO111MODULE=on
-
-IMG ?= kubevirt/kubemacpool
+export GOROOT=$(BIN_DIR)/go/
+export GOBIN=$(GOROOT)/bin/
+export PATH := $(GOROOT)/bin:$(PATH)
```
qinqon: We can use GOBIN directly here.

qinqon:
@phoracek found an alternative way to do this using "go get" https://github.com/qinqon/kube-admission-webhook/blob/master/hack/install-go.sh

phoracek:
@qinqon are you using `go` in order to install `go`? :D Doesn't it defy the purpose?

phoracek:
This fails only on the one problematic test. I don't believe it is a regression. @RamLavi what do you think?

PR #94, opened 2020-02-17
adding a readiness probe to kubemacpool pods

RamLavi:
Currently, manager pods turn Ready as soon as their binary starts.
However, it takes ~40 more seconds for the webhook to actually start
serving requests. In order to avoid that, we introduce a readiness
probe that marks the pod as Ready only once the server starts.

### DIFF
--- a/cluster/sync.sh
+++ b/cluster/sync.sh
@@ -35,9 +35,25 @@ done
 
 ./cluster/kubectl.sh apply -f ./config/test/kubemacpool.yaml
 
-# Make sure all containers are ready
-while [ -n "$(./cluster/kubectl.sh get pods --all-namespaces -o'custom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers | grep false)" ]; do
-    echo "Waiting for all containers to become ready ..."
-    ./cluster/kubectl.sh get pods --all-namespaces -o'custom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers
-    sleep 10
-done
+pods_ready_wait() {
+  echo "Waiting for non-kubemacpool containers to be ready ..."
+  ./cluster/kubectl.sh wait pod --all -n kube-system --for=condition=Ready --timeout=5m
+
+  wait_failed=''
+  max_retries=12
+  retry_counter=0
+  echo "Waiting for kubemacpool leader container to be ready ..."
+  while [[ "$(./cluster/kubectl.sh wait pod -n kubemacpool-system --for=condition=Ready -l kubemacpool-leader=true --timeout=1m)" = $wait_failed ]] && [[ $retry_counter -lt $max_retries ]]; do
+    sleep 5s
+    retry_counter=$((retry_counter + 1))
+  done
+
+  if [ $retry_counter -eq $max_retries ]; then
+    echo "Failed/timed-out waiting for kubemacpool recourses"
+    exit 1
+  else
+    echo "Kubemacpool leader pod is ready"
+  fi
+}
+
+pods_ready_wait
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"fmt"
+	"math"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/rand"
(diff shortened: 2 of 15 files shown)


tests/tests.go
```diff
@@ -206,6 +205,28 @@ func DeleteLeaderManager() {
...
+			if len(currentPod.Status.ContainerStatuses) >= 1 && currentPod.Status.ContainerStatuses[0].Ready {
+				return nil
+			}
+		}
+
+		return fmt.Errorf("have not found any manager in the ready state")
+	}, 2*time.Minute, 3*time.Second).ShouldNot(HaveOccurred(), "timed out while waiting for a new leader")
+}
+
+func min(x, y int32) int32 {
```
qinqon: use https://golang.org/pkg/math/#Min

tests/tests.go
```diff
@@ -141,42 +141,41 @@ func setRange(rangeStart, rangeEnd string) error {
...
 		}
 
-		if len(podsList.Items) != int(macDeploy.Status.Replicas) {
-			return fmt.Errorf("should have %v manager pods", macDeploy.Status.Replicas)
+		for _, currentPod := range currentPods.Items {
+			for _, oldPod := range oldPods.Items {
+				if currentPod.Name == oldPod.Name {
+					return fmt.Errorf("old pod %s has not yet been removed", oldPod.Name)
+				}
+			}
```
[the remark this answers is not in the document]
RamLavi: you sure? I can think of an example where (pod names swapped numbers for ease) oldPods={1,2,4}; currentPods={2,3,5}. 
In my example, both lists have size of 3, and yet the pod named "2" is in both lists. 
Yet `if currentPods.Items[i].Name == oldPods.Items[i].Name` wont find it I'm afraid. I see no alternative to nested search for all permutations. what do you think?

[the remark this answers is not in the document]
qinqon: Ahh ok, elegant way is use sets although golang does not have those, I am ok with that, but better put this in a function with good name like findPod so you just iterate on the oldPod and call findPod on the cuirrentPod (it's a nested for loop in disguise).

tests/tests.go
```diff
@@ -141,42 +151,39 @@ func setRange(rangeStart, rangeEnd string) error {
...
-		podsList, err = testClient.KubeClient.CoreV1().Pods(ManagerNamespce).List(metav1.ListOptions{})
+		currentPods, err := testClient.KubeClient.CoreV1().Pods(ManagerNamespce).List(metav1.ListOptions{})
 		if err != nil {
 			return err
 		}
 
-		if len(podsList.Items) != int(macDeploy.Status.Replicas) {
-			return fmt.Errorf("should have %v manager pods", macDeploy.Status.Replicas)
+		for _, oldPod := range oldPods.Items {
+			if findPodByName(currentPods, oldPod) == true {
```
qinqon: If you use a pointer you compare to != nil

cluster/sync.sh
```diff
@@ -35,9 +35,16 @@ done
...
-    ./cluster/kubectl.sh get pods --all-namespaces -o'custom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers
-    sleep 10
-done
+pods_ready_wait() {
+  echo "Waiting for non-kubemacpool containers to be ready ..."
+  ./cluster/kubectl.sh wait pod --all -n kube-system --for=condition=Ready --timeout=5m
+
+  echo "Waiting for (TBD) to be ready ..."
+#  cluster/kubectl.sh wait --for=condition=available -f ./config/test/kubemacpool.yaml --timeout=3m
+#  cluster/kubectl.sh wait -n kubemacpool-system --for=condition=available deployment.apps/kubemacpool-mac-controller-manager --timeout=2m
```
RamLavi: I did not manage to find a proper kubectl wait to be satisfied when the leader is ready.
the issue is that during make cluster-sync `./cluster/kubectl.sh wait pod -n kubemacpool-system --for=condition=Ready -l kubemacpool-leader=true --timeout=2m` receives error: no matching resources found.

if I sleep 5s then the former wait waits and eventually succeeds. 
Do you have any suggestions?

phoracek:
@RamLavi now when Prow runs e2e and unit tests, we should fix them before we continue with merging

tests/virtual_machines_test.go
```diff
@@ -489,7 +489,7 @@ var _ = Describe("Virtual Machines", func() {
 			})
 		})
 		//2995
```
phoracek: Please link a GitHub Issue where you describe the failure and attach the error log

PR #95, opened 2020-02-23
Use kubevirtci instead of dind

RamLavi:
Currently, kubemacpool project includes both dind and kubevirtci. The usage of both of these is unfortunately quite messy. Travis is using dind, while the Makefile includes untested kubevirtci.

Let's align kubemacpool with our other projects and use kubevirtci only. By doing this switch, we will obtain unified way to run tests on kubemacpool. It will make maintenance and integration to prow easier.


Makefile
```diff
@@ -14,6 +14,14 @@ export PATH := $(GOBIN):$(PATH)
 GOFMT := $(GOBIN)/gofmt
 GO := $(GOBIN)/go
 
+CLUSTER_DIR ?= cluster
```
phoracek: We don't need this change. All the scripts are already in the ./cluster/ directory. Is there any reason to parametrize those?

Makefile
```diff
@@ -14,6 +14,14 @@ export PATH := $(GOBIN):$(PATH)
...
+KUBECONFIG ?= kubevirtci/_ci-configs/$(KUBEVIRT_PROVIDER)/.kubeconfig
+export KUBECTL ?= $(CLUSTER_DIR)/kubectl.sh
```
phoracek: This is the only one I would keep.

Makefile
```diff
@@ -78,19 +86,16 @@ docker-builder:
...
-	./cluster/up.sh
-
-cluster-down:
-	./cluster/down.sh
+cluster-up: $(CLUSTER_UP)
+	$(CLUSTER_UP)
 
-cluster-sync:
-	./cluster/sync.sh
+cluster-down: $(CLUSTER_DOWN)
```
phoracek: No need to specify the dependency. cluster/ directory is part of the repo.

Makefile
```diff
@@ -78,19 +86,16 @@ docker-builder:
...
+	$(CLUSTER_DOWN)
 
-deploy-test-cluster:
-	./hack/deploy-test-cluster.sh
+cluster-sync: $(CLUSTER_SYNC)
```
phoracek: ditto

Makefile
```diff
@@ -29,11 +37,11 @@ functest: $(GO)
 	GO=$(GO) ./hack/functest.sh
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-deploy: generate-deploy
-	kubectl apply -f config/test/kubemacpool.yaml
+deploy: $(KUBECTL) generate-deploy
```
qinqon: $(KUBECTL) is not longer a dependency

Makefile
```diff
@@ -29,11 +37,11 @@ functest: $(GO)
...
+	$(KUBECTL) apply -f config/test/kubemacpool.yaml
 
-deploy-test: generate-test
-	kubectl apply -f config/test/kubemacpool.yaml
+deploy-test: $(KUBECTL) generate-test
```
qinqon: ditto

Makefile
```diff
@@ -78,19 +86,16 @@ docker-builder:
 docker-push:
 	docker push ${REGISTRY}/${IMG}:${IMAGE_TAG}
 
-cluster-up:
-	./cluster/up.sh
-
-cluster-down:
-	./cluster/down.sh
+cluster-up: $(CLUSTER_UP)
```
qinqon: dep not needed

Makefile
RamLavi: just because it's used more than once, then I added it for easier future changes.

PR #107, opened 2020-03-12
[release-v0.8] Skip pods in critical namespaces

phoracek:
This PR fix the issue when a Critical pod like the sdn provider
need to be recreated.

change the mutating webhook to skip namespaces to contains

1. runlevel=[0,1] for kubernetes environments
2. openshift.io/run-level=[0,1] for openshift environments

With this change we can change the FailurePolicy from ignore to fail.

Signed-off-by: Sebastian Sch <sebassch@gmail.com>

### DIFF
--- a/tests/pods_test.go
+++ b/tests/pods_test.go
@@ -1,16 +1,97 @@
 package tests
 
 import (
+	"context"
+	"fmt"
+	"github.com/onsi/gomega/types"
+	"strings"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 )
 
+const defaultNumberOfReplicas = 2
+
 var _ = Describe("Pods", func() {
-	Context("Check the client", func() {
-		It("should not fail", func() {
-			_, err := testClient.KubeClient.CoreV1().Pods("").List(v1.ListOptions{})
+	Context("Check the pod mutating webhook", func() {
+		AfterEach(func() {
+			// Clean pods from our test namespaces after every test to start clean
+			for _, namespace := range []string{TestNamespace, OtherTestNamespace} {
+				podList := &corev1.PodList{}
+				err := testClient.VirtClient.List(context.TODO(), &client.ListOptions{Namespace: namespace}, podList)
+				Expect(err).ToNot(HaveOccurred())
+
+				for _, podObject := range podList.Items {
+					err = testClient.VirtClient.Delete(context.TODO(), &podObject)
+					Expect(err).ToNot(HaveOccurred())
+				}
+
+				Eventually(func() int {
+					podList := &corev1.PodList{}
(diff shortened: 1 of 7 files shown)


tests/tests.go
```diff
@@ -194,6 +216,89 @@ func DeleteLeaderManager() {
...
 }
 
+func changeManagerReplicas(numOfReplica int32) error {
+	Eventually(func() error {
+		managerDeployment, err := testClient.KubeClient.AppsV1().Deployments(ManagerNamespce).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		managerDeployment.Spec.Replicas = &numOfReplica
```
RamLavi: I'm not sure, but if numOfReplica is int32, then why are we using reference operator?

phoracek: Type of Replicas is *int32 to allow nil as the default AFAIK

tests/pods_test.go
```diff
@@ -1,16 +1,97 @@
...
 var _ = Describe("Pods", func() {
-	Context("Check the client", func() {
-		It("should not fail", func() {
-			_, err := testClient.KubeClient.CoreV1().Pods("").List(v1.ListOptions{})
+	Context("Check the pod mutating webhook", func() {
+		AfterEach(func() {
+			// Clean pods from our test namespaces after every test to start clean
+			for _, namespace := range []string{TestNamespace, OtherTestNamespace} {
+				podList := &corev1.PodList{}
+				err := testClient.VirtClient.List(context.TODO(), &client.ListOptions{Namespace: namespace}, podList)
```
RamLavi: Are we sure this is the correct parameter order? in [master](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/75371d2d5442ac33f56a8500f58d6bdce19805d5/tests/pods_test.go#L26) branch it's deifferent

tests/tests.go
RamLavi: yeah my bad thanks.

tests/pods_test.go
```diff
@@ -1,16 +1,97 @@
 package tests
 
 import (
+	"context"
+	"fmt"
+	"github.com/onsi/gomega/types"
```
SchSeba: nit: please move this on the gomega section

tests/pods_test.go
phoracek: Interesting.

The `List` signature on this branch is `List(ctx context.Context, opts *ListOptions, list runtime.Object) error` https://github.com/k8snetworkplumbingwg/kubemacpool/blob/release-v0.8/vendor/sigs.k8s.io/controller-runtime/pkg/client/interfaces.go#L54

On master it is `List(ctx context.Context, list runtime.Object, opts ...ListOption) ` https://github.com/k8snetworkplumbingwg/kubemacpool/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/client/interfaces.go#L61

Seems that it changed between controller-runtime versions. Good catch!

phoracek:
/override  pull-kubemacpool-e2e-k8s
/override pull-kubemacpool-unit-test
/hold

These lanes are not available on release-v0.8.

This passed `make functest` on kubevirtci cluster for me locally except for a single test `should fail to create a pod on a regular namespace when mac pool is down` the same issue happens on Travis.

phoracek:
/override pull-kubemacpool-unit-test
/hold

These lanes are not available on release-v0.8.

PR #109, opened 2020-03-29
WIP change kubemacpool to work with 2 pods in parallel

RamLavi:
currently all kmp pods perform a leader election and only the leader performs the MAC address handling. The leader pod is assigned with an appropriate label.

In order to allow 2 kubemacpool managers to work simultaneously, we should follow the [design document](https://docs.google.com/document/d/1E_L-Bi-bEO7kwj6hHxLV8Wbed0s_GYtO1qRyr4YaDZg/edit)


config/default/manager/manager.yaml
```diff
@@ -21,14 +21,15 @@ data:
...
-kind: Deployment
+kind: StatefulSet
 metadata:
   name: mac-controller-manager
   namespace: system
   labels:
     control-plane: mac-controller-manager
     controller-tools.k8s.io: "1.0"
 spec:
+  serviceName: "kubemacpool-service"
```
RamLavi, line 32: is this necessary? I understand that the service is already attached by the label selector

tests/tests.go
```diff
@@ -188,46 +187,31 @@ func setRange(rangeStart, rangeEnd string) error {
 	return nil
 }
 
-func DeleteLeaderManager() {
+func DeleteOneManager() {
```
alonSadan, line 190: not sure about this, but maybe we should add to test a part that verifies that even though we deleted one manager, we continue to serve, keeps the functionality or something like that.
for example, in the previous revision, when there was a leader, after the deletion, there is a verification that there is still a healthy pod in the manager namespace (https://github.com/k8snetworkplumbingwg/kubemacpool/pull/109/commits/7e983b7906ba77a768f0618d3166760391440a5a#diff-4de094b4f2274c29e9f11102da4ca7d0L217-L230)
while here we just delete kubemacpool pod and...that's it. 
without it I'm not sure this test is needed, it is a bit "meat-less"

pkg/webhook/webhook.go
```diff
@@ -132,12 +132,12 @@ func CreateOwnerRefForService(kubeClient *kubernetes.Clientset, managerNamespace
...
+	managerDeployment, err := kubeClient.AppsV1().StatefulSets(managerNamespace).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
 	if err != nil {
 		return nil, err
 	}
-	ownerRefList := []metav1.OwnerReference{{Name: managerDeployment.Name, Kind: "Deployment", APIVersion: "apps/v1", UID: managerDeployment.UID}}
+	ownerRefList := []metav1.OwnerReference{{Name: managerDeployment.Name, Kind: "StatefulSet", APIVersion: "apps/v1", UID: managerDeployment.UID}}
 
 	return ownerRefList, nil
-}
+}
```
alonSadan, line 143: remove newline?

config/default/manager/manager.yaml
ormergi: serviceName is mandatory for StatefulSets since apps/v1, although it is not mentioned explicitly in k8s documentation, if you dont specify it you get validation error:
`alidating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1.StatefulSetSpec;`
This guy encountered that same error as I did:
https://github.com/kubernetes/kubernetes/issues/69608#issuecomment-590958248

ormergi: serviceName is mandatory for StatefulSets  although it is not mentioned explicitly in k8s documentation, if you dont specify it you get validation error:
alidating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1.StatefulSetSpec;

PR #112, opened 2020-04-02
Deploy kubemacpool-mac-controller-manager with StatefulSet

ormergi:
This PR change the deployment strategy from deploying kubemacpool-mac-controller-manager pod with Deployment to StatufSet.
As part of removing the leader-election mechanism effort #109 , it is necessary to enumerate the `kubemacpool-mac-controller-manager` pod names, with StatfulSet it is out-of-the-box functionality.

### DIFF
--- a/config/default/manager/manager.yaml
+++ b/config/default/manager/manager.yaml
@@ -21,7 +21,7 @@ data:
   RANGE_END: 02:FF:FF:FF:FF:FF
 ---
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
 metadata:
   name: mac-controller-manager
   namespace: system
@@ -29,6 +29,7 @@ metadata:
     control-plane: mac-controller-manager
     controller-tools.k8s.io: "1.0"
 spec:
+  serviceName: kubemacpool-service
   replicas: 2
   selector:
     matchLabels:
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -133,7 +133,7 @@ func CreateOwnerRefForService(kubeClient *kubernetes.Clientset, managerNamespace
 }
 
 func createStatefulSetOwnerRef(kubeClient *kubernetes.Clientset, managerNamespace string) ([]metav1.OwnerReference, error) {
-	managerStatefulset, err := kubeClient.AppsV1().StatefulSets(managerNamespace).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
+	managerStatefulset, err := kubeClient.AppsV1().StatefulSets(managerNamespace).Get(names.MANAGER_STATEFULSET, metav1.GetOptions{})
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -45,7 +45,7 @@ const (
 // +kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups="apiextensions.k8s.io",resources=customresourcedefinitions,verbs=get;list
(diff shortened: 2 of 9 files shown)


pkg/webhook/webhook.go
```diff
@@ -132,12 +132,12 @@ func CreateOwnerRefForService(kubeClient *kubernetes.Clientset, managerNamespace
...
+	managerDeployment, err := kubeClient.AppsV1().StatefulSets(managerNamespace).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
 	if err != nil {
 		return nil, err
 	}
-	ownerRefList := []metav1.OwnerReference{{Name: managerDeployment.Name, Kind: "Deployment", APIVersion: "apps/v1", UID: managerDeployment.UID}}
+	ownerRefList := []metav1.OwnerReference{{Name: managerDeployment.Name, Kind: "StatefulSet", APIVersion: "apps/v1", UID: managerDeployment.UID}}
 
 	return ownerRefList, nil
-}
+}
```
alonSadan, line 143: add new line to drop this red sign

config/default/manager/manager.yaml
```diff
@@ -21,14 +21,15 @@ data:
...
-kind: Deployment
+kind: StatefulSet
 metadata:
   name: mac-controller-manager
   namespace: system
   labels:
     control-plane: mac-controller-manager
     controller-tools.k8s.io: "1.0"
 spec:
+  serviceName: "kubemacpool-service"
```
RamLavi, line 32: why did you use ""?

ormergi: this is string type
https://github.com/kubernetes/api/blob/master/apps/v1/types.go#L152

qinqon:
/hold 

Can we approve this ?

alonSadan:
I think that there are some http requests that needs to be changed, found some in the tests.go file like here : 
https://github.com/k8snetworkplumbingwg/kubemacpool/blob/04b39a173189e9e5eb5c1b6b2f9f3311ecf1ba47/tests/tests.go#L235

for example change : 
`managerDeployment, err := testClient.KubeClient.AppsV1().Deployments (ManagerNamespce).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})`

to 
`managerDeployment, err := testClient.KubeClient.AppsV1().StatefulSets(ManagerNamespce).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})`

tests/tests.go
```diff
@@ -232,14 +236,14 @@ func DeleteLeaderManager() {
 
 func changeManagerReplicas(numOfReplica int32) error {
 	Eventually(func() error {
-		managerDeployment, err := testClient.KubeClient.AppsV1().Deployments(ManagerNamespce).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
+		managerStatefulset, err := testClient.KubeClient.AppsV1().StatefulSets(ManagerNamespce).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
```
RamLavi, line 239: please change names.MANAGER_DEPLOYMENT const to names.MANAGER_STATEFULSET

tests/tests.go
```diff
@@ -232,14 +236,14 @@ func DeleteLeaderManager() {
 
 func changeManagerReplicas(numOfReplica int32) error {
```
RamLavi, line 237: please change the last Eventually's error description from: "failed to change kubemacpool deployment number of replicas" to "failed to change kubemacpool statefulset number of replicas"

tests/tests.go
ormergi: As far as I understand, this const represent the manager deployment name, and by deployment I mean the action but not the method (whether it is done with Deployment, StatfulSet or any other kind).
We can change it to MANAGER_PODS_NAME,
because eventually the name of the pods will be  derivative of this const value, or just leave it.
What you think?

RamLavi: the pod's name is derived from the statefulset/deployment, so I wouldn't change it to that.

I rechecked and names.MANAGER_DEPLOYMENT is currently only used to retrieve the statefulset object. 

I think changing it might reduce future confusion, but if you're adamant then I leave the call to you.

PR #113, opened 2020-04-07
[sharding] WIP change kubemacpool to work with 2 pods in parallel

RamLavi:
The goal of this PR is to introduce an basic active-active approach to kmp, in order to reduce availability downtime. We will follow the [design document](https://docs.google.com/document/d/1E_L-Bi-bEO7kwj6hHxLV8Wbed0s_GYtO1qRyr4YaDZg/edit).

currently all kmp pods perform a leader election and only the leader performs the MAC address handling. The leader pod is assigned with an appropriate label.
In the new state, the 2 kubemacpool pods will work together on different ranges, allowing for better availability.
The PR should be declared done and merged once the kubemacpool passes all relevant tests. For this, we may need to rebase the PR (link TBD) for splitting the ranges.


cluster/sync.sh
```diff
@@ -42,8 +42,8 @@ pods_ready_wait() {
   wait_failed=''
   max_retries=12
   retry_counter=0
-  echo "Waiting for kubemacpool leader container to be ready ..."
-  while [[ "$(./cluster/kubectl.sh wait pod -n kubemacpool-system --for=condition=Ready -l kubemacpool-leader=true --timeout=1m)" = $wait_failed ]] && [[ $retry_counter -lt $max_retries ]]; do
+  echo "Waiting for kubemacpool recourses to be ready ..."
+  while [[ "$(./cluster/kubectl.sh wait pod -n kubemacpool-system --all --for=condition=Ready --timeout=1m)" = $wait_failed ]] && [[ $retry_counter -lt $max_retries ]]; do
```
phoracek, line 46: `--all` means all resources, you are waiting for `pod` specifically, so you can drop `--all`

RamLavi: actually it's mandatory, since I want to hit both pods without specific name, but I think doing it by label will be better. something like
`kubectl.sh wait pod -l app=kubemacpool -n kubemacpool-system --for=condition=Ready --timeout=1m`

ormergi:
Shouldn't this [1] be removed?

[1] https://github.com/k8snetworkplumbingwg/kubemacpool/blob/5bf533ca47ce26fa42475715b085bfe7acca7b73/pkg/manager/manager.go#L54

pkg/pool-manager/pool.go
```diff
@@ -147,6 +150,15 @@ func (p *PoolManager) InitMaps() error {
 	return nil
 }
 
+func (p *PoolManager) IsMacInRange(macAddress string) bool {
+	macAddressHW, _ := net.ParseMAC(macAddress)
```
ormergi, line 154: I think its better to keep the parsing outside of this kind of function, what you think?

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -131,7 +134,8 @@ func (p *PoolManager) UpdateMacAddressesForVirtualMachine(previousVirtualMachine
 	log.V(1).Info("UpdateMacAddressesForVirtualMachine: data",
 		"macmap", p.macPoolMap,
 		"podmap", p.podToMacPoolMap,
-		"currentMac", p.currentMac.String())
+		"currentMac", p.currentMac.String(),
+		"virtualMachine.Name", virtualMachine.Name)
 	if previousVirtualMachine == nil {
 		p.poolMutex.Unlock()
```
ormergi, line 140: I think this can be right next to Lock(),  either way Unlock() is called when this function returns

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -346,26 +346,33 @@ func (p *PoolManager) isRelatedToKubevirt(pod *corev1.Pod) bool {
 	return false
 }
 
-func (p *PoolManager) releaseMacAddressesFromInterfaceMap(allocations map[string]string) {
-	for _, value := range allocations {
-		delete(p.macPoolMap, value)
+// changes mac's macPoolMap status to be deleted in controller context.
+func (p *PoolManager) MarkMacAddressesForRelease(allocations map[string]string) {
+	for _, macAddress := range allocations {
+		if _, exist := p.macPoolMap[macAddress]; p.IsMacInRange(macAddress) && exist {
```
ormergi, line 352: Is it necessary to check if a mac is in the range if it is already in the macPoolMap?

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -346,26 +346,33 @@ func (p *PoolManager) isRelatedToKubevirt(pod *corev1.Pod) bool {
...
+		}
 	}
 }
 
 // Revert allocation if one of the requested mac addresses fails to be allocated
 func (p *PoolManager) revertAllocationOnVm(vmName string, allocations map[string]string) {
 	log.V(1).Info("Revert vm allocation", "vmName", vmName, "allocations", allocations)
-	p.releaseMacAddressesFromInterfaceMap(allocations)
+	for _, value := range allocations {
+		if p.IsMacInRange(value) {
```
ormergi, line 362: same as https://github.com/k8snetworkplumbingwg/kubemacpool/pull/113/files#r411313430

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -389,42 +396,69 @@ func (p *PoolManager) AddMacToWaitingConfig(allocations map[string]string) error
 	}
 
 	for _, macAddress := range allocations {
-		log.V(1).Info("add mac address to waiting config", "macAddress", macAddress)
-		macAddress = strings.Replace(macAddress, ":", "-", 5)
-		configMap.Data[macAddress] = time.Now().Format(time.RFC3339)
+		if p.IsMacInRange(macAddress) {
```
ormergi, line 399: Is it necessary to check it the address is in range as this point? 
The validation is already done right?

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -221,14 +225,10 @@ func (p *PoolManager) allocateRequestedVirtualMachineInterfaceMac(virtualMachine
 		return err
 	}
 
-	if _, exist := p.macPoolMap[requestedMac]; exist {
-		err := fmt.Errorf("failed to allocate requested mac address")
-		log.Error(err, "mac address already allocated")
-
-		return err
+	if p.IsMacInRange(requestedMac) {
```
ormergi, line 228: The parse checking in line 224 is redundant, it is already done inside IsMacInRange

pkg/webhook/virtualmachine/virtualmachine.go
```diff
@@ -66,7 +66,7 @@ func (a *virtualMachineAnnotator) Handle(ctx context.Context, req admission.Requ
 		virtualMachine.Namespace = req.AdmissionRequest.Namespace
 	}
 
-	log.V(1).Info("got a create virtual machine event", "virtualMachineName", virtualMachine.Name, "virtualMachineNamespace", virtualMachine.Namespace)
+	log.V(1).Info("got a virtual machine event", "virtualMachineName", virtualMachine.Name, "virtualMachineNamespace", virtualMachine.Namespace)
```
ormergi, line 69: How about moving this to line 71 and create another one for update events in line 77?

tests/tests.go
```diff
@@ -188,49 +187,34 @@ func setRange(rangeStart, rangeEnd string) error {
 	return nil
 }
 
-func DeleteLeaderManager() {
+func DeleteOneManager() {
```
ormergi, line 190: Is it "enough" to remove only one pod to cover our test cases?

tests/tests.go
```diff
@@ -2,17 +2,16 @@ package tests
...
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/rand"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
```
ormergi, line 13: Is it alright to format this imports? last time I did I have been told to revert in order to prevent diff problems

phoracek: The issue is when you reformat a file that you don't touch otherwise. That may cause collisions. It is fine (and sometimes unavoidable) to format a file that you are changing as a part of the PR.

vendor/modules.txt
```diff
@@ -320,6 +320,7 @@ k8s.io/apimachinery/pkg/runtime
 k8s.io/apimachinery/pkg/api/errors
 k8s.io/apimachinery/pkg/apis/meta/v1
 k8s.io/apimachinery/pkg/util/intstr
+k8s.io/apimachinery/pkg/types
```
ormergi, line 323: Is it ok to format this file? https://github.com/k8snetworkplumbingwg/kubemacpool/pull/113/files#r411325352

pkg/pool-manager/pool.go
RamLavi: I agree. it's redundant. will remove.

RamLavi: Since the inRange() is using byte[] type MAC Address, I need to translate the mac address, which is saved as sting on the cache maps.

since it is an internal matter, I think it's better to keep it here.

pkg/pool-manager/virtualmachine_pool.go
RamLavi: since AllocateVirtualMachineMac() also uses Lock(), we want to avoid unnecessary double locking.

pkg/pool-manager/virtualmachine_pool.go
RamLavi: yes. The mac updated can possibly be manually inserted, and as such - can also be out of range. 
since the webhook/controller only handle Macs that belong to the kmp-mmgr pod's range - then we need to check if it is in range before inserting/deleting

pkg/pool-manager/virtualmachine_pool.go
RamLavi: same answer.

pkg/pool-manager/virtualmachine_pool.go
RamLavi: in AddMacToWaitingConfig() we add new macs allocated to the vmWaitConfigMap.
Since some of them may be in the in range and some don't (manually chosen) - we need to insert only the ones that are in the kmp-mgr range.

pkg/pool-manager/virtualmachine_pool.go
RamLavi: I agree. will change

pkg/webhook/virtualmachine/virtualmachine.go
RamLavi: [mutateCreateVirtualMachinesFn](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/9494a62561c713192706d0c23b14b70dfcdff2a3/pkg/webhook/virtualmachine/virtualmachine.go#L95) and [mutateUpdateVirtualMachinesFn](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/9494a62561c713192706d0c23b14b70dfcdff2a3/pkg/webhook/virtualmachine/virtualmachine.go#L118) alreaddy have prints of their own. maybe this line is irrelevant?

tests/tests.go
RamLavi: The original test deleted the leader to check for how much time it will take to regain kmp operation. In the sharding approach we have 2 running mgr so I think the test should be changed to check the the other kmp-mgr is ready in 0 time. This is what I'm checking here.

But I agree that it may be also interesting to add another test where we kill both pods to make sure there're ready in reasonable time. I will add it to the the list of tests added in the test task.

vendor/modules.txt
RamLavi: it is auto generated after I performed go mod tidy (I updated after eliminating the leder election)

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -92,52 +93,65 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 		return reconcile.Result{}, err
 	}
 
+	podFinalizerName := pool_manager.RuntimeObjectFinalizerName + r.poolManager.GetManagerPodName()
```
ormergi, line 96: Is there a difference between r.pool_manager and pool_manager?

I think this should be done at [NewPoolManager()](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/a8c13fea660f0298ca15b946edb0afbf1ddd1341/pkg/pool-manager/pool.go#L62), 
with string formatting while removing the `/` from RuntimeObjectFinalizerName so ti will be part of the string format
`fmt.Sprintf("%s/%s", RuntimeObjectFinalizerName, managerPodName`

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -149,12 +163,28 @@ func (r *ReconcilePolicy) removeFinalizerAndReleaseMac(virtualMachine *kubevirt.
 	}
 
 	// remove our finalizer from the list and update it.
-	log.V(1).Info("removing finalizers",
-		"virtualMachineName", request.Name,
-		"virtualMachineNamespace", request.Namespace)
-	virtualMachine.ObjectMeta.Finalizers = helper.RemoveString(virtualMachine.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName)
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
ormergi, line 166: Please consider extract this function to a new one

RamLavi: please elaborate.

The reason we want to do Retry on conflict is that potentially, following vm delete, both kmp-mgr pods will try to remove their finalizers from the spec, which will cause conflict. This is ok, as long as they both succeed eventually.
Hopefully the DefaultRetry configuration will be sufficient , as it's supposed to be exactly the case it's supposed to handle.

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -389,42 +398,69 @@ func (p *PoolManager) AddMacToWaitingConfig(allocations map[string]string) error
...
 	if configMap.Data == nil {
 		log.Info("the configMap is empty")
-		return nil
+		return nil, false
 	}
 
+	isMacAllocated := false
 	for _, vmInterface := range vm.Spec.Template.Spec.Domain.Devices.Interfaces {
-		if vmInterface.MacAddress != "" {
+		if isInRange, _ := p.IsMacInRange(vmInterface.MacAddress); vmInterface.MacAddress != "" && isInRange {
```
alonSadan, line 429: sorry if I am to peaky, fill free to ignore.
isn't it worth to check if mac is empty, before we check if it is in range?
spare the function call...

pkg/controller/virtualmachine/virtualmachine_controller.go
RamLavi: I think pool_manager library is used only for using structs/const variables introduced their, while r.pool_manager is the actual instance of the NewPoolManager() that the reconciler holds.

I agree. Will move it to the pool_manager.

pkg/pool-manager/pod_pool.go
```diff
@@ -121,43 +120,42 @@ func (p *PoolManager) ReleasePodMac(podName string) error {
...
-		log.Info("released mac from pod", "mac", macAddr, "pod", podName)
+		if isInRange, _ := p.IsMacInRange(macAddr); isInRange {
+			delete(p.macPoolMap, macAddr)
+			log.Info("released mac from pod", "mac", macAddr, "pod", podName)
+		}
 	}
 
-	delete(p.podToMacPoolMap, podName)
-	log.V(1).Info("removed pod from podToMacPoolMap", "pod", podName)
+	if _, exist := p.podToMacPoolMap[podName]; exist {
```
alonSadan, line 129: why do we need the "if exist..."?
we check it here: 
https://github.com/k8snetworkplumbingwg/kubemacpool/pull/113/files#diff-f178743f06898ca1ae8f344f45065660R109

if we do need another check. why not use the `ok` var?
`if ok := p.podToMacPoolMap[podName]{ ...`

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -389,42 +398,69 @@ func (p *PoolManager) AddMacToWaitingConfig(allocations map[string]string) error
...
 			p.macPoolMap[vmInterface.MacAddress] = AllocationStatusAllocated
 			macAddress := strings.Replace(vmInterface.MacAddress, ":", "-", 5)
 			delete(configMap.Data, macAddress)
+			isMacAllocated = true
 		}
 	}
 
 	_, err = p.kubeClient.CoreV1().ConfigMaps(p.managerNamespace).Update(configMap)
+	if err != nil {
+		isMacAllocated = false
```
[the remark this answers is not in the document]
RamLavi: ack

pkg/pool-manager/pod_pool.go
RamLavi: Either the mac arrived from the pool, or set manually (and then it could be out of range and thus not inserted to podToMacPoolMap). Because of the latter - I check if it exists to avoid error..
correct me if I'm wrong, but since this map doesn't contain the mac, I can't ask if isMacInRange, so I have to check every time.

Regarding the syntax - I'm using the same syntax as [this](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/68e7721321f321911aa05148c68493dc44e6e555/pkg/pool-manager/pod_pool.go#L110). Did you mean change from "exist" to "ok"?

pkg/pool-manager/virtualmachine_pool.go
RamLavi: totally agree. I love picky reviews. please be as picky as you possible can :)

pkg/pool-manager/pod_pool.go
alonSadan: >Either the mac arrived from the pool, or set manually (and then it could be out of range and thus not inserted to podToMacPoolMap). Because of the latter - I check if it exists to avoid error..

sorry but I am not sure I understand.
don't we check the same condition?  `if some_bool := p.podToMacPoolMap[podName]`
if the pod was there the first time. what could remove it?

RamLavi: > if the pod was there the first time. what could remove it?

Please notice that we don't always insert the mac to macPoolMap & podToMacPoolMap in ([allocatePodRequestedMac()](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/f2e6bb8604dc6207ee538ca900d72505b9125f0a/pkg/pool-manager/pod_pool.go#L139))

since it is not inserted in case the mac was manually chosen & out of the pod's range - we cannot assume it's their now.

RamLavi: I see. since I'm already asking if in the start of the [ReleasePodMac()](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/f2e6bb8604dc6207ee538ca900d72505b9125f0a/pkg/pool-manager/pod_pool.go#L109) 
Your'e totally right. will fix. (if not what you meant then please elaborate)

alonSadan: exactly what I meant:)
Thank you.

pkg/pool-manager/pool_test.go
```diff
@@ -82,6 +82,28 @@ var _ = Describe("Pool", func() {
...
+			startMacAddrHW, err := net.ParseMAC(startMacAddr)
+			Expect(err).ToNot(HaveOccurred())
+			endMacAddrHW, err := net.ParseMAC(endMacAddr)
+			Expect(err).ToNot(HaveOccurred())
+			testMacAddrHW, err := net.ParseMAC(testMacAddr)
+			Expect(err).ToNot(HaveOccurred())
+			isInRange := inRange(startMacAddrHW, endMacAddrHW, testMacAddrHW)
+			Expect(isInRange).To(Equal(needToFail))
+		},
+			table.Entry("Start: 02:00:00:00:00:00  End: 03:00:00:00:00:00 Mac-Tested: 02:00:00:01:00:00", "02:00:00:00:00:00", "03:00:00:00:00:00", "02:00:00:01:00:00", true),
```
alonSadan, line 95: isn't the name "needToFail" , is a bit misleading here , since `Mac-Tested` is in range, it is actually should not fail?

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -389,42 +398,73 @@ func (p *PoolManager) AddMacToWaitingConfig(allocations map[string]string) error
...
+
+// This function is handled from the controller, and updates all successfully updated macs in the macPoolMap map
+func (p *PoolManager) UpdateMacsStatusInMap(virtualMachineName string) error {
+	p.poolMutex.Lock()
+	defer p.poolMutex.Unlock()
+	log.V(1).Info("UpdateMacsStatusInMap: pre update", "macmap", p.macPoolMap, "currentVmName", virtualMachineName)
+
+	for macAddress, macStatus := range p.macPoolMap {
+		switch macStatus {
+		case AllocationStatusWaitingForPod:
```
alonSadan, line 459: maybe add a log message stating if it was updated as allocated or  deleted?

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -92,52 +93,65 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 		return reconcile.Result{}, err
 	}
 
+	podFinalizerName := r.poolManager.GetManagerFinalizerName()
```
oshoval, line 96: there is still just one finalizer right ?
asking because when Sebastian explained You and me about it
it was raised that in case there are couple finalizers the code isnt robust and races can be occured
if i remember right

pkg/pool-manager/pool_test.go
```diff
@@ -82,6 +82,28 @@ var _ = Describe("Pool", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			endMacAddrHW, err := net.ParseMAC(endMacAddr)
+			Expect(err).ToNot(HaveOccurred())
+			testMacAddrHW, err := net.ParseMAC(testMacAddr)
+			Expect(err).ToNot(HaveOccurred())
+			isInRange := inRange(startMacAddrHW, endMacAddrHW, testMacAddrHW)
+			Expect(isInRange).To(Equal(shouldSucceed))
+		},
+			table.Entry("Start: 02:00:00:00:00:00  End: 03:00:00:00:00:00 Mac-Tested: 02:00:00:01:00:00", "02:00:00:00:00:00", "03:00:00:00:00:00", "02:00:00:01:00:00", true),
+			table.Entry("Start: 02:00:00:00:00:00  End: 03:00:00:00:00:00 Mac-Tested: 02:00:00:00:00:01", "02:00:00:00:00:00", "03:00:00:00:00:00", "02:00:00:00:00:01", true),
```
oshoval, line 96: how does the range is splitted ?
half half ?
what if the range has just one mac in case its legit ?
(or odd numbers? is it tested i guess)

tests/virtual_machines_test.go
```diff
@@ -258,7 +258,7 @@ var _ = Describe("Virtual Machines", func() {
 				Expect(newMac2.String()).To(Equal(mac2.String()))
 			})
 		})
-		//2179
+		//2179 TODO until we split the range to 2 separate non overlapping - then this test is expected to fail.
```
oshoval, line 261: you can add skip with a reason and then it will skip them if you want
see test/utils.go
bottom please for the predicates that you can use in order to skip

pkg/controller/virtualmachine/virtualmachine_controller.go
RamLavi: there's only one finalizer per pod. 
but the vm may have macs assigned from both ranges (due to multiple update requests) - so it will have 2 finalizers.
Do you think this could cause races? if so, please specify scenarios.

oshoval: i 99% remember that when Sebastian teached us this flow
we raised that in case of multi finalizers, since a reconcile might be undergoing, and the next one pending, it can cause a race that the update will be reverted or so, 
i would need here help from you and Sebastian, or whoever want

will send him link to this conversation

pkg/pool-manager/pool_test.go
RamLavi: good question, but please note that range splitting is not the agenda of this PR. A separate PR will be published soon about it (and once merged then this PR will be rebased on top it - because tests can't pass otherwise)

I will state, that corner cases such as
- Where the range only consists of 1 MACs.
- Where one of the pod’s ranges given runs out.

are intentionally ignored as they are considered not relevant to the user story. This is also described in the [design document](https://docs.google.com/document/d/1oBgw5x3U3LHV2laKdNw9hNLJdPSbK9wjDJwihE8NusU/edit?usp=sharing).

oshoval: thanks for the explanation
not sure what is 
Where one of the pod’s ranges given runs out.
just out of curiosity
thanks

RamLavi: when we assigned all the macs in one pod's pool.

oshoval: thx

tests/virtual_machines_test.go
RamLavi: I plan to remove these tests in a future dedicated task, where I will also add tests relevant to the sharding case

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -92,50 +92,58 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
...
 		"virtualMachineName", request.Name,
 		"virtualMachineNamespace", request.Namespace)
-	err = r.removeFinalizerAndReleaseMac(instance, &request)
+	err = r.removeFinalizerAndReleaseMac(instance, &request, podFinalizerName)
 	return reconcile.Result{}, err
 }
 
-func (r *ReconcilePolicy) addFinalizerAndUpdate(virtualMachine *kubevirt.VirtualMachine, request *reconcile.Request) error {
-	if helper.ContainsString(virtualMachine.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName) {
+func (r *ReconcilePolicy) addFinalizerAndUpdate(virtualMachine *kubevirt.VirtualMachine, request *reconcile.Request, podFinalizerName string) error {
```
phoracek, line 114: Why do you specify that the finalizer name is for "pod"? I think it is irrelevant in this context. This function just adds a finalizer to a VM. It does not matter whether the finalizer is "pod" or "foo".

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -92,50 +92,58 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
...
+		"virtualMachineNamespace", request.Namespace,
+		"podFinalizerName", podFinalizerName)
 
-	err := r.Update(context.Background(), virtualMachine)
-	if err != nil {
-		log.Error(err, "failed to update the VM with the new finalizer",
+	err, isMacAllocatedToRange := r.poolManager.MarkVMAsReady(virtualMachine)
+
+	//we want to add a finalizer only if this pod has allocated a mac to this vm
+	if isMacAllocatedToRange {
```
phoracek, line 131: This should be checked before we call function "addFinalizerAndUpdate". Please make sure to keep functions that do one thing and do it right. And of course, this one thing should be visible in the function's signature.

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -92,50 +92,58 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
...
-	log.V(1).Info("Finalizer was added to the VM",
-		"virtualMachineName", request.Name,
-		"virtualMachineNamespace", request.Namespace)
-
-	return r.poolManager.MarkVMAsReady(virtualMachine)
 }
 
-func (r *ReconcilePolicy) removeFinalizerAndReleaseMac(virtualMachine *kubevirt.VirtualMachine, request *reconcile.Request) error {
-	if !helper.ContainsString(virtualMachine.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName) {
+func (r *ReconcilePolicy) removeFinalizerAndReleaseMac(virtualMachine *kubevirt.VirtualMachine, request *reconcile.Request, podFinalizerName string) error {
```
phoracek, line 150: ditto

except here the logic of finalizer handling and poolManager is already mixed.

pkg/pool-manager/pool.go
```diff
@@ -147,6 +150,15 @@ func (p *PoolManager) InitMaps() error {
...
+	return inRange(p.rangeStart, p.rangeEnd, macAddressHW)
+}
+
+func (p *PoolManager) GetManagerFinalizerName() string {
+	return fmt.Sprintf("%s%s", RuntimeObjectFinalizerName, p.managerPodName)
```
phoracek, line 160: Using pod name as the finalizer name is dangerous. It basically forces us never to rename the manager pod, otherwise there would be forever trailing finalizer. Can we use the index of the range instead? Or maybe better (index-numOfReplicas), so we can calculate what part of range was allocated.

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -112,7 +112,11 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
```
phoracek, line 113: Again, the commit message suggests this should be in multiple commits.

tests/virtual_machines_test.go
```diff
@@ -427,8 +427,8 @@ var _ = Describe("Virtual Machines", func() {
 				_, err = net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
 				Expect(err).ToNot(HaveOccurred())
 
-				By("deleting leader manager")
-				DeleteLeaderManager()
+				By("deleting 1 kubemacpool manager pod")
```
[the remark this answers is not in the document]
RamLavi: I'm not sure what you mean.
In this test I want to check that there is no service downtime when only 1 pod is available. This should imitate a situation where 1 pod has crashed (doesn't really matter which of them)..

Could you kindly suggest how I could make the context more understandable than `When 1 pod mgr is down`?

tests/tests.go
```diff
@@ -259,8 +243,8 @@ func changeManagerReplicas(numOfReplica int32) error {
 		if managerStatefulset.Status.Replicas != numOfReplica {
 			return false
 		}
-		//due to readiness probe only 1 (the leader) pod will be ready (if any)
-		if float64(managerStatefulset.Status.ReadyReplicas) != math.Min(float64(1), float64(numOfReplica)) {
+
+		if managerStatefulset.Status.ReadyReplicas != numOfReplica {
```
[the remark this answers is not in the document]
RamLavi: the kmp is set to 2 replicas. but some of the pods **tests** use changeManagerReplicas() in order to temporarily "kill" kmp by setting the replicas to 0.

pkg/manager/manager.go
```diff
@@ -85,101 +81,81 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
...
 	}
 
-	err = k.newLeaderElection(k.config, leaderElectionManager.GetScheme())
+	err = kubevirt_api.AddToScheme(mgr.GetScheme())
 	if err != nil {
-		return fmt.Errorf("unable to create a leader election resource lock error %v", err)
+		return fmt.Errorf("unable to register kubevirt scheme error %v", err)
 	}
 
-	for k.continueToRunManager {
```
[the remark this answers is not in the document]
RamLavi: my bad. will return it.

config/default/manager/manager.yaml
```diff
@@ -130,7 +130,7 @@ spec:
     - port: 443
       targetPort: 8000
   selector:
-    kubemacpool-leader: "true"
+    app: kubemacpool
```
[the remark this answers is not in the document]
RamLavi: found it ! :)

pkg/controller/virtualmachine/virtualmachine_controller.go
RamLavi: I don't recall this conversation, but it raises a good scenario: In the current design, there could be a situation where both controllers will try to update at a pseudo same time. I dealt with this issue by using retryOnConflict method.
I think it shouldn't be problematic. @SchSeba - do you think otherwise?

oshoval: the 2nd reconcile wont restore the first finalizer  ?

RamLavi: it shouldn't. please consider the example I put here while going over [removeFinalizerAndReleaseMac](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/113/commits/37f48a8d8056463642765876e4af74466351fdc3#diff-cfe89be020e4377cb45005b5d0ce7227L149)  :

let's say 2 of the controllers - cnt1 & cnt2 perform a get and receive the vm object with 2 finalizers (finalizer1 & finalizer2). they both try to remove it's own finalizer.
now let's assume they both delete and try to delete it's own finalizer at the same time. this would cause a conflict for 1 of them and the second should succeed. 
(let's assume cnt1 succeded)
But since we are using retry on conflict on cnt-2, then we will get() the vm object **again** (but now it will include the change done by the cnt1). in other words - the object will include only the fin2. from there cnt should remove fin2 with no other conflicts.
retry on conflict also has built in backoff random times so that there shouldn't be reoccurring conflicts.

pkg/controller/virtualmachine/virtualmachine_controller.go
RamLavi: ack

pkg/pool-manager/pool.go
RamLavi: afaik, the index is best attained inside a pod - from it's name. Good catch, thanks!

Issue #101, opened 2020-03-04
virtual machine e2e test #2633 sometimes fail


RamLavi:
Following the exclusion of virtual machine e2e test [#2633](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/b69c56e0ef0d080b5391aef151273e2746b0bb25/tests/virtual_machines_test.go#L373) in the following [PR](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/100/files), this issue has been open to resolve why the test sometimes fail.

test error logs summary are: (full logs [attached](https://github.com/k8snetworkplumbingwg/kubemacpool/files/4287389/debugging_KMP_TEST_2633.2.txt) )

```
2020-03-04T06:15:56.949Z	ERROR	PoolManager	mac address already allocated	{"error": "failed to allocate requested mac address"}
github.com/k8snetworkplumbingwg/kubemacpool/vendor/github.com/go-logr/zapr.(*zapLogger).Error
	/go/src/github.com/k8snetworkplumbingwg/kubemacpool/vendor/github.com/go-logr/zapr/zapr.go:128
github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager.(*PoolManager).allocateRequestedVirtualMachineInterfaceMac
	/go/src/github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager/virtualmachine_pool.go:226
github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager.(*PoolManager).AllocateVirtualMachineMac
	/go/src/github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager/virtualmachine_pool.go:71
github.com/k8snetworkplumbingwg/kubemacpool/pkg/webhook/virtualmachine.(*virtualMachineAnnotator).mutateCreateVirtualMachinesFn
	/go/src/github.com/k8snetworkplumbingwg/kubemacpool/pkg/webhook/virtualmachine/virtualmachine.go:104
github.com/k8snetworkplumbingwg/kubemacpool/pkg/webhook/virtualmachine.(*virtualMachineAnnotator).Handle
	/go/src/github.com/k8snetworkplumbingwg/kubemacpool/pkg/webhook/virtualmachine/virtualmachine.go:71
github.com/k8snetworkplumbingwg/kubemacpool/vendor/sigs.k8s.io/controller-runtime/pkg/webhook/admission.(*Webhook).Handle
[...]

PR #114, opened 2020-04-12
Ensure there are no VM's before test suite and each test

ormergi:
This PR fixes issue #101 by ensuring that `kubemacpool-vm-configmap` map is empty before test suite and each test,
it also prevent getting error 500 with the reason "range is full"  from `mutatevirtualmachine.kubemacpool.io` webhook.

### DIFF
--- a/pkg/pool-manager/pool.go
+++ b/pkg/pool-manager/pool.go
@@ -24,7 +24,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 
-	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
+    "github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 )
 
 const (
--- a/pkg/pool-manager/pool.go
+++ b/pkg/pool-manager/pool.go
@@ -23,8 +23,6 @@ import (
 
 	"k8s.io/client-go/kubernetes"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
-
-    "github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 )
 
 const (
@@ -33,7 +31,6 @@ const (
 	RuntimeObjectFinalizerName = "k8s.v1.cni.cncf.io/kubeMacPool"
 	networksAnnotation         = "k8s.v1.cni.cncf.io/networks"
 	networksStatusAnnotation   = "k8s.v1.cni.cncf.io/networks-status"
-	vmWaitConfigMapName        = names.WAITING_VMS_CONFIGMAP
 )
 
 var log = logf.Log.WithName("PoolManager")
--- a/pkg/pool-manager/pool.go
+++ b/pkg/pool-manager/pool.go
@@ -23,6 +23,8 @@ import (
 
 	"k8s.io/client-go/kubernetes"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 )
 
 const (
@@ -31,7 +33,7 @@ const (
 	RuntimeObjectFinalizerName = "k8s.v1.cni.cncf.io/kubeMacPool"
 	networksAnnotation         = "k8s.v1.cni.cncf.io/networks"
 	networksStatusAnnotation   = "k8s.v1.cni.cncf.io/networks-status"
-	vmWaitConfigMapName        = "kubemacpool-vm-configmap"
+	vmWaitConfigMapName        = names.WAITING_VMS_CONFIGMAP
 )
 
 var log = logf.Log.WithName("PoolManager")
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -34,3 +34,11 @@ func RemoveString(slice []string, s string) (result []string) {
 	}
 	return
 }
+
+func ClearMap(inputMap map[string]string){
+	if inputMap != nil {
+		for key := range inputMap {
+			delete(inputMap, key)
+		}
+	}
+}
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -34,11 +34,3 @@ func RemoveString(slice []string, s string) (result []string) {
 	}
 	return
 }
-
-func ClearMap(inputMap map[string]string) {
-	if inputMap != nil {
-		for key := range inputMap {
-			delete(inputMap, key)
-		}
-	}
-}
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -35,7 +35,7 @@ func RemoveString(slice []string, s string) (result []string) {
 	return
 }
 
-func ClearMap(inputMap map[string]string){
+func ClearMap(inputMap map[string]string) {
 	if inputMap != nil {
 		for key := range inputMap {
 			delete(inputMap, key)
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -9,10 +9,10 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/rand"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -9,10 +9,10 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/rand"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 	pool_manager "github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager"
+	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/utils"
 	kubevirtv1 "kubevirt.io/client-go/api/v1"
 )
 
@@ -46,9 +47,7 @@ var _ = Describe("Virtual Machines", func() {
 		Expect(err).ToNot(HaveOccurred())
 
 		// Clear the map in-place instead of waiting to garbage-collector
-		for entry := range vmWaitConfigMap.Data {
-			delete(vmWaitConfigMap.Data, entry)
-		}
+		utils.ClearMap(vmWaitConfigMap.Data)
 		vmWaitConfigMap, err = testClient.KubeClient.CoreV1().ConfigMaps(names.MANAGER_NAMESPACE).Update(vmWaitConfigMap)
 		Expect(err).ToNot(HaveOccurred())
 		Expect(vmWaitConfigMap.Data).To(BeEmpty())
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -35,25 +35,25 @@ var _ = Describe("Virtual Machines", func() {
 		Expect(result.Error()).NotTo(HaveOccurred())
 	})
 
-	BeforeEach(func() {
-		// Verify that there are no VMs in the cluster
-		currentVMList := &kubevirtv1.VirtualMachineList{}
-		err := testClient.VirtClient.List(context.TODO(), currentVMList, &client.ListOptions{})
-		Expect(err).ToNot(HaveOccurred())
-		Expect(len(currentVMList.Items)).To(BeZero())
-
-		// Clear vmWaitConfigMap configMap data before each test
-		vmWaitConfigMap, err := testClient.KubeClient.CoreV1().ConfigMaps(names.MANAGER_NAMESPACE).Get(names.WAITING_VMS_CONFIGMAP, meta_v1.GetOptions{})
-		Expect(err).ToNot(HaveOccurred())
-
-		// Clear the map in-place instead of waiting to garbage-collector
-		utils.ClearMap(vmWaitConfigMap.Data)
-		vmWaitConfigMap, err = testClient.KubeClient.CoreV1().ConfigMaps(names.MANAGER_NAMESPACE).Update(vmWaitConfigMap)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(vmWaitConfigMap.Data).To(BeEmpty())
-	})
-
 	Context("Check the client", func() {
+		BeforeEach(func() {
+			// Verify that there are no VMs in the cluster
+			currentVMList := &kubevirtv1.VirtualMachineList{}
+			err := testClient.VirtClient.List(context.TODO(), currentVMList, &client.ListOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(currentVMList.Items)).To(BeZero())
+
+			// Clear vmWaitConfigMap configMap data before each test
+			vmWaitConfigMap, err := testClient.KubeClient.CoreV1().ConfigMaps(names.MANAGER_NAMESPACE).Get(names.WAITING_VMS_CONFIGMAP, meta_v1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+			// Clear the map in-place instead of waiting to garbage-collector
+			utils.ClearMap(vmWaitConfigMap.Data)
+			vmWaitConfigMap, err = testClient.KubeClient.CoreV1().ConfigMaps(names.MANAGER_NAMESPACE).Update(vmWaitConfigMap)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(vmWaitConfigMap.Data).To(BeEmpty())
+		})
+
 		AfterEach(func() {
 			vmList := &kubevirtv1.VirtualMachineList{}
 			err := testClient.VirtClient.List(context.TODO(), vmList, &client.ListOptions{})
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 	pool_manager "github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager"
-	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/utils"
 	kubevirtv1 "kubevirt.io/client-go/api/v1"
 )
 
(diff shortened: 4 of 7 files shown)


tests/virtual_machines_test.go
```diff
@@ -195,7 +195,7 @@ var _ = Describe("Virtual Machines", func() {
 			})
 		})
 		//2162 test postponed due to issue: https://github.com/k8snetworkplumbingwg/kubemacpool/issues/104
-		PContext("When trying to create a VM after all MAC addresses in range have been occupied", func() {
+		Context("When trying to create a VM after all MAC addresses in range have been occupied", func() {
```
RamLavi, line 198: please keep  the P on these tests as it is currently examined by @alonSadan in another [task](https://issues.redhat.com/browse/CNV-4404)

tests/virtual_machines_test.go
```diff
@@ -218,7 +218,7 @@ var _ = Describe("Virtual Machines", func() {
 			})
 		})
 		//2165 test postponed due to issue: https://github.com/k8snetworkplumbingwg/kubemacpool/issues/105
-		PContext("when trying to create a VM after a MAC address has just been released duo to a VM deletion", func() {
+		Context("when trying to create a VM after a MAC address has just been released duo to a VM deletion", func() {
```
RamLavi, line 221: please keep the P on these tests as it is currently examined by @RamLavi  in another [task](https://issues.redhat.com/browse/CNV-4405)

tests/virtual_machines_test.go
```diff
@@ -32,6 +34,26 @@ var _ = Describe("Virtual Machines", func() {
 		Expect(result.Error()).NotTo(HaveOccurred())
 	})
 
+	BeforeEach(func() {
+		// Verify that there are no VMs in the cluster
+		currentVMList := &kubevirtv1.VirtualMachineList{}
```
alonSadan, line 39: Not sure here, but maybe this verification should be warped in`Eventually`, because we need to wait for the controller to remove the VMs.

tests/virtual_machines_test.go
```diff
@@ -50,6 +72,7 @@ var _ = Describe("Virtual Machines", func() {
 				return len(vmList.Items)
 
 			}, timeout, pollingInterval).Should(Equal(0), "failed to remove all vm objects")
+
```
alonSadan, line 75: maybe remove newline here?

tests/virtual_machines_test.go
```diff
@@ -32,6 +34,26 @@ var _ = Describe("Virtual Machines", func() {
...
+		vmWaitConfigMap, err := testClient.KubeClient.CoreV1().ConfigMaps(names.MANAGER_NAMESPACE).Get(names.WAITING_VMS_CONFIGMAP, meta_v1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred())
+
+		// Clear the map in-place instead of waiting to garbage-collector
+		for entry := range vmWaitConfigMap.Data {
+			delete(vmWaitConfigMap.Data, entry)
+		}
+		vmWaitConfigMap, err = testClient.KubeClient.CoreV1().ConfigMaps(names.MANAGER_NAMESPACE).Update(vmWaitConfigMap)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(vmWaitConfigMap.Data).To(BeEmpty())
```
alonSadan, line 54: same here. maybe warp in eventually because it could  take time. not sure.

tests/virtual_machines_test.go
```diff
@@ -32,6 +34,26 @@ var _ = Describe("Virtual Machines", func() {
 		Expect(result.Error()).NotTo(HaveOccurred())
 	})
 
+	BeforeEach(func() {
```
RamLavi, line 37: why is BeforeEach not in the context of "Check the client" like AfterEach?

tests/virtual_machines_test.go
RamLavi, line 39: This is redundant. we are already checking the vm's were deleted in AfterEach.
Please consider moving all to BeforeEach or the other way around..

tests/virtual_machines_test.go
ormergi: This is a double check that there is no VM in the cluster, the deletion is done in AfterEach block and it is done with eventually.

tests/virtual_machines_test.go
ormergi: This is a double check that there is no VM in the cluster, the deletion is done in AfterEach block and it is done with eventually.

tests/virtual_machines_test.go
ormergi: It should?

tests/virtual_machines_test.go
```diff
@@ -12,9 +12,11 @@ import (
 	. "github.com/onsi/gomega"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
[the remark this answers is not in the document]
ormergi: done

tests/virtual_machines_test.go
ormergi: Its make sense to me that AfterEach should clean after the test, in this case delete the VMs that were created during the test, 
and BeforeEach checks that "we start fresh" and the vmWaitConfigmap is empty.
Maybe we can move the map clearing to AfterEach and leave all the checks in BeforaEach.

I understand it is redundant, but think about it if it was there earlier we were not in this situation ;)

tests/virtual_machines_test.go
RamLavi: I think this would mean BeforeEach will work only once - before "Check the client"

and even if not - BeforeEach & AfterEach should be in the same context, it's confusing otherwise, since we want them to all work before and after each test.

tests/virtual_machines_test.go
RamLavi: I don't think there is a need for a double check. If I understand correctly, this expect could never be true, and therefor redundant. 

The reason that it was missed before was the fact that we had remnant in the configmap, not the vms.

tests/virtual_machines_test.go
ormergi: Take a look at this [1] and [2] 
Update invokes Do which get request as a receiver,
and according to the comment in the code we will get the response synchronously, 
so Eventualy seems redundant here, please correct me if I am wrong :)

[1] https://github.com/kubernetes/client-go/blob/v11.0.0/rest/request.go#L674 
[2] https://github.com/kubernetes/client-go/blob/v11.0.0/rest/request.go#L800

tests/virtual_machines_test.go
ormergi: In case AfterEach delete VM part which inside an Eventually times out, a VM might be left in the cluster, then this double check will raise an error about it.

I meant, if there was a check or cleaning remnants in the vmWait configmap we were not in this state.

tests/virtual_machines_test.go
ormergi: done

tests/virtual_machines_test.go
```diff
@@ -195,7 +219,7 @@ var _ = Describe("Virtual Machines", func() {
 			})
 		})
 		//2162 test postponed due to issue: https://github.com/k8snetworkplumbingwg/kubemacpool/issues/104
```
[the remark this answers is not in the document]
ormergi: done

tests/virtual_machines_test.go
```diff
@@ -218,7 +242,7 @@ var _ = Describe("Virtual Machines", func() {
 			})
 		})
 		//2165 test postponed due to issue: https://github.com/k8snetworkplumbingwg/kubemacpool/issues/105
```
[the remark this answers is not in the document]
ormergi: done

tests/virtual_machines_test.go
```diff
@@ -330,7 +354,7 @@ var _ = Describe("Virtual Machines", func() {
 			})
 		})
 		//2633 test postponed due to issue: https://github.com/k8snetworkplumbingwg/kubemacpool/issues/101
```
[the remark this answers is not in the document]
ormergi: done

tests/virtual_machines_test.go
```diff
@@ -194,8 +216,7 @@ var _ = Describe("Virtual Machines", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 		})
-		//2162 test postponed due to issue: https://github.com/k8snetworkplumbingwg/kubemacpool/issues/104
-		PContext("When trying to create a VM after all MAC addresses in range have been occupied", func() {
+		Context("When trying to create a VM after all MAC addresses in range have been occupied", func() {
```
alonSadan, line 219: we need to be consistent on putting newline between Context or not. I think we should.

tests/virtual_machines_test.go
```diff
@@ -217,8 +238,7 @@ var _ = Describe("Virtual Machines", func() {
 				Expect(err).To(HaveOccurred())
 			})
 		})
-		//2165 test postponed due to issue: https://github.com/k8snetworkplumbingwg/kubemacpool/issues/105
-		PContext("when trying to create a VM after a MAC address has just been released duo to a VM deletion", func() {
+		Context("when trying to create a VM after a MAC address has just been released duo to a VM deletion", func() {
```
alonSadan, line 241: ditto

tests/virtual_machines_test.go
```diff
@@ -329,8 +349,7 @@ var _ = Describe("Virtual Machines", func() {
 				}
 			})
 		})
-		//2633 test postponed due to issue: https://github.com/k8snetworkplumbingwg/kubemacpool/issues/101
-		PContext("When we re-apply a failed VM yaml", func() {
+		Context("When we re-apply a failed VM yaml", func() {
```
alonSadan, line 352: ditto

tests/virtual_machines_test.go
ormergi: Most of the Context block has no newline between, I think we should follow.

tests/virtual_machines_test.go
```diff
@@ -33,6 +36,24 @@ var _ = Describe("Virtual Machines", func() {
 	})
 
 	Context("Check the client", func() {
+		BeforeEach(func() {
+			// Verify that there are no VMs in the cluster
```
RamLavi, line 40: is there a scenario where this check will find a live vm in the cluster? if the answer is no then I think it should be dropped. 
I think the [check](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/ee7e920feb78b555e329d8ac47644bc8b0871e20/tests/virtual_machines_test.go#L73) in the AfterEach is already checking this.

RamLavi:
this PR will fix [issue #101 ](https://github.com/k8snetworkplumbingwg/kubemacpool/issues/101)

tests/virtual_machines_test.go
ormergi: In case AfterEach delete VM part which inside an Eventually times out, a VM might be left in the cluster, then this double check will raise an error about it.

phoracek:
@ormergi would you please update the title and description of the PR? It seems that you've actually made some fixes, but the PR seems to be just a verification one.

pkg/pool-manager/pool.go
```diff
@@ -31,7 +33,7 @@ const (
 	RuntimeObjectFinalizerName = "k8s.v1.cni.cncf.io/kubeMacPool"
 	networksAnnotation         = "k8s.v1.cni.cncf.io/networks"
 	networksStatusAnnotation   = "k8s.v1.cni.cncf.io/networks-status"
-	vmWaitConfigMapName        = "kubemacpool-vm-configmap"
+	vmWaitConfigMapName        = names.WAITING_VMS_CONFIGMAP
```
phoracek, line 36: Thy do we need this indirection? Can't you just use it from `names`?

pkg/utils/utils.go
```diff
@@ -34,3 +34,11 @@ func RemoveString(slice []string, s string) (result []string) {
 	}
 	return
 }
+
+func ClearMap(inputMap map[string]string) {
```
phoracek, line 38: This is used only in tests, please move it to tests module or even to the test using it directly as a private function until we have multiple users.

tests/tests.go
```diff
@@ -9,10 +9,10 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/rand"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
```
phoracek, line 14: Please keep this outside of this PR. It is good to fix modules to satisfy linting, but touching modules you don't edit may cause rebase conflicts in other PRs.

tests/virtual_machines_test.go
phoracek, line 241: Please keep `//2165` so we have a reference what to replace later with proper polarion ids.

tests/virtual_machines_test.go
```diff
@@ -33,6 +36,24 @@ var _ = Describe("Virtual Machines", func() {
 	})
 
 	Context("Check the client", func() {
+		BeforeEach(func() {
```
phoracek, line 39: Would it make sense to have this check in the main BeforeAll?

tests/virtual_machines_test.go
phoracek, line 40: Please add these comments to `By("Verifying that there are no VMs in the cluster")`. Then it is easier to tell what failed in the log.

tests/virtual_machines_test.go
```diff
@@ -33,6 +36,24 @@ var _ = Describe("Virtual Machines", func() {
...
+			// Verify that there are no VMs in the cluster
+			currentVMList := &kubevirtv1.VirtualMachineList{}
+			err := testClient.VirtClient.List(context.TODO(), currentVMList, &client.ListOptions{})
+			Expect(err).ToNot(HaveOccurred())
```
phoracek, line 43: Reading errors from blank expects is hard, please add expectations string:

`Expect(err).ToNot(HaveOccurred(), "Should successfully list add VMs")`

The same for other new Expects.

pkg/pool-manager/pool.go
ormergi: I will fix it

ormergi: We can use it directly if  I add `.` before the package path, but I think it easier to understand from where constants strings come from if we leave the object that hold them (`names` in this case)

tests/virtual_machines_test.go
ormergi: done

pkg/utils/utils.go
ormergi: done

tests/tests.go
ormergi: done

tests/virtual_machines_test.go
ormergi: done

tests/virtual_machines_test.go
ormergi: done

tests/tests.go
```diff
@@ -9,10 +9,9 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/rand"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
```
phoracek, line 14: My comment is still pending.

pkg/pool-manager/pool.go
phoracek: You can also just use names.WAITING_VMS_CONFIGMAP in this module to keep it explicit. But if you believe this is more readable, fair enough

tests/virtual_machines_test.go
```diff
@@ -30,9 +32,34 @@ var _ = Describe("Virtual Machines", func() {
...
+		By("Verify that there are no VMs in the cluster before starting test suite")
+		beforeAllVMList := &kubevirtv1.VirtualMachineList{}
+		err := testClient.VirtClient.List(context.TODO(), beforeAllVMList, &client.ListOptions{})
+		Expect(err).ToNot(HaveOccurred(), "Should successfully list add VMs")
+		Expect(len(beforeAllVMList.Items)).To(BeZero(), "There should be no VM's in the cluster before running test suite")
 	})
 
 	Context("Check the client", func() {
+		BeforeEach(func() {
+			By("Verify that there are no VMs left from previous tests")
```
phoracek, line 45: All this clearing can be done in BeforeEach of the whole Describe since all the tests here are dealing with VMs. And if you keep this first check in `Describe` BeforeEach, you can drop the one from BeforeAll

tests/virtual_machines_test.go
```diff
@@ -30,9 +32,34 @@ var _ = Describe("Virtual Machines", func() {
...
+			err := testClient.VirtClient.List(context.TODO(), currentVMList, &client.ListOptions{})
+			Expect(err).ToNot(HaveOccurred(), "Should successfully list add VMs")
+			Expect(len(currentVMList.Items)).To(BeZero(), "There should be no VM's in the cluster before a test")
+
+			By("Clear vmWaitConfigMap configMap data before each test")
+			vmWaitConfigMap, err := testClient.KubeClient.CoreV1().ConfigMaps(names.MANAGER_NAMESPACE).Get(names.WAITING_VMS_CONFIGMAP, meta_v1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Should successfully get %s ConfigMap", names.WAITING_VMS_CONFIGMAP))
+
+			By("Clearing vmWaitConfigMap configMap data")
+			//  Clearing the map in-place instead of waiting to garbage-collector.
```
phoracek, line 56: Redundant wuth `By`. If you want to keep the detail about garbage collector, you can add it to By

tests/virtual_machines_test.go
```diff
@@ -30,9 +32,34 @@ var _ = Describe("Virtual Machines", func() {
...
 
 	Context("Check the client", func() {
+		BeforeEach(func() {
+			By("Verify that there are no VMs left from previous tests")
+			currentVMList := &kubevirtv1.VirtualMachineList{}
+			err := testClient.VirtClient.List(context.TODO(), currentVMList, &client.ListOptions{})
+			Expect(err).ToNot(HaveOccurred(), "Should successfully list add VMs")
+			Expect(len(currentVMList.Items)).To(BeZero(), "There should be no VM's in the cluster before a test")
+
+			By("Clear vmWaitConfigMap configMap data before each test")
```
phoracek, line 51: You are not clearing then now, just getting them to clear later. Let's keep the `By` telling truth for easier debugging

tests/virtual_machines_test.go
```diff
@@ -30,9 +32,34 @@ var _ = Describe("Virtual Machines", func() {
...
+			vmWaitConfigMap, err := testClient.KubeClient.CoreV1().ConfigMaps(names.MANAGER_NAMESPACE).Get(names.WAITING_VMS_CONFIGMAP, meta_v1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Should successfully get %s ConfigMap", names.WAITING_VMS_CONFIGMAP))
+
+			By("Clearing vmWaitConfigMap configMap data")
+			//  Clearing the map in-place instead of waiting to garbage-collector.
+			clearMap(vmWaitConfigMap.Data)
```
phoracek, line 57: Can't we just delete it instead?

tests/virtual_machines_test.go
ormergi: Like in this [1] commit?

[1] https://github.com/k8snetworkplumbingwg/kubemacpool/blob/945b21482f692e77f12ffa4fa9d5dd9ad46b3d1b/tests/virtual_machines_test.go

phoracek: Yeah

tests/virtual_machines_test.go
ormergi: Its more faster to just clear the map that inside the config map instead of create and delete the ConfigMap object every test, less overhead.

phoracek: Ack

tests/tests.go
ormergi: `Make fmt` keep adding them :/

RamLavi:
@ormergi - if the PR fixes issue #101 , then please remove the P and comment from it.

ormergi:
You mentioned this PR fixes #101, lets test it

PR #123, opened 2020-04-28
Bump multus-cni version

ormergi:
This PR fixes [1] by bumping multus-cni dependency to latest version [v3.4.1](https://github.com/intel/multus-cni/releases/tag/v3.4.1), updating nesessery changes and addes unit tests to prevent bugs related to [1].

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1816971


phoracek:
To make sure this indeed fixes the issue we had with multiple IPs, could you please add a test requesting array of ips: https://bugzilla.redhat.com/show_bug.cgi?id=1816971

Since this is for Pods, unit test would be sufficient.

ormergi:
/hold
should be merged after #126

pkg/pool-manager/pool_test.go
```diff
@@ -389,7 +389,7 @@ var _ = Describe("Pool", func() {
 			_, exist = poolManager.macPoolMap["02:00:00:00:00:01"]
 			Expect(exist).To(BeTrue())
 
-			Expect(newPod.Annotations[networksAnnotation]).To(Equal(`[{"name":"ovs-conf","namespace":"default","mac":"02:00:00:00:00:01"}]`))
+			Expect(newPod.Annotations[networksAnnotation]).To(Equal(`[{"name":"ovs-conf","namespace":"default","mac":"02:00:00:00:00:01","cni-args":null}]`))
```
RamLavi, line 393: small typo in commit message: `on of them is CNIArgs [2]` on --> one

ormergi: fixed

pkg/controller/pod/pod_controller.go
```diff
@@ -82,7 +82,7 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 		if errors.IsNotFound(err) {
 			err := r.poolManager.ReleasePodMac(fmt.Sprintf("%s/%s", request.Namespace, request.Name))
 			if err != nil {
-				logging.Printf(logging.ErrorLevel, "failed to release mac for pod %s: %v", request.NamespacedName, err)
+				logging.Errorf("%v, failed to release mac for pod %s: %v",logging.ErrorLevel, request.NamespacedName, err)
```
phoracek, line 85: I don't get how is this related to bump of Multus. If it is really needed, please explain it in the commit message.

pkg/pool-manager/pool_test.go
```diff
@@ -413,5 +414,53 @@ var _ = Describe("Pool", func() {
 			Expect(newPod.Annotations[networksAnnotation]).To(Equal(afterAllocationAnnotation[networksAnnotation]))
 		})
 	})
+	Describe("Multus Network Annotations API Tests", func() {
+		It("should allocate requested mac-address with single ip-address request in network annotation", func() {
+			poolManager := createPoolManager("02:00:00:00:00:00", "02:00:00:00:00:01", &vmConfigMap)
```
phoracek, line 419: I know this follows the existing tests, but let's try to make our tests better today. Setup of the environment should be done in Context by BeforeEach. So in this case:

```go
Describe("... API Tests",... {

    Context("when pool is configured with two available addresses", ... {
        BeforeEach({
            // the pool setup
        })

        Context("and a Pod is created with single IP requested",...
            BeforeEach({
                // the VM setup
            })

            It("should successfullly ....", {
                // actual test
            })
    })
})

File: pkg/pool-manager/pool_test.go [hunk 4]
```diff
@@ -413,5 +414,53 @@ var _ = Describe("Pool", func() {
...
+			Expect(poolManager).ToNot(Equal(nil), "failed to create pool-manager")
```
phoracek, line 420: The expectation string should be what **should** happen. So in this case "Should create pool-manager". If we align to this rule everywhere in the code, then we can just read the "strings" in the tests to understand what is going on.

pkg/pool-manager/pool_test.go
```diff
@@ -413,5 +414,53 @@ var _ = Describe("Pool", func() {
...
 
+			expectedMacRequestWithIpAnnotation := `[{"name":"ovs-conf","namespace":"default","ips":["10.10.0.1"],"mac":"02:00:00:00:00:00","cni-args":null}]`
+			newPod := samplePod
+			newPod.Name = "testPod"
+			newPod.Annotations = map[string]string{networksAnnotation: `[{"name":"ovs-conf","namespace":"default","ips":["10.10.0.1"]}]`}
+
+			err := poolManager.AllocatePodMac(&newPod)
+			Expect(err).ToNot(HaveOccurred(),"failed to handle allocate-pod-mac request with ip-address")
+			Expect(newPod.Annotations[networksAnnotation]).To(Equal(expectedMacRequestWithIpAnnotation),
+				fmt.Sprintf("Pod networtk annotation is not as expected,\n"+
```
qinqon, line 430: You have to use Gomega JSON matcher MatchJSON so this check is not affected by json attribute re-order.

The "obtained" and "expected" typical test log has to be printed out by Gomega not by us, the string should be something like "Should annotate pod with IP and MAC"

pkg/pool-manager/pool_test.go
```diff
@@ -413,5 +414,53 @@ var _ = Describe("Pool", func() {
...
+
+			expectedMacRequestWithIpsAnnotation := `[{"name":"ovs-conf","namespace":"default","ips":["10.10.0.1","10.10.0.2","10.0.0.3"],"mac":"02:00:00:00:00:00","cni-args":null}]`
+			newPod := samplePod
+			newPod.Name = "testPod"
+			newPod.Annotations = map[string]string{networksAnnotation: `[{"name":"ovs-conf","namespace":"default","ips":["10.10.0.1","10.10.0.2","10.0.0.3"]}]`}
+
+			err := poolManager.AllocatePodMac(&newPod)
+			Expect(err).ToNot(HaveOccurred(), "failed to handle allocate-pod-mac request with multiple ip-addresses")
+			Expect(newPod.Annotations[networksAnnotation]).To(Equal(expectedMacRequestWithIpsAnnotation),
+				fmt.Sprintf("Pod networtk annotation is not as expected,\n"+
```
qinqon, line 448: ditto

pkg/controller/pod/pod_controller.go
ormergi: This function so no longer public since this PR https://github.com/intel/multus-cni/pull/339
I updated the commit message.

phoracek: Ack

pkg/pool-manager/pool_test.go
ormergi: For me it is more clear to see these kind of error messages in after test suite logs,
specifying what failed and why instead of what should have happen.
I will align with this kind of error messages :).

phoracek: Hopefully, you will read the code more often than seeing it fail. The positive string helps you read the code alone.

Here is an example :

```go
			It("should return a Result with RequeueAfter set (trigger re-reconciliation)", func() {
				result, err := reconciler.Reconcile(request)
				Expect(err).ToNot(HaveOccurred(), "Should succcessfully reconcile the request")
				Expect(result).To(Equal(reconcile.Result{RequeueAfter: 666 * time.Second}), "Result should have expected requeue value")
			})
```

The result (written in red is):

```
  Result should have expected requeue value
  Expected
      <reconcile.Result>: {
          Requeue: false,
          RequeueAfter: 5000000000,
      }
  to equal
      <reconcile.Result>: {
          Requeue: false,
          RequeueAfter: 666000000000,
      }
```

So you see the expectation and the actual result.

The biggest benefit of this is, that the code is way more readable when you just follow strings and don't need to revert their meaning to see what **should** happen.

pkg/pool-manager/pool_test.go
ormergi: ack

pkg/pool-manager/pool_test.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package pool_manager
 
 import (
+	multus "github.com/intel/multus-cni/types"
```
phoracek, line 20: We usually split imports to 3 groups (sometimes 4 on test modules). The order is:

1. Built-in libraries ("net")
2. Third-party imports ("k8s.io/api/core/v1")
3. Internal imports ("github.com/k8snetworkplumbingwg/kubemacpool/pkg/names")

And then we sometimes plug a separate group with the test imports ("github.com/onsi/gomega") somewhere, but I can't really say we have a specific place for it.

By keeping this order, it is way easier for a reader of the code to quickly figure out what are the dependencies.

pkg/pool-manager/pool_test.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
...
+	"github.com/k8snetworkplumbingwg/kubemacpool/build/_output/bin/go/src/encoding/json"
```
phoracek, line 21: We should not import generated code from "_output". That directory is not a part of the distribution. Your IDE probably automatically loaded a wrong lib.

pkg/pool-manager/pool_test.go
```diff
@@ -414,4 +416,75 @@ var _ = Describe("Pool", func() {
 		})
 	})
 
-})
+	Describe("Multus Network Annotations API Tests", func() {
+		BeforeEach(func() {
```
phoracek, line 420: As noted offline, the before each should keep only the two following lines. poolManager needs to be initialized in the root of Describe and then populated in BeforeEach (we want to reset the manager per each test).

pkg/pool-manager/pool_test.go
```diff
@@ -414,4 +416,75 @@ var _ = Describe("Pool", func() {
...
+			poolManager := createPoolManager("02:00:00:00:00:00", "02:00:00:00:00:01", &vmConfigMap)
+			Expect(poolManager).ToNot(Equal(nil), "should create pool-manager")
+
+			Context("When pool-manager is configured with available addresses", func(){
```
phoracek, line 424: Context without any content (except other contexts) is suspicions. If you are not setting anything, you are probably keeping the context implicit and defined in internal code. That is risky since it may get outdated quickly, it does not keep the context info visible and does not force us to actually follow the context.

I think that in this case, this Context is redundant since you describe the addresses parameter in the following Context. Please drop it.

phoracek, line 424: For the `"When`. I think it is better to start with either lower case (so the test name is readable as one long sentence) or with all capitals `WHEN` (so it is CLEAR in the test name that the context starts there)

pkg/pool-manager/pool_test.go
ormergi: ack

pkg/pool-manager/pool_test.go
ormergi: yep, It was the IDE 
fixed

pkg/pool-manager/pool_test.go
ormergi: Got it :)

pkg/pool-manager/pool_test.go
ormergi: done

ormergi: ack

pkg/pool-manager/pool_test.go
ormergi: Thanks for explaining I will change it :)

pkg/pool-manager/pool_test.go
ormergi: done

pkg/pool-manager/pool_test.go
```diff
@@ -414,4 +416,61 @@ var _ = Describe("Pool", func() {
...
+						err = json.Unmarshal([]byte(obtainedNetworksAnnotationJson), &obtainedNetworksAnnotation)
+						Expect(err).ToNot(HaveOccurred(), "should convert obtained annotation as json to multus.NetworkSelectionElement")
+
+						By("Convert expected networks annotation JSON to multus.NetworkSelectionElement array")
+						expectedNetworksAnnotation := []multus.NetworkSelectionElement{}
+						err = json.Unmarshal([]byte(networkRequestAnnotation), &expectedNetworksAnnotation)
+						Expect(err).ToNot(HaveOccurred(), "should convert expected annotation as json to multus.NetworkSelectionElement")
+
+						By("Compare between each obtained and expected network request according to the mac-address")
+						for _, obtainedNetwork := range obtainedNetworksAnnotation {
```
qinqon, line 450: Check if ContainElement or ContainElements can help you here https://onsi.github.io/gomega/#containelementselement-interface

pkg/pool-manager/pool_test.go
```diff
@@ -414,4 +416,61 @@ var _ = Describe("Pool", func() {
...
+							for _, expectedNetwork := range expectedNetworksAnnotation {
```
qinqon, line 451: Either check size before loop or convert slices into []string with the macRequest so you can use ContaineElements and it will already check the size for you.

pkg/pool-manager/pool_test.go
phoracek: @ormergi I don't see it being changed. There are still 2 contexts, the outter does not set anything.

pkg/pool-manager/pool_test.go
```diff
@@ -414,4 +416,61 @@ var _ = Describe("Pool", func() {
 		})
 	})
 
+	Describe("Multus Network Annotations API Tests", func() {
+		pod := v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "testPod", Namespace: "default"}}
+		poolManager := &PoolManager{}
```
phoracek, line 421: I'm sorry, I commented this should be in `Describe`, while it should be in the `Context` where we set it in `BeforeEach`.

pkg/pool-manager/pool_test.go
```diff
@@ -414,4 +416,61 @@ var _ = Describe("Pool", func() {
...
+						`[{"name":"ovs-conf","namespace":"default","ips":["10.10.0.1","10.10.0.2","10.0.0.3"],"mac":"02:00:00:00:00:00"}]`),
+				)
+
+				table.DescribeTable("should fail to allocate requested mac-address", func(networkRequestAnnotation string) {
+					pod.Annotations = map[string]string{networksAnnotation: networkRequestAnnotation}
+					By("Request specific mac-address by adding the address to the networks pod annotation")
+					err := poolManager.AllocatePodMac(&pod)
+					Expect(err).To(HaveOccurred(), "should fail to allocate mac address due to bad annotation format")
+				},
+					table.Entry("with ip-address request as string instead of string array",
```
phoracek, line 470: No need for a describe table if there is only a single line. It extends the complexity and brings no value.

pkg/pool-manager/pool_test.go
ormergi: We need to bump Gomega to have ContainElements https://github.com/onsi/gomega/releases

pkg/pool-manager/pool_test.go
ormergi: I will add a break for the end of the if that should prevent overlapping, once we bump Gomega  we can use ContaineElements

ormergi:
> Gomga ContainElement[s] can help here also there is a false possitive at one of the test.
> 
> I see the table test a little too much for too tests but I suppose now is easier to increase with more use cases right ?

I undersand why it may look too-much but I reached to a point I have two of the same test

pkg/pool-manager/pool_test.go
ormergi: I use DescribeTable here so I wont need to have the same exact tests with only one line change,where this line is actually the input for the test.
I added another case where we have multiple network requests, its much easier to do it like that.
Also you have the test logic is in once place so its easier to understand
what you think?

pkg/pool-manager/pool_test.go
```diff
@@ -414,4 +416,56 @@ var _ = Describe("Pool", func() {
...
+				table.Entry("with single ip-address request as string array",
+				`[{"name":"ovs-conf","namespace":"default","ips":["10.10.0.1"],"mac":"02:00:00:00:00:00"}]`),
+				table.Entry("with multiple ip-address request as string array",
+				`[{"name":"ovs-conf","namespace":"default","ips":["10.10.0.1","10.10.0.2","10.0.0.3"],"mac":"02:00:00:00:00:00"}]`),
+				table.Entry("with multiple networks requsets",`[
+						{"name":"ovs-conf","namespace":"default","ips":["10.10.0.1","10.10.0.2","10.0.0.3"],"mac":"02:00:00:00:00:00"},
+						{"name":"cnv-bridge","namespace":"openshift-cnv","ips":["192.168.66.100","192.168.66.101"],"mac":"02:F0:F0:F0:F0:F0"}
+				]`),
+			)
+			It("should fail to allocate requested mac-address, with ip-address request as string instead of string array", func(){
```
phoracek, line 462: This is something that should be tested on Multus, but I don't see how it concerns us in KMP.

pkg/pool-manager/pool_test.go
```diff
@@ -414,4 +416,56 @@ var _ = Describe("Pool", func() {
 		})
 	})
 
+	Describe("Multus Network Annotations API Tests", func() {
+		Context("when pool-manager is configured with available addresses, Pod is created with networks annotation with mac-address", func() {
```
phoracek, line 420: This is missleading. You say that "Pod is created", but it is not created in the context (before each), it is created only in the describe table only. You can drop the variable from here and keep it in DescribeTable. You should also remove the "Pod is created" from this title.

qinqon:
/lgtm
@phoracek you good to approve ?

pkg/pool-manager/pool_test.go
ormergi: This is how we check that the pool-manager will not handle request that multus wouldn't that eventually will fail

phoracek: If it will eventually fail on Multus, we should let it fail on Multus. There is a lot of other components that it may fail on later. We should not care about any of those. We should test for our failures and our positive flow. Duplicating tests of other components will leave us with unmaintainable code-base. Unless you convince me this brings any value to this project.

ormergi: This PR is about fixing this bug https://bugzilla.redhat.com/show_bug.cgi?id=1816971
and preventing another issue like that, so I added two cases where we use the new `ips` property and another case where `ips` is not valid.
This is about testing [parsePodNetworkAnnotation](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/97e9201d0609d2f3ae7f23a68b4893f5961f2113/pkg/pool-manager/pod_pool.go#L286) which is where the error comes from, if you think it is redundant I will remove it.

pkg/pool-manager/pool_test.go
phoracek: @ormergi you haven't dropped the `pod` initialization from the context. It is no longer needed there

pkg/pool-manager/pool_test.go
phoracek: I don't see how adding a specific negative test helps us. But maybe I'm missing something, I'd like to hear from the other reviewers.

@RamLavi @qinqon

qinqon: Well testing the behaviour on bad format is not a bad thing, at least we know that we don't panic.

phoracek: Fair enough then

ormergi: If this test fail its indicate we are not aligned with the API of NetworkSelectionElement we parse and use

pkg/pool-manager/pool_test.go
ormergi: done

phoracek:
Feel free to cancel the hold

PR #125, opened 2020-04-30
Revert "Deploy kubemacpool-mac-controller-manager with StatefulSet (#…

RamLavi:
…112)"

This reverts commit 33679fcc78d925f7afc9e0ac60003b0a34017de3.

```release-note
NONE
```

### DIFF
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -5,13 +5,13 @@ import (
 	"math"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/rand"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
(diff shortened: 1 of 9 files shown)


tests/tests.go
```diff
@@ -128,9 +128,9 @@ func addNetworksToPod(pod *corev1.Pod, networks []map[string]string) {
 	}
 }
 
-func findPodByUid(pods *corev1.PodList, podToFind corev1.Pod) *corev1.Pod {
```
ormergi, line 131: We can keep this change, Its more accurate when looking for a pod.

RamLavi: yes, but let's insert it in a different PR, as it's not solely related to statefulSet, and I want to remove the commit as a whole

PR #124, opened 2020-04-28
Integrate the new kube-admission-webhook that generate certificates

qinqon:
The new version of kube-admission-webhook works at k8s and openshift and also handles server/ca certificate rotations.

It also includes some infra changes 
- Update kubevirtci
- Remove multistage build

```release-note
Use self generated certificate/key for webhook
```

PR #128, opened 2020-05-04
Add opt in functionality

RamLavi:
Currently the kubemacpool operate on all non-critical namespaces by default.

We want to invert this behavior to opt-in: only namespaces that state the kubemacpool label would be served by the kmp.
In this PR we will add namespace selector to the mutating webhook manifest, and also to the controller reconcile logic. Additionally, the necessary changes will be made to tests, and other tests are added.

### DIFF
--- a/config/default/manager/manager.yaml
+++ b/config/default/manager/manager.yaml
@@ -168,6 +168,8 @@ webhooks:
         values:
           - "0"
           - "1"
+      matchLabels:
+        mutatepods.kubemacpool.io: allocateForAll
     rules:
       - operations: ["CREATE"]
         apiGroups: [""]
@@ -193,6 +195,8 @@ webhooks:
         values:
           - "0"
           - "1"
+      matchLabels:
+        mutatevirtualmachines.kubemacpool.io: allocateForAll
     rules:
       - operations: ["CREATE", "UPDATE"]
         apiGroups: ["kubevirt.io"]
--- a/config/default/rbac/rbac_role.yaml
+++ b/config/default/rbac/rbac_role.yaml
@@ -97,3 +97,9 @@ rules:
   - create
   - update
   - patch
+- apiGroups:
+    - ""
+  resources:
+    - namespaces
+  verbs:
+    - get
--- a/pkg/controller/pod/pod_controller.go
+++ b/pkg/controller/pod/pod_controller.go
@@ -21,8 +21,9 @@ import (
 	"fmt"
 
 	"github.com/intel/multus-cni/logging"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	kuberneteserror "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -76,26 +77,22 @@ type ReconcilePolicy struct {
 
 // Reconcile reads that state of the cluster for a Pod object and makes changes based on the state
 func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	log.V(1).Info("got a pod event in the controller")
+	logger := log.WithName("Reconcile").WithValues("podName", request.Name, "podNamespace", request.Namespace)
+
+	logger.V(1).Info("got a pod event in the controller")
 	instanceOptedIn, err := r.poolManager.IsPodInstanceOptedIn(request.Namespace)
 	if err != nil {
-		// Error reading the object - requeue the request.
-		log.Error(err, "failed to check opt-in selection",
-			"podName", request.Name,
-			"podNamespace", request.Namespace)
-		return reconcile.Result{}, err
+		return reconcile.Result{}, errors.Wrap(err, "failed to check opt-in selection for pod")
 	}
 	if !instanceOptedIn {
-		log.V(1).Info("pod is opted-out from kubemacpool",
-			"pod", request.Name,
-			"podNamespace", request.Namespace)
-		return reconcile.Result{}, err
+		logger.V(1).Info("pod is opted-out from kubemacpool")
+		return reconcile.Result{}, nil
 	}
 
 	instance := &corev1.Pod{}
 	err = r.Get(context.TODO(), request.NamespacedName, instance)
 	if err != nil {
-		if errors.IsNotFound(err) {
+		if kuberneteserror.IsNotFound(err) {
 			err := r.poolManager.ReleasePodMac(fmt.Sprintf("%s/%s", request.Namespace, request.Name))
 			if err != nil {
 				logging.Printf(logging.ErrorLevel, "failed to release mac for pod %s: %v", request.NamespacedName, err)
@@ -103,7 +100,7 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
-		return reconcile.Result{}, err
+		return reconcile.Result{}, errors.Wrap(err, "Failed to read the request object")
 	}
 
 	// allocate only when the pod is ready
@@ -113,7 +110,7 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 
 	err = r.poolManager.AllocatePodMac(instance)
 	if err != nil {
-		log.Error(err, "failed to allocate mac for pod")
+		logger.Error(err, "failed to allocate mac for pod")
 	}
 
 	return reconcile.Result{}, nil
--- a/pkg/controller/pod/pod_controller.go
+++ b/pkg/controller/pod/pod_controller.go
@@ -19,8 +19,8 @@ package pod
 import (
 	"context"
 	"fmt"
+
 	"github.com/intel/multus-cni/logging"
-	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -77,17 +77,17 @@ type ReconcilePolicy struct {
 // Reconcile reads that state of the cluster for a Pod object and makes changes based on the state
 func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	log.V(1).Info("got a pod event in the controller")
-	instanceOptedIn, err := r.poolManager.IsInstanceOptedIn(request.Namespace, names.NAMESPACE_OPT_IN_LABEL_PODS, "allocateForAll")
-	if !instanceOptedIn {
-		log.V(1).Info("pod is opted-out from kubemacpool",
-			"pod", request.Name,
-			"podNamespace", request.Namespace)
-		return reconcile.Result{}, err
-	}
+	instanceOptedIn, err := r.poolManager.IsPodInstanceOptedIn(request.Namespace)
 	if err != nil {
 		// Error reading the object - requeue the request.
 		log.Error(err, "failed to check opt-in selection",
 			"podName", request.Name,
+			"podNamespace", request.Namespace)
+		return reconcile.Result{}, err
+	}
+	if !instanceOptedIn {
+		log.V(1).Info("pod is opted-out from kubemacpool",
+			"pod", request.Name,
 			"podNamespace", request.Namespace)
 		return reconcile.Result{}, err
 	}
--- a/pkg/controller/pod/pod_controller.go
+++ b/pkg/controller/pod/pod_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/intel/multus-cni/logging"
+	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -76,8 +77,23 @@ type ReconcilePolicy struct {
 // Reconcile reads that state of the cluster for a Pod object and makes changes based on the state
 func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	log.V(1).Info("got a pod event in the controller")
+	instanceOptedIn, err := r.poolManager.IsInstanceOptedIn(request.Namespace, names.NAMESPACE_OPT_IN_LABEL_PODS, "allocateForAll")
+	if !instanceOptedIn {
+		log.V(1).Info("pod is opted-out from kubemacpool",
+			"pod", request.Name,
+			"podNamespace", request.Namespace)
+		return reconcile.Result{}, err
+	}
+	if err != nil {
+		// Error reading the object - requeue the request.
+		log.Error(err, "failed to check opt-in selection",
+			"podName", request.Name,
+			"podNamespace", request.Namespace)
+		return reconcile.Result{}, err
+	}
+
 	instance := &corev1.Pod{}
-	err := r.Get(context.TODO(), request.NamespacedName, instance)
+	err = r.Get(context.TODO(), request.NamespacedName, instance)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			err := r.poolManager.ReleasePodMac(fmt.Sprintf("%s/%s", request.Namespace, request.Name))
--- a/pkg/controller/virtualmachine/virtualmachine_controller.go
+++ b/pkg/controller/virtualmachine/virtualmachine_controller.go
@@ -19,7 +19,8 @@ package virtualmachine
 import (
 	"context"
 
-	"k8s.io/apimachinery/pkg/api/errors"
+	"github.com/pkg/errors"
+	kuberneteserror "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	kubevirt "kubevirt.io/client-go/api/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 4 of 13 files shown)


phoracek:
This PR has a couple of gaps compared to our requirement. Is this supposed to implement the whole story or do you plan to implement the rest in follow-up PRs?

RamLavi:
>Is this supposed to implement the whole story or do you plan to implement the rest in follow-up PRs?
upcoming commits are 
- change labels to env. variables
- fix&add tests.

I didn't find anything missing other than that - please let me know if there is

pkg/controller/pod/pod_controller.go
```diff
@@ -76,8 +77,23 @@ type ReconcilePolicy struct {
 // Reconcile reads that state of the cluster for a Pod object and makes changes based on the state
 func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	log.V(1).Info("got a pod event in the controller")
+	err, instanceOptedIn := r.poolManager.IsInstanceOptedIn(request.Namespace, names.NAMESPACE_OPT_IN_LABEL_PODS)
```
phoracek, line 80: In go, we return the error last.

config/default/manager/manager.yaml
```diff
@@ -158,6 +158,8 @@ webhooks:
         values:
           - "0"
           - "1"
+      matchLabels:
+        mutatepods.kubemacpool.io: all
```
phoracek, line 162: What does "all" mean? (looking forward for the docs)

pkg/pool-manager/pool.go
```diff
@@ -176,3 +177,24 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
 
 	return currentMac
 }
+
+// Checks if the namespace of an instance is opted in for kubemacpool
+func (p *PoolManager) IsInstanceOptedIn(namespaceName string, labelSelector string) (error, bool) {
+
```
phoracek, line 183: Let me do some nitpicking. There is no need for an empty line after you open the function. The first line is already visually divided by the identation.

pkg/pool-manager/pool.go
```diff
@@ -176,3 +177,24 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+	isNamespaceOptedIn := false
```
phoracek, line 184: There is no logic relation between this line and the next one. Please insert an empty line

pkg/pool-manager/pool.go
```diff
@@ -176,3 +177,24 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+	ns, err := p.kubeClient.CoreV1().Namespaces().Get(namespaceName, metav1.GetOptions{})
+
```
phoracek, line 186: Error handling should follow the line that returned the error. Without a space in between it is clearer who's error we handle.

pkg/pool-manager/pool.go
```diff
@@ -176,3 +177,24 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
 
 	return currentMac
 }
+
+// Checks if the namespace of an instance is opted in for kubemacpool
+func (p *PoolManager) IsInstanceOptedIn(namespaceName string, labelSelector string) (error, bool) {
```
phoracek, line 182: In Go, we return errors last

pkg/pool-manager/pool.go
```diff
@@ -176,3 +177,24 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+	ns, err := p.kubeClient.CoreV1().Namespaces().Get(namespaceName, metav1.GetOptions{})
+
+	if err != nil {
+		return err, isNamespaceOptedIn
+	}
+
+	log.V(1).Info("namespaceName Labels", "Labels", ns.GetLabels())
+	for label := range ns.GetLabels() {
+		if label == labelSelector {
+			isNamespaceOptedIn = true
```
phoracek, line 194: Would you please check the label value as well. So we make sure that users are following allowed values. Otherwise it can happen that they will set the value to "FOO". Then, once we fix this and check for a specific value, they would be surprised KMP does not work anymore.

config/default/manager/manager.yaml
RamLavi: the thought was that this label could either be specific to a network name:
mutatepods.kubemacpool.io: mynetwork
or ready for all networks: 
mutatepods.kubemacpool.io: all
or all opted out:
mutatepods.kubemacpool.io: disable
what do you think?

config/default/rbac/rbac_role.yaml
```diff
@@ -97,3 +97,9 @@ rules:
   - create
   - update
   - patch
+- apiGroups:
```
RamLavi, line 100: @phoracek btw I tried to add this via kubebuilder command such as was done in [webhook](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/8f197a045e5e092550677df8794271a8708e8a66/pkg/webhook/webhook.go#L48):
`// +kubebuilder:rbac:groups="",resources=namespaces,verbs=get`
it didn't realy work, but would you prefer I insist on it?

config/default/manager/manager.yaml
phoracek: I think that there should be additional label on the NetworkAttachmentDefinition. Having list of networks on Namespace is risky since there is no garbage collection and it would require more logic IMO.

I'm ok with `disabled`. `all` should be more specific, maybe `allocateForAll`? For other options, that does not need to be discussed now IMO.

config/default/rbac/rbac_role.yaml
phoracek: You say you have added this manually? Does not it get overwritten by generate?

config/default/manager/manager.yaml
RamLavi: ack

config/default/rbac/rbac_role.yaml
RamLavi: generate doesn't do anything to it, so I added it manually.. not sure what I'm doing wrong

phoracek: I'm ok with that.

RamLavi: ok I'll leave it without the kubebuilder remark..

tests/virtual_machines_test.go
```diff
@@ -69,11 +69,42 @@ var _ = Describe("Virtual Machines", func() {
...
+
+			for _, namespace := range []string{TestNamespace, OtherTestNamespace} {
+				err = cleanNamespaceLabels(namespace)
+				Expect(err).ToNot(HaveOccurred())
+			}
+		})
+
+		Context("When the client wants to create a vm on an opted-out namespace", func() {
+			It("should create a vm object without kubemacpool MAC address assignment when no label is set to namespace", func() {
+				err := initKubemacpoolParams(rangeStart, rangeEnd, map[string]string{})
```
phoracek, line 81: This should be in a BeforeEach of the `Context` to set up the context. I know it is not the case in the rest of the module, but let's take this opportunity to write it properly.

tests/virtual_machines_test.go
```diff
@@ -69,11 +69,42 @@ var _ = Describe("Virtual Machines", func() {
...
+
+				vm := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br", "")},
+					[]kubevirtv1.Network{newNetwork("br")})
+
+				err = testClient.VirtClient.Create(context.TODO(), vm)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress).Should(Equal(""))
+			})
+
+			It("should create a vm object without kubemacpool MAC address assignment when disabled label is set to namespace", func() {
```
phoracek, line 92: This is a different context. Context describes what is the environment Context: "When KMP is disabled on a namespace", It: "should create a VM object without a MAC assigned".

pkg/pool-manager/pool.go
```diff
@@ -176,3 +191,38 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
 
 	return currentMac
 }
+
+// Checks if the namespace of an instance is opted in for kubemacpool
+func (p *PoolManager) IsInstanceOptedIn(namespaceName, labelSelector, labelValue string) (bool, error) {
```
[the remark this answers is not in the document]
RamLavi: I like your thinking, but I tried to find something like this and didn't come up with anything that does exactly this... closest thing I found was the [apimachinery code](https://github.com/kubernetes/apimachinery/blob/15d95c0b2af3f4fcf46dce24105e5fbb9379af5a/pkg/labels/selector.go#L198) that handles the label selection, but it seems like it's using internal parameters and not set out for external use.

tests/tests.go
```diff
@@ -188,6 +175,38 @@ func setRange(rangeStart, rangeEnd string) error {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func initKubemacpoolParams(rangeStart, rangeEnd string, namespaceLabelMap map[string]string) error {
+	err := setRangeInRangeConfigMap(rangeStart, rangeEnd)
+	Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
RamLavi: do you mean `By("Should succeed setting range in the range config map")`?

tests/virtual_machines_test.go
RamLavi: Not sure I understand. 
The reason these parameters are not in BeforeEach is that they change on *each* test.
please elaborate

tests/virtual_machines_test.go
```diff
@@ -69,11 +69,42 @@ var _ = Describe("Virtual Machines", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				Expect(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress).Should(Equal(""))
+			})
 		})
 
-		Context("When the client wants to create a vm", func() {
+		Context("When the client wants to create a vm on an opted-in namespace", func() {
 			It("should create a vm object and automatically assign a static MAC address", func() {
-				err := setRange(rangeStart, rangeEnd)
+				err := initKubemacpoolParams(rangeStart, rangeEnd, map[string]string{names.NAMESPACE_OPT_IN_LABEL_VMS: "allocateForAll"})
```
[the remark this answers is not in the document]
RamLavi: ack

tests/virtual_machines_test.go
RamLavi: ok I think I got what you mean

pkg/pool-manager/pool.go
```diff
@@ -72,16 +76,27 @@ func NewPoolManager(kubeClient kubernetes.Interface, rangeStart, rangeEnd net.Ha
...
+		isLeader:           false,
+		isKubevirt:         kubevirtExist,
+		rangeEnd:           rangeEnd,
+		rangeStart:         rangeStart,
+		currentMac:         currentMac,
+		managerNamespace:   managerNamespace,
+		podToMacPoolMap:    map[string]map[string]string{},
+		macPoolMap:         map[string]AllocationStatus{},
+		poolMutex:          sync.Mutex{},
+		PodOptInLabelValue: podOptInValue,
```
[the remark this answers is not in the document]
RamLavi: ack

tests/tests.go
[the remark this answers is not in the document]
phoracek: No. You can add an explanation to `Expect`. In case it fails, it would be clear which expectation has not been fulfilled.

`Expect(err).ToNot(HaveOccurred(), "Should succeed setting range in the KubeMacPool config map")`

pkg/pool-manager/pool.go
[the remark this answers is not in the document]
phoracek: We already use apimachinery everywhere. I think it is quite safe to use it here.

```
pkg/apis/apis.go:       "k8s.io/apimachinery/pkg/runtime"
pkg/controller/pod/pod_controller.go:   "k8s.io/apimachinery/pkg/api/errors"
pkg/controller/pod/pod_controller.go:   "k8s.io/apimachinery/pkg/runtime"
pkg/controller/virtualmachine/virtualmachine_controller.go:     "k8s.io/apimachinery/pkg/api/errors"
pkg/controller/virtualmachine/virtualmachine_controller.go:     "k8s.io/apimachinery/pkg/runtime"
pkg/manager/leaderelection.go:  metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
pkg/manager/leaderelection.go:  "k8s.io/apimachinery/pkg/runtime"
pkg/manager/recorder.go:        "k8s.io/apimachinery/pkg/runtime"
pkg/pool-manager/pod_pool.go:   metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
pkg/pool-manager/pool_test.go:  "k8s.io/apimachinery/pkg/runtime"
pkg/pool-manager/pool_test.go:  metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
pkg/pool-manager/virtualmachine_pool.go:        "k8s.io/apimachinery/pkg/api/errors"
pkg/pool-manager/virtualmachine_pool.go:        metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
pkg/webhook/pod/pod.go: "k8s.io/apimachinery/pkg/apis/meta/v1"
pkg/webhook/virtualmachine/virtualmachine.go:   "k8s.io/apimachinery/pkg/api/errors"
pkg/webhook/virtualmachine/virtualmachine.go:   v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
pkg/webhook/webhook.go: "k8s.io/apimachinery/pkg/api/errors"
pkg/webhook/webhook.go: metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
pkg/webhook/webhook.go: "k8s.io/apimachinery/pkg/util/intstr"
tests/tests.go: "k8s.io/apimachinery/pkg/util/rand"
tests/tests.go: "k8s.io/apimachinery/pkg/api/errors"
tests/tests.go: metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
tests/tests_suite_test.go:      metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
tests/virtual_machines_test.go: "k8s.io/apimachinery/pkg/api/errors"
```

tests/tests.go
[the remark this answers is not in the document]
RamLavi: got it. thx

pkg/pool-manager/pool.go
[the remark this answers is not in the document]
RamLavi: yeah, but there is no apimachinery method that checks if there there is a namespace with this label.. 
[NamespaceSelector](https://github.com/kubernetes/api/blob/63b8484e576699c2f714dd6319c827a568a95090/admissionregistration/v1/types.go#L260) is of type *metav1.LabelSelector which is manged in an internal sizedBuffer...
I'm not sure how to find such a function, should one exist.. help?

pkg/controller/pod/pod_controller.go
```diff
@@ -76,8 +77,23 @@ type ReconcilePolicy struct {
 // Reconcile reads that state of the cluster for a Pod object and makes changes based on the state
 func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	log.V(1).Info("got a pod event in the controller")
+	instanceOptedIn, err := r.poolManager.IsPodInstanceOptedIn(request.Namespace)
```
qinqon, line 80: I would say that we check the error first so we are sure that the instanceOptedIn content is legit.

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -82,8 +82,24 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 	log.V(1).Info("got a virtual machine event in the controller",
 		"virtualMachineName", request.Name,
 		"virtualMachineNamespace", request.Namespace)
+
+	instanceOptedIn, err := r.poolManager.IsVmInstanceOptedIn(request.Namespace)
+	if !instanceOptedIn {
```
qinqon, line 87: ditto

pkg/pool-manager/pool.go
```diff
@@ -176,3 +191,41 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+	log.V(3).Info("checking for label in namespace", "webhook labelSelector", labelSelector, "webhook labelValue", labelValue)
+	isNamespaceOptedIn := false
+
+	ns, err := p.kubeClient.CoreV1().Namespaces().Get(namespaceName, metav1.GetOptions{})
+	if err != nil {
+		return isNamespaceOptedIn, err
+	}
+
+	log.V(3).Info("namespaceName Labels", "namespaceName", namespaceName, "Labels", ns.GetLabels())
+	if value := ns.GetLabels()[labelSelector]; labelValue == value {
```
qinqon, line 206: This will panic if labelSelector is not there, can you do `value, ok := ...` and then `ok && labelValue == value` so we prevent panics ?

pkg/pool-manager/pool.go
```diff
@@ -176,3 +191,41 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+	return isNamespaceOptedIn, err
+}
+
+func getOptInLabelValueFromMutatingWebhookConfiguration(kubeClient kubernetes.Interface, mutatingWebhookConfigName, webhookName, optInLabel string) (string, error) {
+	if mutatingWebhookConfigName == "" {
+		return "", nil
+	}
+	mutatingWebhookConfiguration, err := kubeClient.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(mutatingWebhookConfigName, metav1.GetOptions{})
+	if err != nil {
+		return "", err
```
qinqon, line 219: Add some context to err with errors.Wrap(err, "...) make easier to debug it.

pkg/pool-manager/pool.go
```diff
@@ -176,3 +191,41 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+
+	for _, webhook := range mutatingWebhookConfiguration.Webhooks {
+		if webhook.Name == webhookName {
+			if value := webhook.NamespaceSelector.MatchLabels[optInLabel]; value != "" {
+				return value, nil
+			}
+		}
+	}
+
+	return "", err
```
qinqon, line 230: ditto

pkg/pool-manager/pool.go
```diff
@@ -176,3 +191,41 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+		return "", nil
+	}
+	mutatingWebhookConfiguration, err := kubeClient.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(mutatingWebhookConfigName, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+
+	for _, webhook := range mutatingWebhookConfiguration.Webhooks {
+		if webhook.Name == webhookName {
+			if value := webhook.NamespaceSelector.MatchLabels[optInLabel]; value != "" {
```
qinqon, line 224: check key existence

pkg/pool-manager/pool.go
RamLavi: thought of this but ns.[GetLabels](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/meta.go#L54)() doesn't return error. what would you suggest?

qinqon: I don't mean error I mean golang maps, to prevent panic you have to return also the boolean flag that denotes key presence
```golang
if val, ok := dict["foo"]; ok {
    //do something here
}
```

RamLavi: got you.

pkg/pool-manager/pool.go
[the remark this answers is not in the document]
phoracek: The label selector field `matchLabels` in `MutatingWebhookConfiguration` is of type `*metav1.LabelSelector`. You can convert that to `labels.Selector` by using https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelectorAsSelector. And finally, you can use `labels.Selector`'s `Matches` that you have mentioned here.

So something like this should do the job:

```go
metav1.LabelSelectorAsSelector(webhook.NamespaceSelector).Matches(namespaceLabels)
```

RamLavi:
>make sense to do this check instead of Reconcile at watch function ?
@qinqon - the point is to do this in the controller's Reconsile() to abort the reconsile if the namespace of the pod/vm is not opted in.. 
I would have preferred to use NamespaceSelector on the controller like I did with the webhook, but AFAIK it's not supported

pkg/controller/pod/pod_controller.go
```diff
@@ -76,8 +77,23 @@ type ReconcilePolicy struct {
...
+		log.Error(err, "failed to check opt-in selection",
+			"podName", request.Name,
+			"podNamespace", request.Namespace)
+		return reconcile.Result{}, err
+	}
+	if !instanceOptedIn {
+		log.V(1).Info("pod is opted-out from kubemacpool",
+			"pod", request.Name,
+			"podNamespace", request.Namespace)
+		return reconcile.Result{}, err
```
qinqon, line 92: Better be explicit `return reconcile.Result{}, nil`

pkg/controller/pod/pod_controller.go
```diff
@@ -76,8 +77,23 @@ type ReconcilePolicy struct {
 // Reconcile reads that state of the cluster for a Pod object and makes changes based on the state
 func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	log.V(1).Info("got a pod event in the controller")
+	instanceOptedIn, err := r.poolManager.IsPodInstanceOptedIn(request.Namespace)
+	if err != nil {
+		// Error reading the object - requeue the request.
+		log.Error(err, "failed to check opt-in selection",
+			"podName", request.Name,
+			"podNamespace", request.Namespace)
+		return reconcile.Result{}, err
```
qinqon, line 86: Add context to `return reconcile.Result{}, errors.Wrapf(err, "failed to check opt-in selection for pod %s/%s", request.Namespace, request.Name)`

pkg/controller/pod/pod_controller.go
```diff
@@ -76,8 +77,23 @@ type ReconcilePolicy struct {
 // Reconcile reads that state of the cluster for a Pod object and makes changes based on the state
 func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	log.V(1).Info("got a pod event in the controller")
```
qinqon, line 79: Add the pod name and namespace to a logger so whenever we log something it appears there.
`logger := log.WithName("Reconcile").WithValues("podName", request.Name, "podNamespace", request.Namespace)`

Then we do normal

`logger.V(1).Info("got a pod event in the controller")` 

and the function, an pod name and namespace will appear at every log.

pkg/controller/pod/pod_controller.go
```diff
@@ -76,8 +77,23 @@ type ReconcilePolicy struct {
...
+	instanceOptedIn, err := r.poolManager.IsPodInstanceOptedIn(request.Namespace)
+	if err != nil {
+		// Error reading the object - requeue the request.
+		log.Error(err, "failed to check opt-in selection",
```
qinqon, line 83: No need for this, controller-runtime will print the log for us, just do the errors.Wrap thing

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -82,8 +82,24 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 	log.V(1).Info("got a virtual machine event in the controller",
 		"virtualMachineName", request.Name,
 		"virtualMachineNamespace", request.Namespace)
+
+	instanceOptedIn, err := r.poolManager.IsVmInstanceOptedIn(request.Namespace)
+	if err != nil {
+		// Error reading the object - requeue the request.
+		log.Error(err, "failed to check opt-in selection",
```
qinqon, line 89: ditto

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -82,8 +82,24 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
...
+			"virtualMachineName", request.Name,
+			"virtualMachineNamespace", request.Namespace)
+		return reconcile.Result{}, err
+	}
+	if !instanceOptedIn {
+		log.V(1).Info("vm is opted-out from kubemacpool",
+			"virtualMachineName", request.Name,
+			"virtualMachineNamespace", request.Namespace)
+		return reconcile.Result{}, err
```
qinqon, line 98: ditto

pkg/pool-manager/pool.go
```diff
@@ -72,16 +77,27 @@ func NewPoolManager(kubeClient kubernetes.Interface, rangeStart, rangeEnd net.Ha
 	currentMac := make(net.HardwareAddr, len(rangeStart))
 	copy(currentMac, rangeStart)
 
+	podOptInValue, err := getOptInLabelValueFromMutatingWebhookConfiguration(kubeClient, mutatingWebhookConfigName, "mutatepods.kubemacpool.io", names.NAMESPACE_OPT_IN_LABEL_PODS)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get Namespace selector pod label value from MutatingWebhookConfiguration %v", err)
```
qinqon, line 82: `errors.Wrap(err, "failed to get Namespace selector pod label value from MutatingWebhookConfiguration"`

pkg/pool-manager/pool.go
```diff
@@ -72,16 +77,27 @@ func NewPoolManager(kubeClient kubernetes.Interface, rangeStart, rangeEnd net.Ha
...
+	}
+	vmOptInValue, err := getOptInLabelValueFromMutatingWebhookConfiguration(kubeClient, mutatingWebhookConfigName, "mutatevirtualmachines.kubemacpool.io", names.NAMESPACE_OPT_IN_LABEL_VMS)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get Namespace selector vm label value from MutatingWebhookConfiguration %v", err)
```
qinqon, line 86: ditto

pkg/pool-manager/pool.go
```diff
@@ -176,3 +192,42 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
 }
+
+// Checks if the namespace of an instance is opted in for kubemacpool
+func (p *PoolManager) isInstanceOptedIn(namespaceName, labelSelector, labelValue string) (bool, error) {
+	log.V(3).Info("checking for label in namespace", "webhook labelSelector", labelSelector, "webhook labelValue", labelValue)
+	isNamespaceOptedIn := false
+
+	ns, err := p.kubeClient.CoreV1().Namespaces().Get(namespaceName, metav1.GetOptions{})
+	if err != nil {
+		return isNamespaceOptedIn, err
```
qinqon, line 203: Add context to errors with `errors.Wrap` so it's easier to debug and see where the error comes from.

pkg/pool-manager/pool.go
```diff
@@ -176,3 +192,42 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+	}
+
+	log.V(3).Info("namespaceName Labels", "namespaceName", namespaceName, "Labels", ns.GetLabels())
+	if value, ok := ns.GetLabels()[labelSelector]; ok && labelValue == value {
+		isNamespaceOptedIn = true
+	}
+
+	return isNamespaceOptedIn, err
```
qinqon, line 211: ditto

pkg/pool-manager/pool.go
```diff
@@ -176,3 +192,42 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+
+	for _, webhook := range mutatingWebhookConfiguration.Webhooks {
+		if webhook.Name == webhookName {
+			if labelValue, ok := webhook.NamespaceSelector.MatchLabels[optInLabel]; ok && labelValue != "" {
+				return labelValue, nil
+			}
+		}
+	}
+
+	return "", errors.Wrap(err, fmt.Sprintf("label %s not found in mutatingWebhookConfig %s", optInLabel, mutatingWebhookConfigName))
```
qinqon, line 232: Do this like `errors.Wrapf(err, "label %s not found in mutatingWebhookConfig %s", optInLabel, mutatingWebhookConfigName)`

pkg/controller/pod/pod_controller.go
RamLavi: neat!

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -79,32 +80,37 @@ type ReconcilePolicy struct {
 
 // Reconcile reads that state of the cluster for a virtual machine object and makes changes based on the state
 func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	log.V(1).Info("got a virtual machine event in the controller",
-		"virtualMachineName", request.Name,
-		"virtualMachineNamespace", request.Namespace)
+	log = logf.Log.WithName("VirtualMachine Controller").WithValues("virtualMachineName", request.Name, "virtualMachineNamespace", request.Namespace)
```
qinqon, line 83: Use different variable, you don't want to replace global log here.

RamLavi: but it's [set](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/3fe1379b3c1744976ee89258dde900c964971294/pkg/controller/virtualmachine/virtualmachine_controller.go#L38) in the start of the module. Should I move it from there?

qinqon: You can use the global log to generate a new one so we have the WithName just do the following
 ```golang
logger := log.WithName("Reconcile").WithValues("virtualMachineName", request.Name, "virtualMachineNamespace", request.Namespace)
```
And use logger at that function, I think is bot "names" will appear "VirtualMachine Controller" and "Reconcile"

RamLavi: ack, (cool!)

RamLavi:
/hold
let's retest after #124 is merged - I want to make sure CI passes smoothly

PR #131, opened 2020-05-05
Remove 1.17.0 provider reference

qinqon:
It also make possible to specify different KUBEVIRT_PROVIDER

### DIFF
--- a/automation/check-patch.e2e-k8s.sh
+++ b/automation/check-patch.e2e-k8s.sh
@@ -6,7 +6,7 @@ teardown() {
 }
 
 main() {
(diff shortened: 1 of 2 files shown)


automation/check-patch.e2e-k8s.sh
```diff
@@ -6,7 +6,6 @@ teardown() {
 }
 
 main() {
-    export KUBEVIRT_PROVIDER='k8s-1.17.0'
```
phoracek, line 9: This should stay. The lane is explicitly pinned to a certain version. Using the default from another module is implicit.

phoracek:

Only fails on the flaky connection refused

PR #133, opened 2020-05-06
Replace multistage build by vendored golang

qinqon:
We have already the vendored golang so let's remove indirections
and simplify project by just use it to compile since the only
dep here is the golang runtime.

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -68,9 +68,12 @@ docker-goveralls: docker-builder docker-test
...
 	DOCKER_BASE_IMAGE=${REGISTRY}/${IMG}:kubemacpool_builder ./hack/run.sh
 
+manager: $(GO)
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GO) build -o $(BIN_DIR)/manager github.com/k8snetworkplumbingwg/kubemacpool/cmd/manager
+
 # Build the docker image
-docker-build: docker-builder
-	docker build . -t ${REGISTRY}/${IMG}:${IMAGE_TAG} --build-arg GO_VERSION=1.12.12
+container: manager
+	docker build build/ -t ${REGISTRY}/${IMG}:${IMAGE_TAG}
```
RamLavi, line 76: don't we want to keep the same GO_VERSION? (more asking really)

qinqon: If we remove the bugfix level we will get the latest 1.12 with latest fixes, so it's better like that the version is now taken from go.mod at hack/install-go.sh

RamLavi: cool, thanks

PR #134, opened 2020-05-06
Vendor kustomize

qinqon:
It install all the needed tools from ./vendor directory using the instdaled golang version, it also add
some Makefile targest so they get installed if does not exist.

```Makefile
$(GOVERALLS): go.mod
	$(MAKE) tools-vendoring
```

This simplify the projects by removing docker container indirections and maintenance and also reduce build time, we can do that since all the tools are golang.

Looks like RBAC is missing some changes, I have add them here.

**Release note**:

```release-note
NONE
```


RamLavi:
@qinqon  can you rebase?

Makefile
```diff
@@ -37,15 +53,15 @@ deploy: generate-deploy
 deploy-test: generate-test
 	$(KUBECTL) apply -f config/test/kubemacpool.yaml
 
-generate-deploy: manifests
-	kustomize build config/release > config/release/kubemacpool.yaml
+generate-deploy: $(KUSTOMIZE) manifests
```
RamLavi, line 55: regarding `deploy:` here [above](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/134/commits/2da21b8d68ac67b71292831313dc5180c52b8be4#diff-b67911656ef5d18c4ae36cb6741b7965R50): shouldn't it apply `config/release/kubemacpool.yaml` instead of `config/test/kubemacpool.yaml`?

ormergi: We deploy test/kubemacpol.yaml because its for development environment, and in this yaml we pass 
''debug'' flag to the logs so we get everything for debugging

Makefile
```diff
@@ -13,14 +13,30 @@ export GOBIN=$(GOROOT)/bin/
...
 
 export KUBECTL ?= cluster/kubectl.sh
 
-all: generate generate-deploy generate-test
+all: generate
 
 $(GO):
 	hack/install-go.sh $(BIN_DIR) > /dev/null
 
+$(KUSTOMIZE): go.mod
```
ormergi, line 27: Can you please separate the changes in Makefile to different commit and explain what are the new entries do?

go.mod
```diff
@@ -1,25 +1,26 @@
 module github.com/k8snetworkplumbingwg/kubemacpool
 
-go 1.12
+go 1.13
```
ormergi, line 3: I think we should align with CI team using go 1.12.8 or at lease make sure we wont break anything,
if we dont then we can even use the latest version

ormergi:
About dropping docker builder, isn't it a good practice to build inside a container so your enviroment is the same on every machine? can you please explain why it is necessary? 

About vending kustomize, I think we should figure out how to use `kubectl kustomize ..` instead,
or just stop using kustomize since we dont utilize it much.
What you guy think?
@RamLavi @phoracek

qinqon:
@ormergi since our only dependency is golang just using a installed version of goland in the project is enough, multistage add another indirection and also we need golang outside to run e2e tests.

About using kubectl kustomize then we need a version of kubectl so it's kind of the same thing maybe we can use the kubectl version from the kubevirtci provider but then we need to start it up since it's copied from the cluster.

go.mod
qinqon: Let me try to re-generate with go 1.12 maybe go mod tidy find a compatible version for libraries.

qinqon: @ormergi I have being able to to goback to go 1.12 looks like we have to use kustomize v3 [1]

[1] https://github.com/kubernetes-sigs/kustomize/blob/master/docs/INSTALL.md

I will re-organize this PR it's a mess already.

RamLavi:
>I think we should figure out how to use `kubectl kustomize ..` instead,
 or just stop using kustomize since we dont utilize it much.

@ormergi , I personally would prefer to control the kustomize version myself. using kubectl could backfire cause if in the future we need to upgrade the kubectl version we installed (for other reason) we won't have to end up with fixing kustomize things they broke while upgrading..

qinqon:
@ormergi @RamLavi I have re-organize the PR and being able to stick to golang 1.12 also remove the part about Makefile targets renaming so it's much better to review now.

Makefile
qinqon: I cannot separate them, since it will render the commit not functional, let me explain at this commit I have now.

Makefile
```diff
@@ -80,7 +80,7 @@ generate: generate-go generate-deploy generate-test
 goveralls: $(GOVERALLS)
 	GOVERALLS=$(GOVERALLS) ./hack/goveralls.sh
 
-manager: $(GO)
+manager: generate-go
```
RamLavi, line 83: I think this should be `generate` since `generate-go` does not create the manifests..

qinqon: It generates go code

RamLavi: but don't we want to also generate the manifests files?

qinqon: this just depends on minimal stuff needed to compile manager and that is the generated golang code the manifests are needed for deploy

RamLavi: ack

go.mod.orig
```diff
@@ -0,0 +1,65 @@
...
+
+go 1.13
+
+require (
+	github.com/containernetworking/cni v0.6.0 // indirect
+	github.com/go-logr/logr v0.1.0
+	github.com/go-logr/zapr v0.1.1 // indirect
+	github.com/imdario/mergo v0.3.7 // indirect
+	github.com/intel/multus-cni v0.0.0-20190127194101-cd6f9880ac19
+<<<<<<< HEAD
```
phoracek, line 11: Unresolved merge conflict

go.mod.orig
```diff
@@ -0,0 +1,65 @@
+module github.com/k8snetworkplumbingwg/kubemacpool
```
phoracek: Why do we need this file?

go.mod
```diff
@@ -8,24 +8,25 @@ require (
 	github.com/go-logr/zapr v0.1.1 // indirect
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/intel/multus-cni v0.0.0-20190127194101-cd6f9880ac19
-	github.com/onsi/ginkgo v1.10.1
-	github.com/onsi/gomega v1.7.0
-	github.com/pkg/errors v0.9.1 // indirect
+	github.com/mattn/goveralls v0.0.5
```
phoracek, line 11: Please comment in the commit  message what are you doing and why/

Makefile
```diff
@@ -13,14 +13,30 @@ export GOBIN=$(GOROOT)/bin/
 export PATH := $(GOBIN):$(PATH)
 GOFMT := $(GOBIN)/gofmt
 GO := $(GOBIN)/go
+KUSTOMIZE := $(GOBIN)/kustomize
```
phoracek, line 16: Please ellaborate in the commit message what are you changing and why

Makefile
```diff
@@ -92,7 +104,30 @@ cluster-down:
...
+	fmt \
+	vet \
+	goveralls \
+	manager \
+	container \
+	push \
+	cluster-up \
+	cluster-down \
+	cluster-sync \
+	tools \
```
phoracek, line 133: There should be no escape here

RamLavi:
/hold

just want to make sure #136 fails before this is merged

PR #136, opened 2020-05-07
adding check that all genereated files are commited before performing automation tests

RamLavi:

currently there is no check that makes sure that all generated manifests are committed.
In order to fix this we will add it in travis test like they do in [kubevirt](https://github.com/kubevirt/kubevirt/blob/6e87ec791e03fdb9899bbe85bbdd441a72c534ab/.travis.yml#L19).
 
**Release note**:

```release-note
NONE
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ services:
 
 install: true
 script:
-  - make container
+  - make check
 
 deploy:
 - provider: script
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -9,6 +9,10 @@ services:
 install: true
 script:
   - make container
+  - make generate
```
qinqon, line 12: let's do a `make check` that just calls hack/check.sh and the script do 
```bash
/bin/env bash

set -xe

make container generate
if [[ -n "$(git status --porcelain)" ]] ;
    then echo "It seems like you need to run `make generate`. Please run it and commit the changes"; git status --porcelain; false;
i
```
This way we can move away from Travis easily

RamLavi: ack

PR #145, opened 2020-05-14
Paramtereize manager namespaces used by functest

qinqon:
To run functest at HCO provided cluster we need to be able to change the namespace to map with clustern-network-addons-operator one and also filter pods by app=kubemacpool they don't mess with not kubemacpool pods

**Release note**:

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ On releases v0.8.4 and above, kubemacpool is set to apply on pods/vms that resid
 - `mutatepods.kubemacpool.io=allocateForAll` - to opt in pods mac allocation in your namespace
 - `mutatevirtualmachines.kubemacpool.io=allocateForAll` - to opt in vms mac allocation in your namespace
 
-#### How to enable/disable kubemacpool for a namespace 
+#### How to enable/disable kubemacpool for a namespace
 
 Kubemacpool is disabled by default on a new namespace.
 To enable kubemacpool on a specific namespace:
@@ -86,7 +86,7 @@ webhooks:
     matchLabels:
       mutatevirtualmachines.kubemacpool.io: allocateForAll
  ...
-``` 
+```
 
 **note:** if the kubemacpool's mutatingwebhookconfiguration `kubemacpool-mutator` namespace-selector value per vm/pod is set to `allocateForAll`, then you can also opt-out your namespace by setting the label value to `disable` in your namespace:
 ```bash
@@ -305,4 +305,15 @@ make cluster-down
 
 # Run function test
 make functest
+```
+
+In just the functest want to be deployed at pre-deployed external cluster
+the namespace where kubemacpool is living can be parameterize at functest
+
+```bash
+# If kubemacpool is living at HCO
+MANAGER_NAMESPACE=kubevirt-hyperconverged make functest
+
+# If kubemacpool is living at CNAO
+MANAGER_NAMESPACE=cluster-network-addons make functest
 ```
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"fmt"
 	"math"
+	"os"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/rand"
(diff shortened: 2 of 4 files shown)


tests/tests.go
```diff
@@ -331,6 +335,14 @@ func addLabelsToNamespace(namespace string, labels map[string]string) error {
 	return err
 }
 
+func getEnv(key, fallback string) string {
```
ormergi, line 338: I think we should remove `fallback` parameter and initialize [this](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/145/commits/74ecb40c55a63216809d28e1fa17f17714e2a64c#diff-4de094b4f2274c29e9f11102da4ca7d0R39) with the `names.MANAGER_NAMESAPCE

tests/tests.go
```diff
@@ -28,14 +29,14 @@ import (
...
 	OtherTestNamespace     = "kubemacpool-test-alternative"
-	ManagerNamespce        = names.MANAGER_NAMESPACE
 	nadPostUrl             = "/apis/k8s.cni.cncf.io/v1/namespaces/%s/network-attachment-definitions/%s"
 	linuxBridgeConfCRD     = `{"apiVersion":"k8s.cni.cncf.io/v1","kind":"NetworkAttachmentDefinition","metadata":{"name":"%s","namespace":"%s"},"spec":{"config":"{ \"cniVersion\": \"0.3.1\", \"type\": \"bridge\", \"bridge\": \"br1\"}"}}`
 	podNamespaceOptInLabel = "mutatepods.kubemacpool.io"
 	vmNamespaceOptInLabel  = "mutatevirtualmachines.kubemacpool.io"
 )
 
 var (
+	managerNamespace         = getEnv("MANAGER_NAMESPACE", names.MANAGER_NAMESPACE)
```
ormergi, line 39: https://github.com/k8snetworkplumbingwg/kubemacpool/pull/145/commits/74ecb40c55a63216809d28e1fa17f17714e2a64c#r425070907

tests/tests.go
qinqon: This is a generic getEnv function, we don't want to harcode it here, or do you mean to do getManagerNamespace function ?

ormergi:
Hey I just realize we already have the manager namespace as an environment variable
we can take it from the pods like this: 
````
pods := getKubemacpoolPods
	for envVar := range pods.Items[0].Spec.Containers[0].Env {
		if envVar.Name == "POD_NAMESPCAE" {
			managerNamespce = envVar.Value
		}
	}
````
What do you think?

**Comment by qinqon** [comment]
@ormergi this code is for golang code that runs inside the pod, not at the e2e tests.

File: README.md [hunk 3]
```diff
@@ -306,3 +306,14 @@ make cluster-down
 # Run function test
 make functest
 ```
+
+In just the functest want to be deployed at pre-deployed external cluster
```
ormergi, line 310: How about this wording?
 "If you would like run e2e tests on an already deployed external cluster.."
or "In order to run e2e tests on external cluster.."

tests/tests.go
```diff
@@ -28,14 +29,16 @@ import (
...
 	nadPostUrl             = "/apis/k8s.cni.cncf.io/v1/namespaces/%s/network-attachment-definitions/%s"
 	linuxBridgeConfCRD     = `{"apiVersion":"k8s.cni.cncf.io/v1","kind":"NetworkAttachmentDefinition","metadata":{"name":"%s","namespace":"%s"},"spec":{"config":"{ \"cniVersion\": \"0.3.1\", \"type\": \"bridge\", \"bridge\": \"br1\"}"}}`
 	podNamespaceOptInLabel = "mutatepods.kubemacpool.io"
 	vmNamespaceOptInLabel  = "mutatevirtualmachines.kubemacpool.io"
 )
 
 var (
+	// The MANAGER_NAMESPACE variables is passed to the test suite in case
+	// different kmp manager is at not default namespace
+	managerNamespace         = getEnv("MANAGER_NAMESPACE", names.MANAGER_NAMESPACE)
```
ormergi, line 41: Since this function is falling back to a default value how about we change its signature to 
`tryGetEnvVar(target, default string)`?

PR #149, opened 2020-05-21
Remove CSR's related RBAC

qinqon:
The current kube-admission-webhook library does not use CSRs anymore
so RBAC is not needed.

**Release note**:

```release-note
NONE
```


config/default/rbac/rbac_role.yaml
```diff
@@ -4,13 +4,6 @@ metadata:
   creationTimestamp: null
   name: manager-role
```
ormergi, line 5: Is that a standard name?

PR #151, opened 2020-05-22
Check manager pod readyness changing leadership at test

qinqon:
Test were failing at CNAO looks like it takes more time to reach ready
state at manager pods after changing leardership, with this change
the test check not only that there is at pod with leadership  but also
that that pod is Ready.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -14,6 +14,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


tests/virtual_machines_test.go
```diff
@@ -497,8 +497,7 @@ var _ = Describe("Virtual Machines", func() {
 				_, err = net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
 				Expect(err).ToNot(HaveOccurred())
 
-				By("deleting leader manager")
-				DeleteLeaderManager()
+				ChangeManagerLeadership()
```
ormergi, line 500: Please add that line with 'By', in case something goes wrong in tests suite we will know 
where it comes from

tests/tests.go
```diff
@@ -14,6 +14,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
```
RamLavi, line 20: I don't understand why the original code is not sufficient. 

I mean, 
* the old leader was deleted for sure (otherwise we would have failed the Eventually [here](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/151/commits/e7e4217b541adcc0db556ef99be20cd0c3275e14#diff-4de094b4f2274c29e9f11102da4ca7d0L231))
* The pod is checked [here](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/151/commits/e7e4217b541adcc0db556ef99be20cd0c3275e14#diff-4de094b4f2274c29e9f11102da4ca7d0L245) for readiness (second part of if condition)
* the're no way that the non-leader pod got ready.
 
can you specify why exactly it did not work?

qinqon: Leader pod can exists and have the leader label but not be ready yet, so we continue with a non functional leader.

RamLavi: but isn't the [second condition](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/151/files#diff-4de094b4f2274c29e9f11102da4ca7d0L263) enough to check readiness?

RamLavi: issue is there is a time when we have three kmp pods: two running and one terminating
and Terminating is at Ready since webhook is already running there

So there is a point to this PR.

qinqon:
Just to be sure, I have reproduce the issue at asciiname 

https://asciinema.org/a/i04XTIqM9M50ajsMhpBTYwZzP

tests/virtual_machines_test.go
qinqon: Added inside `ChangeManagerLeadership`

PR #152, opened 2020-05-24
add additional wait time in tests that check failed vm creation scenarios

RamLavi:

When a vm creation fails, the allocated mac stays in the configmap until a set wait-time - and then released in [vmWaitingCleanupLook() routine](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/2982d1a7646a75eab32fa0e831a7f81b643d8c6e/pkg/pool-manager/virtualmachine_pool.go#L436).
This wait-time is given as an argument to the kubemacpool container in the deployment manifest.
Specifically, the wait-time in production manifests is bigger than the time we wait for the vm to be released in tests.
In order to fix that, we now read the wait-time from kubemacpool deployment and consider it during the relevant tests.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"regexp"
+	"strconv"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/rand"
@@ -11,6 +13,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -28,12 +31,13 @@ import (
 )
 
 const (
-	TestNamespace          = "kubemacpool-test"
-	OtherTestNamespace     = "kubemacpool-test-alternative"
-	nadPostUrl             = "/apis/k8s.cni.cncf.io/v1/namespaces/%s/network-attachment-definitions/%s"
-	linuxBridgeConfCRD     = `{"apiVersion":"k8s.cni.cncf.io/v1","kind":"NetworkAttachmentDefinition","metadata":{"name":"%s","namespace":"%s"},"spec":{"config":"{ \"cniVersion\": \"0.3.1\", \"type\": \"bridge\", \"bridge\": \"br1\"}"}}`
-	podNamespaceOptInLabel = "mutatepods.kubemacpool.io"
-	vmNamespaceOptInLabel  = "mutatevirtualmachines.kubemacpool.io"
+	TestNamespace           = "kubemacpool-test"
+	OtherTestNamespace      = "kubemacpool-test-alternative"
+	nadPostUrl              = "/apis/k8s.cni.cncf.io/v1/namespaces/%s/network-attachment-definitions/%s"
+	linuxBridgeConfCRD      = `{"apiVersion":"k8s.cni.cncf.io/v1","kind":"NetworkAttachmentDefinition","metadata":{"name":"%s","namespace":"%s"},"spec":{"config":"{ \"cniVersion\": \"0.3.1\", \"type\": \"bridge\", \"bridge\": \"br1\"}"}}`
+	podNamespaceOptInLabel  = "mutatepods.kubemacpool.io"
+	vmNamespaceOptInLabel   = "mutatevirtualmachines.kubemacpool.io"
+	deploymentContainerName = "manager"
 )
 
 var (
@@ -226,6 +230,38 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
 	Expect(err).ToNot(HaveOccurred(), "Should succeed resetting the kubemacpool pods")
 
 	return nil
+}
+
+func getWaitTimeValueFromArguments(args []string) (time.Duration, bool) {
+	r := regexp.MustCompile(fmt.Sprintf("--%s=(\\d+)", names.WAIT_TIME_ARG))
+	for _, arg := range args {
+		match := r.FindStringSubmatch(arg)
+		if match != nil {
+			waitTimeValue, err := strconv.Atoi(match[1])
+			Expect(err).ToNot(HaveOccurred(), "Should successfully parse wait-time argument")
+			return time.Duration(waitTimeValue) * time.Second, true
+		}
+	}
+
+	return 0, false
+}
+
+func getVmFailCleanupWaitTime() time.Duration {
+	managerDeployment := v1.Deployment{}
+	err := testClient.VirtClient.Get(context.TODO(), client.ObjectKey{Namespace: managerNamespace, Name: names.MANAGER_DEPLOYMENT}, &managerDeployment)
+	Expect(err).ToNot(HaveOccurred(), "Should successfully get manager's Deployment")
+	Expect(managerDeployment.Spec.Template.Spec.Containers).ToNot(BeEmpty(), "Manager's deployment should contain containers")
+
+	for _, container := range managerDeployment.Spec.Template.Spec.Containers {
+		if container.Name == deploymentContainerName {
+			if waitTimeDuration, found := getWaitTimeValueFromArguments(container.Args); found {
+				return waitTimeDuration
+			}
+		}
+	}
+
+	Fail(fmt.Sprintf("Failed to find wait-time argument in %s container inside %s deployment", deploymentContainerName, names.MANAGER_DEPLOYMENT))
+	return 0
 }
 
 func ChangeManagerLeadership() {
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -403,6 +403,13 @@ var _ = Describe("Virtual Machines", func() {
 		})
 		//2633
 		Context("When we re-apply a failed VM yaml", func() {
+			totalTimeout := time.Duration(0)
+			BeforeAll(func() {
+				vmFailCleanupWaitTime := getVmFailCleanupWaitTime()
+				// since this test checks vmWaitingCleanupLook routine, we nned to adjust the total timeout with the wait-time argument.
+				// we also add some extra timeout apart form wait-time to be sure that we catch the vm mac release.
+				totalTimeout = timeout + vmFailCleanupWaitTime
+			})
 			It("should allow to assign to the VM the same MAC addresses, with name as requested before and do not return an error", func() {
 				err := initKubemacpoolParams(rangeStart, rangeEnd)
 				Expect(err).ToNot(HaveOccurred())
@@ -425,7 +432,7 @@ var _ = Describe("Virtual Machines", func() {
 					}
 					return err
 
-				}, timeout, pollingInterval).ShouldNot(HaveOccurred(), "failed to apply the new vm object")
+				}, totalTimeout, pollingInterval).ShouldNot(HaveOccurred(), "failed to apply the new vm object")
 			})
 			It("should allow to assign to the VM the same MAC addresses, different name as requested before and do not return an error", func() {
 				err := initKubemacpoolParams(rangeStart, rangeEnd)
@@ -450,7 +457,7 @@ var _ = Describe("Virtual Machines", func() {
 					}
 					return err
 
-				}, timeout, pollingInterval).ShouldNot(HaveOccurred(), "failed to apply the new vm object")
+				}, totalTimeout, pollingInterval).ShouldNot(HaveOccurred(), "failed to apply the new vm object")
 			})
 		})
 
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -25,6 +25,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 
 	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/manager"
+	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 	poolmanager "github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager"
 )
 
@@ -47,7 +48,7 @@ func main() {
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&logType, "v", "production", "Log type (debug/production).")
-	flag.IntVar(&waitingTime, "wait-time", 600, "waiting time to release the mac if object was not created")
+	flag.IntVar(&waitingTime, names.WAIT_TIME_ARG, 600, "waiting time to release the mac if object was not created")
 	flag.Parse()
 
 	if logType == "debug" {
--- a/pkg/names/names.go
+++ b/pkg/names/names.go
@@ -19,3 +19,5 @@ const K8S_RUNLABEL = "runlevel"
 const OPENSHIFT_RUNLABEL = "openshift.io/run-level"
 
 const WAITING_VMS_CONFIGMAP = "kubemacpool-vm-configmap"
+
+const WAIT_TIME_ARG = "wait-time"


tests/tests.go
```diff
@@ -209,6 +209,22 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
...
+
+	for _, arg := range managerDeployment.Spec.Template.Spec.Containers[0].Args {
+		waitTime := 0
+		if elementsParsed, err := fmt.Sscanf(arg, "--wait-time=%d", &waitTime); elementsParsed == 1 {
+			Expect(err).ToNot(HaveOccurred(), "Should succeed parsing wait-time arg")
+			return time.Duration(waitTime) * time.Second
+		}
+	}
+
+	return 0
```
RamLavi, line 225: I'm not sure if I should return an Error if this argument is not there. what do you think?

tests/tests.go
```diff
@@ -209,6 +209,22 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
...
 }
 
+func getVmFailCleanupWaitTime() time.Duration {
+	managerDeployment, err := testClient.KubeClient.AppsV1().Deployments(managerNamespace).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
+	Expect(err).ToNot(HaveOccurred(), "Should succeed getting Deployment object")
+	Expect(len(managerDeployment.Spec.Template.Spec.Containers)).To(Equal(1))
+
+	for _, arg := range managerDeployment.Spec.Template.Spec.Containers[0].Args {
+		waitTime := 0
+		if elementsParsed, err := fmt.Sscanf(arg, "--wait-time=%d", &waitTime); elementsParsed == 1 {
```
ormergi, line 219: Please split this line to two lines one for initialization and one for checking so it will be easier to review :)

tests/tests.go
```diff
@@ -209,6 +209,22 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
 	return nil
 }
 
+func getVmFailCleanupWaitTime() time.Duration {
+	managerDeployment, err := testClient.KubeClient.AppsV1().Deployments(managerNamespace).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
+	Expect(err).ToNot(HaveOccurred(), "Should succeed getting Deployment object")
+	Expect(len(managerDeployment.Spec.Template.Spec.Containers)).To(Equal(1))
```
ormergi, line 215: I think we should check if the manager container is there by its name instead of assuming there is one and only container in case we will add more containers in the future.

something like:
```
managerContainerName="manager"
...
Expect(len(managerDeployment.Spec.Template.Spec.Containers) >= 1).To(BeTrue())

managerContainer := corev1.Container{}
for container := managerDeployment.Spec.Template.Spec.Containers {
    if managerContainer.Name == managerContainerName {
       managerContainer = container
       break
    }
}
Expect(managerContainer).ToNot(BeEmpty())
Expect(managerContainer.Name).To(Equal(managerContainerName))
...
```
Also this code will be more reliable by not access an object by index|
(like we saw in other places in the code that this approach is prune to errors)

tests/virtual_machines_test.go
```diff
@@ -425,7 +425,7 @@ var _ = Describe("Virtual Machines", func() {
 					}
 					return err
```
ormergi, line 427: How about we export `timeout+getVmFailCleanupWaitTime()` to variable with the name
totalTimeout?
its easier to debug when a function is called in separate line

RamLavi: not sure I got what you mean by export. can you elaborate?

tests/tests.go
RamLavi: totally agree

tests/tests.go
RamLavi: I think it's the standard golang way to `if`... at least that's what I saw from other reviews that commented against separating the line to 2 lines. @phoracek - is there a format we should stick to?

tests/virtual_machines_test.go
ormergi: I mean to create another variable :sweat_smile: 
`totalTime := timeout+getVmFailCleanupWaitTime()`

tests/tests.go
ormergi: I think we should return an error with a message "could not get ..."

tests/tests.go
```diff
@@ -209,6 +210,26 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
 	return nil
 }
 
+func getVmFailCleanupWaitTime() time.Duration {
+	managerDeployment, err := testClient.KubeClient.AppsV1().Deployments(managerNamespace).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
+	Expect(err).ToNot(HaveOccurred(), "Should succeed getting Deployment object")
```
ormergi, line 215: small nit, with this wording it will be aligned with all the rest of those messages we create
"Should successfully get Deployment object"

tests/tests.go
```diff
@@ -209,6 +210,26 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
...
+	Expect(len(managerDeployment.Spec.Template.Spec.Containers) != 0).To(BeTrue())
```
ormergi, line 216: Please add Expect message "Container count should be greater then zero / > 0 .."

tests/virtual_machines_test.go
RamLavi: ack

tests/tests.go
RamLavi: ack. I agree

tests/virtual_machines_test.go
```diff
@@ -403,6 +403,12 @@ var _ = Describe("Virtual Machines", func() {
 		})
 		//2633
 		Context("When we re-apply a failed VM yaml", func() {
+			totalTimeout := time.Duration(0)
+			BeforeAll(func() {
```
ormergi, line 407: You should use `BeforeEach` here I am almost sure `BeforeAll` invoked before test suit starts

tests/virtual_machines_test.go
```diff
@@ -403,6 +403,12 @@ var _ = Describe("Virtual Machines", func() {
...
+				vmFailCleanupWaitTime, err := getVmFailCleanupWaitTime()
+				Expect(err).ToNot(HaveOccurred())
```
ormergi, line 409: Could you add  Expect message please? :slightly_smiling_face:

tests/virtual_machines_test.go
RamLavi: If I do BeforeEach thenit will get run getVmFailCleanupWaitTime before each tests - which is unnecessary since it's not changing at all.
IIUC, BeforeAll will simply do it once in the beginning of the context, but only inside `Context("When we re-apply a failed VM yaml"` context.

tests/tests.go
ormergi: Most of  `if` clauses in kubemacpool code base wont initialize in the same line
it is possible but I think its confusing.

tests/virtual_machines_test.go
ormergi: You are right, [`BeforeAll` ](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/29c33895384710a7d2528c70b4059ee2663e50fc/tests/tests.go#L358) its not standard BTW

tests/virtual_machines_test.go
```diff
@@ -403,6 +403,12 @@ var _ = Describe("Virtual Machines", func() {
 		})
 		//2633
 		Context("When we re-apply a failed VM yaml", func() {
+			totalTimeout := time.Duration(0)
+			BeforeAll(func() {
+				vmFailCleanupWaitTime, err := getVmFailCleanupWaitTime()
+				Expect(err).ToNot(HaveOccurred(), "Should succeed getting wait-time argument")
```
ormergi, line 409: Can you change it like you did [here](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/152#discussion_r429628949)? :slightly_smiling_face:

tests/virtual_machines_test.go
ormergi: You are right, BTW [BeforeAll](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/29c33895384710a7d2528c70b4059ee2663e50fc/tests/tests.go#L358) is not standard

RamLavi: well the shoe fits I think :)

RamLavi: didn't even notice :) well somebody ought to add it to their API asap! :)

RamLavi:
>I think wait-time should be part of names.go, so we would use it by names.WAIT_TIME_ARG. What you think?

well In production and development environments - the wait-times are different, and since in names.go we only have consts parameters so I don't think it the right place. 
I could ask why it is passed as a input argument to the container instead of env. variable, but I don't object it to be either way.

ormergi:
> > I think wait-time should be part of names.go, so we would use it by names.WAIT_TIME_ARG. What you think?
> 
> well In production and development environments - the wait-times are different, and since in names.go we only have consts parameters so I don't think it the right place.
> I could ask why it is passed as a input argument to the container instead of env. variable, but I don't object it to be either way.

I meant to put in names.go the **name**(wait-time) of that argument we define in manifest not the value(600/10).

tests/tests.go
```diff
@@ -209,6 +210,26 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
...
+	managerDeployment, err := testClient.KubeClient.AppsV1().Deployments(managerNamespace).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
+	Expect(err).ToNot(HaveOccurred(), "Should successfully get Deployment object")
+	Expect(len(managerDeployment.Spec.Template.Spec.Containers) != 0).To(BeTrue(), "Container array size should be greater then zero")
+
+	for _, container := range managerDeployment.Spec.Template.Spec.Containers {
+		if container.Name == deploymentContainerName {
+			for _, arg := range container.Args {
+				waitTime := 0
+				if elementsParsed, err := fmt.Sscanf(arg, "--wait-time=%d", &waitTime); elementsParsed == 1 {
+					Expect(err).ToNot(HaveOccurred(), "Should succeed parsing wait-time arg")
```
ormergi, line 223: [Same](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/152#discussion_r429636994)

tests/tests.go
```diff
@@ -209,6 +210,26 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
...
+func getVmFailCleanupWaitTime() (time.Duration, error) {
+	managerDeployment, err := testClient.KubeClient.AppsV1().Deployments(managerNamespace).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
+	Expect(err).ToNot(HaveOccurred(), "Should successfully get Deployment object")
+	Expect(len(managerDeployment.Spec.Template.Spec.Containers) != 0).To(BeTrue(), "Container array size should be greater then zero")
+
+	for _, container := range managerDeployment.Spec.Template.Spec.Containers {
+		if container.Name == deploymentContainerName {
+			for _, arg := range container.Args {
+				waitTime := 0
+				if elementsParsed, err := fmt.Sscanf(arg, "--wait-time=%d", &waitTime); elementsParsed == 1 {
```
ormergi, line 222: I think this function shouldn't handle this error with `Expect` and just return it `return 0, error`
so the caller function will return an error like: 
"Should successfully get wait-time argument: <parse error from `Sscanf`>"

RamLavi: I agree

RamLavi:
>I meant to put in names.go the name(wait-time) of that argument we define in manifest not the value(600/10).

ok, well It's possible but I don't find a good reason to, since the actual string `wait-time` is used once in main.go, and in the deployment manifest (where it out of reach from the names.go I think)

tests/tests.go
```diff
@@ -209,6 +210,26 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
 	return nil
 }
 
+func getVmFailCleanupWaitTime() (time.Duration, error) {
+	managerDeployment, err := testClient.KubeClient.AppsV1().Deployments(managerNamespace).Get(names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
```
qinqon, line 214: Let's start to use controller-runtime client `testClient.VirtClient` (name is very unfortunate, it has nothing to do with virt) at the new places and near future we have replace all KubeClient with controller-runtime client so we benefict from caching and from some test facilities it has (like deleting everything creating with it at `AfterSuite`)

tests/tests.go
```diff
@@ -209,6 +210,26 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
...
+	Expect(err).ToNot(HaveOccurred(), "Should successfully get Deployment object")
```
qinqon, line 215: "Should successfully get manager's Deployment" better be specific to find the error at tests.

tests/tests.go
```diff
@@ -209,6 +210,26 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
...
+	Expect(len(managerDeployment.Spec.Template.Spec.Containers) != 0).To(BeTrue(), "Container array size should be greater then zero")
```
qinqon, line 216: Use gomega matchers as much as possible it make clear what was the value

```golang
Expect(managerDeployment.Spec.Template.Spec.Containers).ToNot(BeEmpty(), "Manager's deployment should contain containers")
```

tests/tests.go
qinqon, line 222: Let's reduce a little the identation level, if there is an error we don't care for elementsParsed
```golang
elementsParsed, err := fmt.Sscanf(arg, "--wait-time=%d", &waitTime)
Expect(err).ToNot(HaveOccurred(), "Should successfully parse wait-time arg")
if elementsParsed == 1 {
    return time.Duration(waitTime) * time.Second
}
```

tests/tests.go
```diff
@@ -209,6 +210,26 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
 	return nil
 }
 
+func getVmFailCleanupWaitTime() (time.Duration, error) {
```
qinqon, line 213: Don't return error, we just expect it.

tests/tests.go
```diff
@@ -209,6 +210,26 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
...
+				waitTime := 0
+				if elementsParsed, err := fmt.Sscanf(arg, "--wait-time=%d", &waitTime); elementsParsed == 1 {
+					Expect(err).ToNot(HaveOccurred(), "Should successfully parse wait-time arg")
+					return time.Duration(waitTime) * time.Second, err
+				}
+			}
+		}
+	}
+
+	return 0, fmt.Errorf("Failed to find wait-time argument in %s container inside %s deployment", deploymentContainerName, names.MANAGER_DEPLOYMENT)
```
qinqon, line 230: Let's just `Fail` if we are here and return 0
```golang
Fail(fmt.Sprintf("Failed to find wait-time argument in %s container inside %s deployment", deploymentContainerName, names.MANAGER_DEPLOYMENT))
return 0
```

tests/virtual_machines_test.go
```diff
@@ -403,6 +403,12 @@ var _ = Describe("Virtual Machines", func() {
 		})
 		//2633
 		Context("When we re-apply a failed VM yaml", func() {
+			totalTimeout := time.Duration(0)
+			BeforeAll(func() {
+				vmFailCleanupWaitTime, err := getVmFailCleanupWaitTime()
+				Expect(err).ToNot(HaveOccurred(), "Should successfully get wait-time argument")
```
qinqon, line 409: Let's just fail inside `getVm...` no need to return error.

tests/virtual_machines_test.go
```diff
@@ -403,6 +403,12 @@ var _ = Describe("Virtual Machines", func() {
...
+				totalTimeout = timeout + vmFailCleanupWaitTime
```
qinqon, line 410: Add a comment here explaining that we add some time window apart form vmFailCleanup to be sure that we hit it.

tests/tests.go
RamLavi: I ca't use Expect(err), we have more than 1 argument, and I want it to run through all the arguments and return only once it found the wait-time arg.
I can do something like 
```
			for _, arg := range container.Args {
				waitTime := 0
				elementsParsed, err := fmt.Sscanf(arg, "--wait-time=%d", &waitTime)
				if err != nil {
					continue
				}
				if elementsParsed == 1 {
					Expect(err).ToNot(HaveOccurred(), "Should successfully parse wait-time arg")
					return time.Duration(waitTime) * time.Second
				}
```
what do you think?

qinqon: Sscanf does not fail if does not find it if it fails is because the string is wrong or the like so we should stop the test.

RamLavi: ack

config/test/manager_image_patch.yaml
```diff
@@ -10,4 +10,4 @@ spec:
       - name: manager
         args:
           - "--v=debug"
-          - "--wait-time=10"
+          - "--wait-time=600"
```
qinqon, line 13: Do you kno if we can say to kustomize "Add debug flag" instead of rewritting args ? so we don't have to specify wait-time here.

tests/tests.go
```diff
@@ -209,6 +214,29 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
...
 }
 
+func getVmFailCleanupWaitTime() time.Duration {
+	managerDeployment := v1.Deployment{}
+	err := testClient.VirtClient.Get(context.TODO(), client.ObjectKey{Namespace: managerNamespace, Name: names.MANAGER_DEPLOYMENT}, &managerDeployment)
+	Expect(err).ToNot(HaveOccurred(), "Should successfully get manager's Deployment")
+	Expect(managerDeployment.Spec.Template.Spec.Containers).ToNot(BeEmpty(), "Manager's deployment should contain containers")
+
+	for _, container := range managerDeployment.Spec.Template.Spec.Containers {
+		if container.Name == deploymentContainerName {
```
qinqon, line 224: Let's remove a little of nesting
```golang
container.Name != deploymentContainerName {
    continue
}
```

config/test/manager_image_patch.yaml
RamLavi: I briefly google  it and didn't find anything of the kind..

tests/tests.go
```diff
@@ -228,6 +232,33 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
...
 }
 
+func getVmFailCleanupWaitTime() time.Duration {
+	managerDeployment := v1.Deployment{}
+	err := testClient.VirtClient.Get(context.TODO(), client.ObjectKey{Namespace: managerNamespace, Name: names.MANAGER_DEPLOYMENT}, &managerDeployment)
+	Expect(err).ToNot(HaveOccurred(), "Should successfully get manager's Deployment")
+	Expect(managerDeployment.Spec.Template.Spec.Containers).ToNot(BeEmpty(), "Manager's deployment should contain containers")
+
+	for _, container := range managerDeployment.Spec.Template.Spec.Containers {
+		if container.Name != deploymentContainerName {
```
ormergi, line 242: This `if` can be more straightforward by changing it to `==` and drop  `continue`
```
if container.Name != deploymentContainerName {
 .
.
.
}

File: tests/tests.go [hunk 22]
```diff
@@ -228,6 +232,33 @@ func initKubemacpoolParams(rangeStart, rangeEnd string) error {
...
+			continue
+		}
+		r := regexp.MustCompile(fmt.Sprintf("--%s=(\\d+)", names.WAIT_TIME_ARG))
+		for _, arg := range container.Args {
+			match := r.FindStringSubmatch(arg)
+			if match == nil {
```
ormergi, line 248: Same here, we can change this `if` to 
```
if match != nil { 
.
.
.
}
```

tests/virtual_machines_test.go
ormergi: Can we reword this comment so it will be clear why we need wait-time 
> It is necessary to extend the timeout given to VM creation,
> by the time we wait until a mac is released from vmWait-ConfigMap

config/test/manager_image_patch.yaml
ormergi: I think we should leave config/test  w/o `wait-time` argument and change the default value in main.go to be 60
so we could iterate faster on test suite.

config/test/kubemacpool.yaml
```diff
@@ -260,7 +260,7 @@ spec:
       containers:
       - args:
         - --v=debug
-        - --wait-time=10
+        - --wait-time=600
```
ormergi, line 263: I think we should leave config/test w/o wait-time argument and change the default value in main.go to be 60
so we could iterate faster on test suite.

tests/tests.go
RamLavi: yes I changed it to reduce indetation as @qinqon asked [here](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/152#discussion_r429901532)

tests/tests.go
RamLavi: [Same](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/152#discussion_r429901532)

config/test/kubemacpool.yaml
RamLavi: We need to match test & production behavior, to eliminate hidden issues like the one this commit has exposed.

tests/tests.go
ormergi: @qinqon @RamLavi 
 I think its not removing the nesting its just hiding it, eventually the stack depth will be the same and using if-continue in this case make the code less readable IMO

phoracek: I believe early exit (the one Quique suggested) is a good and recommended practice in go. The positive flow should be as close to the left side of the screen as possible. Ideally, you can just read the left-most block, ignoring all ifs, to get a gist on what happens in the code. That makes it more readable in my opinion.

ormergi: When looking for specific object in a collection the code in most cases, where there is no sophisticated logic, the code should be "positive" by looking for what we want `(if .. == ..`) and not what we dont (`if .. != ..`) IMO.
In this case there is no early exit when by using continue

qinqon:
@ormergi @RamLavi another option to reduce nesting is put code at a private function, it also helps with document it it.

tests/virtual_machines_test.go
```diff
@@ -403,6 +403,12 @@ var _ = Describe("Virtual Machines", func() {
 		})
 		//2633
 		Context("When we re-apply a failed VM yaml", func() {
+			totalTimeout := time.Duration(0)
+			BeforeAll(func() {
+				vmFailCleanupWaitTime := getVmFailCleanupWaitTime()
+				// we add some time window apart form vmFailCleanup() to be sure that we catch the vm mac release.
```
ormergi, line 409: Can we reword this comment so it will be clear why we need wait-time
It is necessary to extend the timeout given to VM creation,
by the time we wait until a mac is released from vmWait-ConfigMap

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -123,12 +124,25 @@ func (r *ReconcilePolicy) addFinalizerAndUpdate(virtualMachine *kubevirt.Virtual
...
+		// refresh vm instance
+		err := r.Get(context.TODO(), request.NamespacedName, virtualMachine)
+		if err != nil {
+			return errors.Wrap(err, "Failed to refresh vm instance")
+		}
+		virtualMachine.ObjectMeta.Finalizers = append(virtualMachine.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName)
+
+		err = r.Update(context.Background(), virtualMachine)
+		if err != nil {
+			return errors.Wrap(err, "failed to update the VM with the new finalizer")
```
qinqon, line 138: There is a bug here, when RetryOnConflict is used, you have to return the error exactly at it was produced so it can check it out.

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -150,14 +164,27 @@ func (r *ReconcilePolicy) removeFinalizerAndReleaseMac(virtualMachine *kubevirt.
...
+			return errors.Wrap(err, "Failed to refresh vm instance")
+		}
+
+		// remove our finalizer from the list and update it.
+		logger.V(1).Info("removing finalizers")
+		virtualMachine.ObjectMeta.Finalizers = helper.RemoveString(virtualMachine.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName)
+
+		err = r.Update(context.Background(), virtualMachine)
+		if err != nil {
+			return errors.Wrap(err, "failed to update vm instance to remove the finalizer")
```
qinqon, line 180: ditto

RamLavi:
/hold

should not be in the upcoming release.

RamLavi:
will pass test only after #159 is merged

tests/tests.go
```diff
@@ -4,13 +4,16 @@ import (
...
 	"math"
+	"regexp"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/rand"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	"github.com/k8snetworkplumbingwg/kubemacpool/build/_output/bin/go/src/strconv"
```
qinqon, line 15: Where is this coming from ? maybe the IDE you are using is doing this ?

tests/tests.go
RamLavi: will move it back and function to also document this.

pkg/controller/virtualmachine/virtualmachine_controller.go
RamLavi: will be fixed in another PR

pkg/controller/virtualmachine/virtualmachine_controller.go
RamLavi: will be fixed in another PR

tests/tests.go
RamLavi: yes. will fix

PR #159, opened 2020-05-26
adding retryOnConflict methods to protect from conflicts

RamLavi:

when creating and deleting multiple vms we may encounter conflict when updating the configmap and the vm instance.
In order to properly handle these conflict, we introduce the retryOnConflict method.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/virtualmachine/virtualmachine_controller.go
+++ b/pkg/controller/virtualmachine/virtualmachine_controller.go
@@ -20,8 +20,9 @@ import (
 	"context"
 
 	"github.com/pkg/errors"
-	kuberneteserror "k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/retry"
 	kubevirt "kubevirt.io/client-go/api/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -95,7 +96,7 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 	instance := &kubevirt.VirtualMachine{}
 	err = r.Get(context.TODO(), request.NamespacedName, instance)
 	if err != nil {
-		if kuberneteserror.IsNotFound(err) {
+		if apierrors.IsNotFound(err) {
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
@@ -106,7 +107,10 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 		logger.V(1).Info("The VM is not being deleted")
 		// The object is not being deleted, so if it does not have a finalizer,
 		// then lets add the finalizer and update the object.
-		err = r.addFinalizerAndUpdate(instance, &request)
+		if !helper.ContainsString(instance.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName) {
+			err = r.addFinalizerAndUpdate(&request)
+		}
+
 		return reconcile.Result{}, err
 	}
 	// The object is being deleted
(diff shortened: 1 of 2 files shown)


pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -150,14 +161,27 @@ func (r *ReconcilePolicy) removeFinalizerAndReleaseMac(virtualMachine *kubevirt.
...
+			return errors.Wrap(err, "Failed to refresh vm instance")
+		}
+
+		// remove our finalizer from the list and update it.
+		logger.V(1).Info("removing finalizers")
+		virtualMachine.ObjectMeta.Finalizers = helper.RemoveString(virtualMachine.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName)
+
+		err = r.Update(context.Background(), virtualMachine)
+		if err != nil {
+			return errors.Wrap(err, "failed to update vm instance to remove the finalizer")
```
qinqon, line 177: return error do not wrap it.

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -22,9 +22,11 @@ import (
 	"strings"
 	"time"
 
+	errorWrap "github.com/pkg/errors"
```
qinqon, line 25: Don't rename it like this.

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -22,9 +22,11 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
```
qinqon, line 27: call this `apierrors`

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -456,21 +486,21 @@ func (p *PoolManager) vmWaitingCleanupLook(waitTime int) {
...
-				log.V(1).Info("released mac address in waiting loop", "macAddress", macAddress)
+				logger.V(1).Info("released mac address in waiting loop", "macAddress", macAddress)
 			}
 		}
 
-		_, err = p.kubeClient.CoreV1().ConfigMaps(p.managerNamespace).Update(configMap)
-		if err != nil {
-			log.Error(err, "failed to update config map", "configMapName", names.WAITING_VMS_CONFIGMAP)
+		if configMapUpdateNeeded {
+			_, err = p.kubeClient.CoreV1().ConfigMaps(p.managerNamespace).Update(configMap)
```
qinqon, line 503: Don't we miss a retry here ?

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -123,12 +124,22 @@ func (r *ReconcilePolicy) addFinalizerAndUpdate(virtualMachine *kubevirt.Virtual
...
 
 	logger.V(1).Info("The VM does not have a finalizer")
-	virtualMachine.ObjectMeta.Finalizers = append(virtualMachine.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName)
 
-	err := r.Update(context.Background(), virtualMachine)
-	if err != nil {
-		logger.Error(err, "failed to update the VM with the new finalizer")
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		// refresh vm instance
+		err := r.Get(context.TODO(), request.NamespacedName, virtualMachine)
```
phoracek, line 130: In the original code we for some reason pass virtualMachine down to this function. With your change, you overwrite it with whatever is found on the cluster. Aren't we losing some value? If so, we need to change this. If we don't, we should pass only name and namespace to make this clearly unrelated.

pkg/pool-manager/virtualmachine_pool.go
RamLavi: In this routine it is not needed, since if it fails then it will succeed the next 3 seconds, no?

pkg/controller/virtualmachine/virtualmachine_controller.go
qinqon: This function is call if we can get a virtual machine at `Reconcile` so only virtualmachines are reaching this function.

pkg/pool-manager/virtualmachine_pool.go
qinqon: ack

pkg/controller/virtualmachine/virtualmachine_controller.go
RamLavi: we don't loose values, so I should pass only name & namespace like you suggest. 
But we do use the vm instance before the retryOnConflict to check if we already have a finalizer, but that could also be done outside of this function (and it will still be ok since it's name is `addFinalizerAndUpdate`.

PR #160, opened 2020-05-26
aligning wait-time of test to those of production

RamLavi:

wait-time is an argument passed by deployment to the kubemacpool manager pod.
When a vm is created but fails on a webhook that is not kubemacpool, then the mac should be retruned to the pool.
Since the kubemacpool controller cannot catch this cases, we use a goroutine thread that collects failed vms after the wait-time.

Currently we run on tests in a separate wait-time (10 seconds) than the one configured in production (10 min)
This can mask potential issues where the controller fails to do it's job from some reason, but the goroutine releases the mac and makes it look like all is good.

in order to run exactly as production, we want to change the wait-time to 10 mins - like in production.
This will cause tests to run +20min time (since the goroutine is chcked in 2 tests), so we change the test timeout accordingly

```release-note
NONE
```

### DIFF
--- a/config/test/kubemacpool.yaml
+++ b/config/test/kubemacpool.yaml
@@ -260,7 +260,7 @@ spec:
       containers:
       - args:
         - --v=debug
-        - --wait-time=10
+        - --wait-time=600
         command:
         - /manager
         env:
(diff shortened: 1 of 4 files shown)


RamLavi:
/hold

we may want to do it via kustomize

qinqon:
/hold

We can use jsonpatch with kustomize to add the an element to container Args so we don't duplicate wait-time [1]

[1] https://github.com/kubernetes-sigs/kustomize/issues/581#issuecomment-442444675

RamLavi:
/hold

we have an issue with the test manifest (debug is removed)

RamLavi:
/unhold

test manifest looks ok now

PR #161, opened 2020-05-27
moving finalizer addition to webhook

RamLavi:

Currently when a vm is created the finalizer is added in the controller, allowing for a small window for the vm to be deleted before creation reconcile puts the finalizer.
This can result with a mac to not be released properly from the kubemacpool internal cache.
The situation occurs especially in tests when create vm and deletion is very quick, and there are retryOnConflicts
    
In order to fix this, we insert the finalizer during the webhook request initiated by the vm creation.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/virtualmachine/virtualmachine_controller.go
+++ b/pkg/controller/virtualmachine/virtualmachine_controller.go
@@ -101,6 +101,7 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 	err = r.Get(context.TODO(), request.NamespacedName, instance)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
+			logger.V(1).Info("vm not found. aborting reconcile")
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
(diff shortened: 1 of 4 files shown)


RamLavi:
/hold

holding until further discussion with @alonSadan

qinqon:
@RamLavi can you rebase it ?

pkg/webhook/virtualmachine/virtualmachine.go
```diff
@@ -101,7 +102,18 @@ func (a *virtualMachineAnnotator) mutateCreateVirtualMachinesFn(ctx context.Cont
 	if err != nil {
 		// If the VM does not exist yet, allocate a new MAC address
 		if errors.IsNotFound(err) {
-			return a.poolManager.AllocateVirtualMachineMac(virtualMachine)
+			err = a.poolManager.AllocateVirtualMachineMac(virtualMachine)
+
+			if err != nil {
+				return err
```
qinqon, line 108: Wrap it

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -108,82 +108,62 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
...
 
 	if instance.ObjectMeta.DeletionTimestamp.IsZero() {
-		logger.V(1).Info("The VM is not being deleted")
-		// The object is not being deleted, so if it does not have a finalizer,
-		// then lets add the finalizer and update the object.
-		if !helper.ContainsString(instance.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName) {
-			err = r.addFinalizerAndUpdate(&request, logger)
+		logger.V(1).Info("vm create/update event")
+		// The object is not being deleted, so we can set the macs to allocated
+		err = r.poolManager.MarkVMAsReady(instance, logger)
```
qinqon, line 113: question: MarkVMAsReady is to be done after VM is created right ?

RamLavi: yes. **successfully** created...

alonSadan:
> /hold
> 
> holding until further discussion with @alonSadan

I like it!

alonSadan:
lgtm

pkg/webhook/virtualmachine/virtualmachine.go
```diff
@@ -101,7 +102,18 @@ func (a *virtualMachineAnnotator) mutateCreateVirtualMachinesFn(ctx context.Cont
 	if err != nil {
 		// If the VM does not exist yet, allocate a new MAC address
```
RamLavi, line 103: I'm not sure I understand why is this `if` there anymore. @phoracek ?

qinqon:
/hold cancel

All good and clear now

RamLavi:
/hold 

I want to make sure this doesn't insert a bug in test_id:2995 where the vm gets removed

RamLavi:
> /hold
> 
> I want to make sure this doesn't insert a bug in test_id:2995 where the vm gets removed from unknown reason

/unhold
suspicion removed.

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -433,6 +433,12 @@ func (p *PoolManager) MarkVMAsReady(vm *kubevirt.VirtualMachine, parentLogger lo
 			return nil
 		}
 
+		if len(vm.Spec.Template.Spec.Domain.Devices.Interfaces) == 0 {
```
qinqon, line 436: Why do we need this ?

pkg/webhook/virtualmachine/virtualmachine.go
```diff
@@ -92,20 +94,29 @@ func (a *virtualMachineAnnotator) Handle(ctx context.Context, req admission.Requ
...
-		if errors.IsNotFound(err) {
-			return a.poolManager.AllocateVirtualMachineMac(virtualMachine)
+		if apierrors.IsNotFound(err) {
+			err = a.poolManager.AllocateVirtualMachineMac(virtualMachine)
+			if err != nil {
+				return errors.Wrap(err, "Failed to allocate mac to the vm object")
+			}
+
+			if virtualMachine.ObjectMeta.DeletionTimestamp.IsZero() {
```
qinqon, line 110: Since this is at webhook we don't need this anymore.

pkg/pool-manager/virtualmachine_pool.go
RamLavi: this intended to remove a configmap update if there is no real change made to it.

pkg/webhook/virtualmachine/virtualmachine.go
RamLavi: I added this after reading this [discussion](https://github.com/knative/eventing/issues/643#issuecomment-443017409) on adding finalizers on webhooks.
The scenario is that a possibly the vm is already being deleted by **another** webhook who had rejected it. so basically if it is already being deleted then no use to add our finalizer.. 

but when I think of it, maybe also the allocation should be rejected..

pkg/pool-manager/virtualmachine_pool.go
qinqon: We are already iterating we will not iterate if this is empty we don't need this pre-check

PR #162, opened 2020-05-27
Add external CNAO support

ormergi:
This PR adds new flag  `CNAO_PROVIDER` in cluster/up.sh
script, in order to support cluster that deployed with CNAO externally.

This is part of running tier1 tests on D/S  effort PRs, 
it will also improve the flow of how we run functest on external clusters.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,3 +307,12 @@ make cluster-down
 # Run function test
 make functest
 ```
+
+In order to run functests on external cluster
+with CNAO installed:
+
+``` bash 
+export KUBEVIRT_PROVIDER=external
+export CNAO_PROVIDER=external
+make functest
+```
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -30,11 +30,18 @@ if [[ "$CNAO_PROVIDER" != external ]]; then
   ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/namespace.yaml
   ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/network-addons-config.crd.yaml
   ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/operator.yaml
+  # Deploy NetworkAddonsConfig
+  ./cluster/kubectl.sh apply -f ./hack/cna/cna-cr.yaml
+  # wait for cluster operator
+  ./cluster/kubectl.sh wait networkaddonsconfig cluster --for condition=Available --timeout=800s
+else # CNAO_PROVIDER == external
+  if [[ -z $(./cluster/kubectl.sh get pod -A -l app=kubemacpool) ]]; then
+    if [[ -z $(./cluster/kubectl.sh get networkaddonsconfig --no-headers -o custom-columns=:.metadata.name) ]]; then
+      ./cluster/kubectl.sh apply -f ./hack/cna/cna-cr.yaml
+    fi
+    ./cluster/kubectl.sh patch networkaddonsconfig cluster --type json -p='[{"op":"replace","path":"/spec/kubeMacPool","value":{}}]'
+  fi
 fi
-
-# wait for cluster operator
-./cluster/kubectl.sh wait networkaddonsconfig cluster --for condition=Available --timeout=800s
-
 
 # deploy kubevirt
 ./cluster/kubectl.sh apply -f https://github.com/kubevirt/kubevirt/releases/download/${KUBEVIRT_VERSION}/kubevirt-operator.yaml
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -25,11 +25,12 @@ if [[ "$KUBEVIRT_PROVIDER" != external ]]; then
     $(kubevirtci::path)/cluster-up/up.sh
 fi
 
-# Deploy CNA
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/namespace.yaml
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/network-addons-config.crd.yaml
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/operator.yaml
-./cluster/kubectl.sh create -f ./hack/cna/cna-cr.yaml
+if [[ "$CNAO_PROVIDER" != external ]]; then
+  # Deploy CNAO
+  ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/namespace.yaml
+  ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/network-addons-config.crd.yaml
+  ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/operator.yaml
+fi
 
 # wait for cluster operator
 ./cluster/kubectl.sh wait networkaddonsconfig cluster --for condition=Available --timeout=800s


phoracek:
This will not work if you need to change manifests. CNAO would have them outdated.

cluster/up.sh
```diff
@@ -25,11 +25,23 @@ if [[ "$KUBEVIRT_PROVIDER" != external ]]; then
...
+
+# Deploy NetworkAddonsConfig
+if [[ -z $(./cluster/kubectl.sh get pod -A -l app=kubemacpool) ]]; then
+  cna=$(./cluster/kubectl.sh get networkaddonsconfig -o custom-columns=:.metadata.name --no-headers)
+  if [[ -z $cna ]]; then
+    ./cluster/kubectl.sh apply -f ./hack/cna/cna-cr.yaml
+  fi
+
+  # Deploy kubeMacPool with CNAO
+  ./cluster/kubectl.sh patch networkaddonsconfig cluster --type json -p='[{"op":"replace","path":"/spec/kubeMacPool","value":{}}]'
```
RamLavi, line 43: why do we do this? isn't it always `{}` ?

```
  spec:
    kubeMacPool: {}
    linuxBridge: {}
    multus: {}
    nmstate: {}
    ovs: {}

```

ormergi: This patch adds the line `kubeMacPool: {}`

ormergi:
> This will not work if you need to change manifests. CNAO would have them outdated.

First I check if there is CNA object in the cluster, then I patch the CNA object

cluster/up.sh
```diff
@@ -25,15 +25,23 @@ if [[ "$KUBEVIRT_PROVIDER" != external ]]; then
...
-# Deploy CNA
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/namespace.yaml
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/network-addons-config.crd.yaml
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/operator.yaml
-./cluster/kubectl.sh create -f ./hack/cna/cna-cr.yaml
-
-# wait for cluster operator
-./cluster/kubectl.sh wait networkaddonsconfig cluster --for condition=Available --timeout=800s
-
+if [[ "$CNAO_PROVIDER" != external ]]; then
```
phoracek, line 28: If you have else, you should do the positive check first.

Where do you set the default value?

cluster/up.sh
```diff
@@ -25,15 +25,23 @@ if [[ "$KUBEVIRT_PROVIDER" != external ]]; then
...
+  # Deploy CNAO
+  ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/namespace.yaml
+  ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/network-addons-config.crd.yaml
+  ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/operator.yaml
+  # Deploy NetworkAddonsConfig
+  ./cluster/kubectl.sh apply -f ./hack/cna/cna-cr.yaml
+  # wait for cluster operator
+  ./cluster/kubectl.sh wait networkaddonsconfig cluster --for condition=Available --timeout=800s
+else # CNAO_PROVIDER == external
```
phoracek, line 37: You added a new attribute for users to use. Document it. And follow https://raw.githubusercontent.com/k8snetworkplumbingwg/kubemacpool/master/.github/PULL_REQUEST_TEMPLATE.md next time you open a PR

cluster/up.sh
qinqon: Let's use `CNAO_VERSION` instead of `CNAO_PROVIDER` here by default it has the CNAO_VERSION we use at project but can contain `external` value meaning we don't install it.

cluster/up.sh
```diff
@@ -25,15 +25,23 @@ if [[ "$KUBEVIRT_PROVIDER" != external ]]; then
...
-# wait for cluster operator
-./cluster/kubectl.sh wait networkaddonsconfig cluster --for condition=Available --timeout=800s
-
+if [[ "$CNAO_PROVIDER" != external ]]; then
+  # Deploy CNAO
+  ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/namespace.yaml
+  ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/network-addons-config.crd.yaml
+  ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/operator.yaml
+  # Deploy NetworkAddonsConfig
+  ./cluster/kubectl.sh apply -f ./hack/cna/cna-cr.yaml
```
qinqon, line 34: It's time that we rename cna-cr.yaml to cna-cr-multus.yaml and add a new one that is cna-cr-multus-kubemacpool.yaml

cluster/up.sh
```diff
@@ -25,15 +25,23 @@ if [[ "$KUBEVIRT_PROVIDER" != external ]]; then
...
+  # wait for cluster operator
+  ./cluster/kubectl.sh wait networkaddonsconfig cluster --for condition=Available --timeout=800s
+else # CNAO_PROVIDER == external
+  if [[ -z $(./cluster/kubectl.sh get pod -A -l app=kubemacpool) ]]; then
+    if [[ -z $(./cluster/kubectl.sh get networkaddonsconfig --no-headers -o custom-columns=:.metadata.name) ]]; then
+      ./cluster/kubectl.sh apply -f ./hack/cna/cna-cr.yaml
+    fi
+    ./cluster/kubectl.sh patch networkaddonsconfig cluster --type json -p='[{"op":"replace","path":"/spec/kubeMacPool","value":{}}]'
```
qinqon, line 42: It can be that the already deployed networkaddonsconfig has no multus configured (I doubt it though), so we have to patch for multus too just in case.

cluster/up.sh
```diff
@@ -25,15 +25,23 @@ if [[ "$KUBEVIRT_PROVIDER" != external ]]; then
...
+  ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/network-addons-config.crd.yaml
+  ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/operator.yaml
+  # Deploy NetworkAddonsConfig
+  ./cluster/kubectl.sh apply -f ./hack/cna/cna-cr.yaml
+  # wait for cluster operator
+  ./cluster/kubectl.sh wait networkaddonsconfig cluster --for condition=Available --timeout=800s
+else # CNAO_PROVIDER == external
+  if [[ -z $(./cluster/kubectl.sh get pod -A -l app=kubemacpool) ]]; then
+    if [[ -z $(./cluster/kubectl.sh get networkaddonsconfig --no-headers -o custom-columns=:.metadata.name) ]]; then
+      ./cluster/kubectl.sh apply -f ./hack/cna/cna-cr.yaml
```
qinqon, line 40: Let's just create new cna-cr-multus-kubemacpool.yaml and apply that.

cluster/up.sh
phoracek: This script is getting bloated and the logic is tailorfit for a single usecase.

This PR adds an option to reconfigure a deployed cluster that happens to have KMP installed. It seems more like a script you want to keep in your home directory for a single setup you happen to have. Do we want to reconfigure external clusters? It won't work if NAC is owned by HCO anyway.

CNAO should not be used in this script to deploy KMP for testing purposes. That completely reverts the dependency... This kind of integration script should be on CNAO "sync with KMP commit and deploy it" and then you can run KMP tests on such a cluster.

We don't want to entangle projects in both directions. And we should be cautious about scripts we include in the repo. People may use them.

README.md
```diff
@@ -307,3 +307,12 @@ make cluster-down
 # Run function test
 make functest
 ```
+
+In order to run functests on external cluster
+with CNAO installed:
```
phoracek, line 312: Add an inline example how to patch deployed CNAO to enable KMP if you really want to document this. User should be fully aware of it.

qinqon:
The idea about automatize kubemacpool tests at CNAO would be a job at CNAO that runs the tier1 tests from the commit id from kubemacpool so as @phoracek this is the opposite.

Also would be nice to be able to trigger those jobs at kubemacpool PRs or at least releases.

cluster/up.sh
ormergi: ack

cluster/up.sh
ormergi: ack

ormergi:
This is no longer needed and not the right place,
following PR https://github.com/kubevirt/cluster-network-addons-operator/pull/408

PR #166, opened 2020-05-31
adding reconcile request id to better track multiple thread logs

RamLavi:

Currently it's hard to follow along the multiple reconcile logs, that sometimes reconcile the same vm (for example, a create followed by a update request).
In order to make it easier to follow the logs, we added a pseudo random "unique" ID, to be attached to the start of all relevant logs in reconcile. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/virtualmachine/virtualmachine_controller.go
+++ b/pkg/controller/virtualmachine/virtualmachine_controller.go
@@ -18,7 +18,9 @@ package virtualmachine
 
 import (
 	"context"
+	"math/rand"
 
+	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -81,7 +83,9 @@ type ReconcilePolicy struct {
 
 // Reconcile reads that state of the cluster for a virtual machine object and makes changes based on the state
 func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	logger := log.WithName("Reconcile").WithValues("virtualMachineName", request.Name, "virtualMachineNamespace", request.Namespace)
+	//used for multi thread log separation
+	reconcileRequestId := rand.Int()
+	logger := log.WithName("Reconcile").WithValues("RequestId", reconcileRequestId, "virtualMachineName", request.Name, "virtualMachineNamespace", request.Namespace)
 	logger.V(1).Info("got a virtual machine event in the controller")
 
 	instanceOptedIn, err := r.poolManager.IsVmInstanceOptedIn(request.Namespace)
@@ -108,19 +112,19 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 		// The object is not being deleted, so if it does not have a finalizer,
 		// then lets add the finalizer and update the object.
 		if !helper.ContainsString(instance.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName) {
-			err = r.addFinalizerAndUpdate(&request)
+			err = r.addFinalizerAndUpdate(&request, logger)
 		}
 
 		return reconcile.Result{}, err
 	}
 	// The object is being deleted
 	logger.V(1).Info("The VM is being marked for deletion")
-	err = r.removeFinalizerAndReleaseMac(instance, &request)
+	err = r.removeFinalizerAndReleaseMac(instance, &request, logger)
 	return reconcile.Result{}, err
 }
 
-func (r *ReconcilePolicy) addFinalizerAndUpdate(request *reconcile.Request) error {
-	logger := log.WithName("addFinalizerAndUpdate").WithValues("virtualMachineName", request.Name, "virtualMachineNamespace", request.Namespace)
+func (r *ReconcilePolicy) addFinalizerAndUpdate(request *reconcile.Request, parentLogger logr.Logger) error {
+	logger := parentLogger.WithName("addFinalizerAndUpdate")
 	logger.V(1).Info("The VM does not have a finalizer")
 
 	virtualMachine := &kubevirt.VirtualMachine{}
@@ -143,18 +147,18 @@ func (r *ReconcilePolicy) addFinalizerAndUpdate(request *reconcile.Request) erro
 
 	logger.V(1).Info("Finalizer was added to the VM")
 
-	return r.poolManager.MarkVMAsReady(virtualMachine)
+	return r.poolManager.MarkVMAsReady(virtualMachine, parentLogger)
 }
 
-func (r *ReconcilePolicy) removeFinalizerAndReleaseMac(virtualMachine *kubevirt.VirtualMachine, request *reconcile.Request) error {
-	logger := log.WithName("removeFinalizerAndReleaseMac").WithValues("virtualMachineName", request.Name, "virtualMachineNamespace", request.Namespace)
+func (r *ReconcilePolicy) removeFinalizerAndReleaseMac(virtualMachine *kubevirt.VirtualMachine, request *reconcile.Request, parentLogger logr.Logger) error {
+	logger := parentLogger.WithName("removeFinalizerAndReleaseMac")
 
 	if !helper.ContainsString(virtualMachine.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName) {
 		return nil
 	}
 	// our finalizer is present, so lets handle our external dependency
 	logger.V(1).Info("The VM contains the finalizer. Releasing mac")
-	err := r.poolManager.ReleaseVirtualMachineMac(virtualMachine)
+	err := r.poolManager.ReleaseVirtualMachineMac(virtualMachine, parentLogger)
 	if err != nil {
 		return errors.Wrap(err, "failed to release mac")
 	}
--- a/pkg/pool-manager/virtualmachine_pool.go
+++ b/pkg/pool-manager/virtualmachine_pool.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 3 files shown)


pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -108,19 +111,19 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
...
 	logger.V(1).Info("The VM is being marked for deletion")
-	err = r.removeFinalizerAndReleaseMac(instance, &request)
+	err = r.removeFinalizerAndReleaseMac(instance, &request, reconcileRequestId)
 	return reconcile.Result{}, err
 }
 
-func (r *ReconcilePolicy) addFinalizerAndUpdate(request *reconcile.Request) error {
-	logger := log.WithName("addFinalizerAndUpdate").WithValues("virtualMachineName", request.Name, "virtualMachineNamespace", request.Namespace)
+func (r *ReconcilePolicy) addFinalizerAndUpdate(request *reconcile.Request, RequestId int) error {
+	logger := log.WithName("addFinalizerAndUpdate").WithValues("RequestId", RequestId, "virtualMachineName", request.Name, "virtualMachineNamespace", request.Namespace)
```
phoracek, line 126: Would it make sense to pass the whole logger and just extend it with the info about finalizer? It is still the same transaction. That way finalizer does not need to know about any transaction I'd being a thing. The same for other instances where you pass the id.

Passing the whole logger wuod also make logs clearer. You would see that this finalizer function has been called as a part of the reconciliation.

RamLavi: you mean that the same logger instance is passed from the reconcile function, and each time I concat the `log.WithName(<func name>).WithValues(bla bla..)` to it?

phoracek: I would pass the logger from Reconcile to addFinalizerAndUpdate as an argument. I'm not sure you can add another WithName. Maybe just WithValues("operation", "adding finalizer") or "function", "addFinalizerAndUpdate"

RamLavi: what do you think about adding it with `WithName`? that way in the log name I get the entire stack path: `VirtualMachine Controller.Reconcile.addFinalizerAndUpdate...`

qinqon:
@RamLavi can you paste here an example of how the log looks like ? I want to see how multiple `WithName` works

RamLavi:
> @RamLavi can you paste here an example of how the log looks like ? I want to see how multiple `WithName` works

```
2020-06-01T07:08:30.784Z	DEBUG	VirtualMachine Controller.Reconcile	got a virtual machine event in the controller	{"RequestId": 4751997750760398084, "virtualMachineName": "testvmbmjxhqm2lnsqd98q7mz8g49tqskvjvjt", "virtualMachineNamespace": "kubemacpool-test"}
2020-06-01T07:08:30.789Z	DEBUG	VirtualMachine Controller.Reconcile	The VM is not being deleted	{"RequestId": 4751997750760398084, "virtualMachineName": "testvmbmjxhqm2lnsqd98q7mz8g49tqskvjvjt", "virtualMachineNamespace": "kubemacpool-test"}
2020-06-01T07:08:30.789Z	DEBUG	VirtualMachine Controller.Reconcile.addFinalizerAndUpdate	The VM does not have a finalizer	{"RequestId": 4751997750760398084, "virtualMachineName": "testvmbmjxhqm2lnsqd98q7mz8g49tqskvjvjt", "virtualMachineNamespace": "kubemacpool-test"}
2020-06-01T07:08:30.791Z	DEBUG	controller-runtime.webhook.webhooks	received request	{"webhook": "/mutate-virtualmachines", "UID": "56c0b313-ae49-4679-812f-9855c802c3d8", "kind": "kubevirt.io/v1alpha3, Kind=VirtualMachine", "resource": {"group":"kubevirt.io","version":"v1alpha3","resource":"virtualmachines"}}

```

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -98,35 +99,32 @@ func (p *PoolManager) AllocateVirtualMachineMac(virtualMachine *kubevirt.Virtual
 	return nil
 }
 
-func (p *PoolManager) ReleaseVirtualMachineMac(vm *kubevirt.VirtualMachine) error {
+func (p *PoolManager) ReleaseVirtualMachineMac(vm *kubevirt.VirtualMachine, reconcileLog logr.Logger) error {
```
qinqon, line 102: Hey what do you think about removing log from here and creating the following at PoolManager
```golang
func (p *PoolManager) String() string {
...
}
```

This way you can do something like the following

```golang
fmt.Sprintf("KMP pool manager: %s", poolManager)
```

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -108,19 +112,19 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 		// The object is not being deleted, so if it does not have a finalizer,
 		// then lets add the finalizer and update the object.
 		if !helper.ContainsString(instance.ObjectMeta.Finalizers, pool_manager.RuntimeObjectFinalizerName) {
-			err = r.addFinalizerAndUpdate(&request)
+			err = r.addFinalizerAndUpdate(&request, logger)
```
qinqon, line 115: I am tented to sugget to have requestLogger as member of `PolicyReconcile` since we have only one goroutine running `Reconcile` and finalizer functions, but that's wrong if we do so and we change that we are broken, the alternative so we don't pass the logger to functions would be to add a struct with `PolicyReconcile`and logger and addFinalizer and removeFinalizers are functions from that struct.

pkg/pool-manager/virtualmachine_pool.go
phoracek: How does it help with the identification of single execution?

pkg/controller/virtualmachine/virtualmachine_controller.go
phoracek: The point of this PR is to add unique ID per each execution of Reconcile, so we can track those.

pkg/pool-manager/virtualmachine_pool.go
qinqon: Not much, just passing the logger to all functions as an specific case of `sillver bullet` feels uncomplete, maybe we need to pass some kind of requestContext to all the calls here so we can really trace things.

phoracek: I hate the appeal of dependency injection (countless arguments passed down the stack everywhere), but what is better context than single purpose obejct (logger) doing one thing, keeping all the information while allowing its users to make transparent logs without understanding that they are part of Reconcile loop

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -108,19 +112,19 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
...
 	// The object is being deleted
 	logger.V(1).Info("The VM is being marked for deletion")
-	err = r.removeFinalizerAndReleaseMac(instance, &request)
+	err = r.removeFinalizerAndReleaseMac(instance, &request, logger)
 	return reconcile.Result{}, err
 }
 
-func (r *ReconcilePolicy) addFinalizerAndUpdate(request *reconcile.Request) error {
-	logger := log.WithName("addFinalizerAndUpdate").WithValues("virtualMachineName", request.Name, "virtualMachineNamespace", request.Namespace)
+func (r *ReconcilePolicy) addFinalizerAndUpdate(request *reconcile.Request, reconcileLog logr.Logger) error {
```
phoracek, line 126: It does not matter to this method that the logger is from reconciler. For us it is just a logger. Maybe `parentLogger`?

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -143,18 +147,18 @@ func (r *ReconcilePolicy) addFinalizerAndUpdate(request *reconcile.Request) erro
 
 	logger.V(1).Info("Finalizer was added to the VM")
 
-	return r.poolManager.MarkVMAsReady(virtualMachine)
+	return r.poolManager.MarkVMAsReady(virtualMachine, reconcileLog)
 }
 
-func (r *ReconcilePolicy) removeFinalizerAndReleaseMac(virtualMachine *kubevirt.VirtualMachine, request *reconcile.Request) error {
-	logger := log.WithName("removeFinalizerAndReleaseMac").WithValues("virtualMachineName", request.Name, "virtualMachineNamespace", request.Namespace)
+func (r *ReconcilePolicy) removeFinalizerAndReleaseMac(virtualMachine *kubevirt.VirtualMachine, request *reconcile.Request, reconcileLog logr.Logger) error {
```
phoracek, line 153: ditto

pkg/pool-manager/virtualmachine_pool.go
phoracek, line 102: ditto

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -98,35 +99,32 @@ func (p *PoolManager) AllocateVirtualMachineMac(virtualMachine *kubevirt.Virtual
...
 	p.poolMutex.Lock()
 	defer p.poolMutex.Unlock()
-	log.V(1).Info("ReleaseVirtualMachineMac: data",
+	logger.V(1).Info("data",
 		"macmap", p.macPoolMap,
 		"podmap", p.podToMacPoolMap,
 		"currentMac", p.currentMac.String())
 
 	if len(vm.Spec.Template.Spec.Domain.Devices.Interfaces) == 0 {
-		log.V(1).Info("no interfaces found for virtual machine, skipping mac release",
```
phoracek, line 110: Here the passed logger starts to be a little weird. We assume that it contains VM info, so we are dropping it from here. But there is no guarantee somebody removes it from the parent logger. I'm puzzled. I don't think we can do anything about it. I don't like the fact that it semantically glues the whole chain together :grimacing: Any suggestion?

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -417,8 +415,8 @@ func (p *PoolManager) AddMacToWaitingConfig(allocations map[string]string) error
 }
 
 // Remove all the mac addresses from the waiting configmap this mean the vm was saved in the etcd and pass validations
-func (p *PoolManager) MarkVMAsReady(vm *kubevirt.VirtualMachine) error {
-	logger := log.WithName("MarkVMAsReady").WithValues("macPoolMap", p.macPoolMap, "virtualMachineNamespace", vm.Namespace, "virtualMachineName", vm.Name)
+func (p *PoolManager) MarkVMAsReady(vm *kubevirt.VirtualMachine, reconcileLog logr.Logger) error {
```
phoracek, line 418: ditto

pkg/pool-manager/virtualmachine_pool.go
qinqon: ack, fair enough.

pkg/pool-manager/virtualmachine_pool.go
RamLavi: well for better and for worse, the reason we did this is because there is information (the unique id, the vm info) that we hold in the first reconcile function, that we want to remember while reading the logs but don't necessarily want to carry with us to each future function. 
But I think that since the both (the vm info and the id) are consistent and always relevant through out the Reconcile function - then it's ok. what do you think?

qinqon:
@RamLavi @phoracek maybe you kill me but the golang `context` package is also designed to pass request related info across different APIs [1], maybe we can pass there whatever we need like request ID and the like, so we can log it and it's transparent.

[1] https://golang.org/pkg/context/#pkg-variables

phoracek:
@qinqon I don't want to carry context ID all the way down to helper functions. Helpers should not know or care about that value. They just want to log, so let's give them a logger. Sharing of context with this value sounds like unnecessary coupling.

pkg/pool-manager/virtualmachine_pool.go
phoracek: I hate that we depend on implicitly given context, but I don't have a better option to propose. So I guess it's ok

phoracek:
Seems to be already backported.

PR #168, opened 2020-06-03
Remove finalizer always

qinqon:
When a VM is created at an opt-in namespace and namespace is changed to
opt-out has to be possible to delete that VM. This PR try to delete finalizer and update
mac cache always independent of being at opt-in or not.

**Release note**:

```release-note
Remove finalizer is independent of opt-in option.
```

### DIFF
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -12,7 +12,8 @@ import (
 	. "github.com/onsi/gomega"
 
 	"k8s.io/apimachinery/pkg/api/errors"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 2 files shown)


tests/virtual_machines_test.go
```diff
@@ -124,19 +125,44 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
 				Expect(err).ToNot(HaveOccurred())
 
-				vm := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br", "")},
+				vm = CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br", "")},
 					[]kubevirtv1.Network{newNetwork("br")})
 
 				err = testClient.VirtClient.Create(context.TODO(), vm)
 				Expect(err).ToNot(HaveOccurred())
+			})
+			It("should has the vm with an automatically assign static MAC address within range", func() {
```
RamLavi, line 142: please rephrase
`should automatically assign the vm with static MAC address within range`

tests/virtual_machines_test.go
```diff
@@ -124,19 +125,44 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+				vmKey := types.NamespacedName{Namespace: vm.Namespace, Name: vm.Name}
+				err := testClient.VirtClient.Get(context.TODO(), vmKey, vm)
+				Expect(err).ToNot(HaveOccurred())
+
 				_, err = net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
 				Expect(err).ToNot(HaveOccurred())
 			})
+			Context("and we opt-out the namespace", func() {
```
RamLavi, line 150: please rephrase
`and then when we opt-out the namespace`

tests/virtual_machines_test.go
```diff
@@ -124,19 +125,44 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+				BeforeEach(func() {
+					err := cleanNamespaceLabels(TestNamespace)
+					Expect(err).ToNot(HaveOccurred(), "should be able to remove the namespace labels")
+				})
+				It("should able to be deleted", func() {
+					foregroundPolicy := metav1.DeletePropagationForeground
```
RamLavi, line 156: what does this do?

qinqon: Is suppose to wait until object is deleted including objects referencing it, I think is the case of the test passing without the fix, but I am investigating it.

PR #171, opened 2020-06-04
Ubi8 minimal

qinqon:
This PR change manager container base image to ubi8-minimal, this micis rhel8 containers and reduce container size.

**Release note**:

```release-note
Moved to ubi8-minimal base container.
```


Makefile
```diff
@@ -3,7 +3,6 @@ REGISTRY ?= quay.io
 REPO ?= kubevirt
```
phoracek, line 3: The commit message does not say why and I feel this should be a different PR. Than serious upgrade of base image.

PR #179, opened 2020-06-08
Use the controller-runtime manager leader election

qinqon:

**Release note**:

```release-note

```


pkg/manager/leaderelection.go
```diff
@@ -3,28 +3,86 @@ package manager
...
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 )
 
 func (k *KubeMacPoolManager) waitToStartLeading() error {
 	<-k.mgr.Elected()
-	return k.markPodAsLeader()
+	err := k.setLeadershipConditions(corev1.ConditionFalse)
```
RamLavi, line 17: did you mean to add this before the <-k.mgr.Elected() so that both pods will be set to false ( until the elected one is chosen)?

qinqon: Nope, If we set it before it just cover the startup case not the ledership change.

RamLavi: so, why set if to OFF only to set it to on again?

PR #182, opened 2020-06-09
Change leadership mechanism

RamLavi:
Currently our backup-active implementation make it so that only 1 kubemacpool pod is in ready state. This causes a [problem](https://bugzilla.redhat.com/show_bug.cgi?id=1844907) with the deployment status to be not ready. 
In order to fix this, we need to change the leadership election to occur using the controller-runtime api, and not externally as implemented now.

**Release note**:

```release-note
NONE
```


pkg/manager/manager.go
```diff
@@ -44,14 +43,13 @@ type KubeMacPoolManager struct {
...
-	podName                  string         // manager pod name
-	waitingTime              int            // Duration in second to lock a mac address before it was saved to etcd
-	leaderElection           *leaderelection.LeaderElector
+	restartChannel           chan struct{}   // Close the channel if we need to regenerate certs
+	kubevirtInstalledChannel chan struct{}   // This channel is close after we found kubevirt to reload the manager
+	stopSignalChannel        chan os.Signal  // stop channel signal
+	podNamespace             string          // manager pod namespace
+	podName                  string          // manager pod name
+	waitingTime              int             // Duration in second to lock a mac address before it was saved to etcd
+	mgr                      manager.Manager // Delegated controller-runtime manager
```
ormergi, line 53: Would  you change this name to `manger` or `runtimeManager`?

ormergi, line 53: Would you change this name to manger or runtimeManager?

pkg/manager/leaderelection.go
```diff
@@ -13,13 +14,18 @@ func (k *KubeMacPoolManager) waitToStartLeading() error {
 }
 
 func (k *KubeMacPoolManager) markPodAsLeader() error {
-	pod, err := k.clientset.CoreV1().Pods(k.podNamespace).Get(context.TODO(), k.podName, metav1.GetOptions{})
-	if err != nil {
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
ormergi, line 34: Can you please elaborate why do we need RetryOnConflict?

pkg/manager/manager.go
```diff
@@ -82,34 +83,31 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
...
-			log.Error(err, "failed to wait for leader election")
-			continue
-		}
-
-		healthProbeHost, ok := os.LookupEnv("HEALTH_PROBE_HOST")
-		if !ok {
-			log.Error(err, "Failed to load HEALTH_PROBE_HOST from environment variable")
-			os.Exit(1)
-		}
+	healthProbeHost, ok := os.LookupEnv("HEALTH_PROBE_HOST")
```
ormergi, line 86: Environment variables parsing  should be on main function then pass it to `NewKubeMacPoolManager`,
then if something go wrong it could fail earlier.

We should also do something with environment variables names (like we do with constant strings - exporting them  to a file like names.go

pkg/manager/manager.go
```diff
@@ -82,34 +83,31 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
...
+		log.Error(err, "Failed to load HEALTH_PROBE_HOST from environment variable")
+		os.Exit(1)
+	}
 
-		healthProbePort, ok := os.LookupEnv("HEALTH_PROBE_PORT")
-		if !ok {
-			log.Error(err, "Failed to load HEALTH_PROBE_PORT from environment variable")
-			os.Exit(1)
-		}
+	healthProbePort, ok := os.LookupEnv("HEALTH_PROBE_PORT")
```
ormergi, line 92: same

pkg/manager/manager.go
```diff
@@ -82,34 +83,31 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
...
+	if !ok {
+		log.Error(err, "Failed to load HEALTH_PROBE_PORT from environment variable")
+		os.Exit(1)
+	}
 
+	for k.continueToRunManager {
```
ormergi, line 98: why do we need `for`?

pkg/manager/manager.go
```diff
@@ -82,34 +83,31 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
...
 		log.Info("Setting up Manager")
 		mgr, err := manager.New(k.config, manager.Options{
-			MetricsBindAddress:     k.metricsAddr,
-			HealthProbeBindAddress: fmt.Sprintf("%s:%s", healthProbeHost, healthProbePort),
+			MetricsBindAddress:      k.metricsAddr,
```
ormergi, line 101: Let export lines 101-110 to another function `initRuntimeManager`

pkg/manager/leaderelection.go
```diff
@@ -13,13 +14,18 @@ func (k *KubeMacPoolManager) waitToStartLeading() error {
 }
 
 func (k *KubeMacPoolManager) markPodAsLeader() error {
```
ormergi, line 33: Can you change this function name to something like `labelLeaderElectedPod`?

pkg/manager/leaderelection.go
```diff
@@ -33,3 +50,33 @@ func (k *KubeMacPoolManager) markPodAsLeader() error {
...
+	}
+	for _, pod := range podList.Items {
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			podKey := types.NamespacedName{Namespace: k.podNamespace, Name: pod.Name}
+			err := k.mgr.GetClient().Get(context.TODO(), podKey, &pod)
+			if err != nil {
+				return err
+			}
+
+			pod.Status.Conditions = append(pod.Status.Conditions, corev1.PodCondition{Type: "kubemacpool.io/leader-ready", Status: status, LastProbeTime: metav1.Time{}})
```
ormergi, line 69: Let export `kubemacpool.io/leader-ready` to a variable or to a file like names.go

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -466,7 +467,11 @@ func (p *PoolManager) MarkVMAsReady(vm *kubevirt.VirtualMachine, parentLogger lo
 // mutating webhook but we didn't get the creation event in the controller loop
 // this mean the create was failed by some other mutating or validating webhook
 // so we release the virtual machine
-func (p *PoolManager) vmWaitingCleanupLook(waitTime int) {
+func (p *PoolManager) vmWaitingCleanupLook(waitTime int, mgr manager.Manager) {
+	if mgr != nil {
```
ormergi, line 471: Shouldn't we check if its the leader pod inside the mutex lock?

tests/tests.go
```diff
@@ -337,8 +316,8 @@ func changeManagerReplicas(numOfReplica int32) error {
 		if managerDeployment.Status.Replicas != numOfReplica {
 			return false
 		}
-		//due to readiness probe only 1 (the leader) pod will be ready (if any)
-		if float64(managerDeployment.Status.ReadyReplicas) != math.Min(float64(1), float64(numOfReplica)) {
+
+		if managerDeployment.Status.ReadyReplicas != numOfReplica {
```
ormergi, line 320: Isnt it enough to check `statusReadyReplicas` only?

RamLavi:
/hold

new elected does not acquire old cache

pkg/manager/leaderelection.go
```diff
@@ -2,86 +2,81 @@ package manager
...
 }
 
-func (k *KubeMacPoolManager) markPodAsLeader() error {
-	pod, err := k.clientset.CoreV1().Pods(k.podNamespace).Get(context.TODO(), k.podName, metav1.GetOptions{})
+// By setting this status to true in all pods, we declare the kubemacpool as ready and allow the webhooks to start running.
+func (k *KubeMacPoolManager) setLeadershipConditions(status corev1.ConditionStatus) error {
+	podList := corev1.PodList{}
+	err := k.mgr.GetClient().List(context.TODO(), &podList, &client.ListOptions{Namespace: k.podNamespace})
 	if err != nil {
 		return err
```
qinqon, line 59: Wrap this error with specific string

pkg/manager/leaderelection.go
```diff
@@ -2,86 +2,81 @@ package manager
...
 	}
+	for _, pod := range podList.Items {
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			podKey := types.NamespacedName{Namespace: k.podNamespace, Name: pod.Name}
+			err := k.mgr.GetClient().Get(context.TODO(), podKey, &pod)
+			if err != nil {
+				return err
```
qinqon, line 66: We can wrap this error is not going to return conflict

pkg/manager/leaderelection.go
```diff
@@ -2,86 +2,81 @@ package manager
...
 
-	log.Info("marked this manager as leader for webhook", "podName", k.podName)
+			err = k.mgr.GetClient().Status().Update(context.TODO(), &pod)
+			if err != nil {
+				return err
+			}
+			return nil
+		})
+		if err != nil {
+			return err
```
qinqon, line 78: wrap the error

pkg/manager/leaderelection.go
qinqon: If k8s changes something like annotation at pods (there is a lot of stuff going on at annotations and labels that we don't control) this is going to conflict, better be safe here, it does not introduce any cost.

pkg/manager/manager.go
qinqon: @RamLavi we don't need this env variables anymore since we are not using controller-runtime healtz mechanism, so we can remove the HEALT_PROBE_x variable retrieval and also manifests and remove `HealthProbeBindAddress` from manager, but this can be done at master, I am going to put a PR in place.

PR removing those variables, so we reduce the size of this PR, https://github.com/k8snetworkplumbingwg/kubemacpool/pull/184

pkg/pool-manager/pool.go
```diff
@@ -93,7 +94,7 @@ func NewPoolManager(kubeClient kubernetes.Interface, rangeStart, rangeEnd net.Ha
 	}
 
 	if kubevirtExist {
-		go poolManger.vmWaitingCleanupLook(waitTime)
+		go poolManger.vmWaitingCleanupLook(waitTime, mgr)
```
qinqon, line 97: store mgr as field of the PoolManager so we don't have to pass it to function.

tests/tests.go
```diff
@@ -351,11 +330,17 @@ func changeManagerReplicas(numOfReplica int32) error {
...
-				return false
+			for _, condition := range podObject.Status.Conditions {
+				if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
+					numberOfReadyPods += 1
+				}
 			}
 		}
+		if numberOfReadyPods < numOfReplica {
+			return false
+		}
```
qinqon, line 343: As @ormergi point out, just checking `managerDeployment.Status.ReadyReplicas != numOfReplica` is enough since it's more or less the same as this check.

pkg/manager/manager.go
RamLavi: notice [waitForSignal go routine](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/7bffdd3ca7959767a3e10a81d6bf272d98a8fca4/pkg/manager/manager.go#L177). it uses channels to signal stuff like "kubevir tInstalled".

Following the kubevirt example, Should kubevirt not be present during the start of kmp, then all is paused.
but when kubevirt is installed then kubevirtInstalledChannel willbe closed, which will pop up a signal in `waitForSignal` which will cause us to return to the start of the loop and rerun the kmp without a total pod restart..

pkg/pool-manager/virtualmachine_pool.go
RamLavi: I don't think we want to hold the mutex locked until the pod is elected. if it's our intention, it should be in a separate function that will document it properly.

pkg/manager/leaderelection.go
RamLavi: changed to `AddLeaderLabelToElectedPod`. is that ok?

pkg/manager/leaderelection.go
```diff
@@ -2,86 +2,89 @@ package manager
...
 
 	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
+	poolmanager "github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager"
 )
 
-func (k *KubeMacPoolManager) newLeaderElection(config *rest.Config, scheme *runtime.Scheme) error {
-	recorderProvider, err := NewProvider(config, scheme, log.WithName("events"))
+func (k *KubeMacPoolManager) waitToStartLeading(poolManger *poolmanager.PoolManager) error {
+	<-k.runtimeManager.Elected()
+	// If we reach here then we are in the elected pod.
```
qinqon, line 19: I think you have to do poolManager.Start here, so pool manager does not need the mgr.

pkg/pool-manager/pool.go
RamLavi: oh now I see what you mean. ack.

RamLavi:
/unhold

fixed issues and also added a test for it in other PR #183

pkg/manager/leaderelection.go
ormergi: perfect

PR #183, opened 2020-06-10
add collision detection test in leader change context.

RamLavi:

Currently, there is no test checking that a vm that was created in kubemacpool pod leader is successfully migrated to a new leader should leadership change occurs.
This commit adds such a test.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -130,7 +130,6 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 				vm *kubevirtv1.VirtualMachine
 			)
 			BeforeEach(func() {
-				By("Restart kubemacpool to reset cache and mac range")
 				err := initKubemacpoolParams(rangeStart, rangeEnd)
 				Expect(err).ToNot(HaveOccurred(), "should success restarting kubemacpool to reset mac range and cache")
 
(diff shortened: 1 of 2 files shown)


tests/virtual_machines_test.go
```diff
@@ -516,26 +516,54 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+				By("checking that the new vm's allocated mac is valid")
+				_, err = net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
+				Expect(err).ToNot(HaveOccurred(), "should success parsing the vm's mac")
+
+				ChangeManagerLeadership()
+
+				By("checking that a new VM with same mac as old vm cannot be created")
+				err = testClient.VirtClient.Create(context.TODO(), anotherVm)
+				Expect(err).To(HaveOccurred(), "should fail creating vm with duplicate mac")
+				Expect(strings.Contains(err.Error(), "failed to allocate requested mac address")).To(Equal(true))
```
ormergi, line 566: We should avoid checks like these please change/remove it.
Depending on specific error string is bad practice, we had  a test that checked for specific error string, that error string eventually changed and that test became false positive.
The time that it took to debug and and fix it could have been spent on something more important :wink:

qinqon:
Let's move this tests here https://github.com/k8snetworkplumbingwg/kubemacpool/blob/master/tests/virtual_machines_test.go#L153-L175

So they just change ownership and create VMs.

RamLavi:
> Let's move this tests here https://github.com/k8snetworkplumbingwg/kubemacpool/blob/master/tests/virtual_machines_test.go#L153-L175
> 
> So they just change ownership and create VMs.

I'm not sure I understand. do you mean to merge the 2 tests together?
If so then I think it should be in a separate "leadership change" context.

tests/virtual_machines_test.go
```diff
@@ -515,27 +515,46 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 		})
 
 		Context("When the leader is changed", func() {
```
qinqon, line 517: "When a VM is created at opt-in namespace and leadership is changed"

tests/virtual_machines_test.go
```diff
@@ -515,27 +515,46 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
-			It("should be able to create a new virtual machine", func() {
+			var (
+				vm *kubevirtv1.VirtualMachine
+			)
+			BeforeEach(func() {
+				By("Restart kubemacpool to reset cache and mac range")
```
qinqon, line 522: Can we put this `By` iniside `initKubemacpoolParams` ? if it's not already there.

tests/virtual_machines_test.go
```diff
@@ -515,27 +515,46 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+
+				By("checking that a new VM can be created on new leader")
+				err := testClient.VirtClient.Create(context.TODO(), anotherVm)
+				Expect(err).ToNot(HaveOccurred(), "should success creating the vm")
 				_, err = net.ParseMAC(anotherVm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
-				Expect(err).ToNot(HaveOccurred())
+				Expect(err).ToNot(HaveOccurred(), "should success parsing the vm's mac")
+			})
+			It("should successfully prevent collisions with virtual machine created on previous leader", func() {
+				By("Restart kubemacpool to reset cache and mac range")
```
qinqon, line 550: not needed

tests/virtual_machines_test.go
RamLavi: I think all contexts except the 1 (the opted out) is opt in by default.
Should I change it anyway?

qinqon: Nah all good.

PR #188, opened 2020-06-11
fix race between webhook and controller

RamLavi:

Currently, it's possible that the a webhook 1 will add a mac and wrongly re-insert it to the cache on controller reconcile if it was bypassed by a new webhook 2 deleting the macs. This would cause the belated controller to re-add the mac.
To fix it we add a check that the mac is not already deleted in the controller.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/pool-manager/pool_test.go
+++ b/pkg/pool-manager/pool_test.go
@@ -17,14 +17,17 @@ limitations under the License.
 package pool_manager
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
-	. "github.com/onsi/ginkgo"
-	"github.com/onsi/ginkgo/extensions/table"
-	. "github.com/onsi/gomega"
 	"net"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 
 	multus "github.com/intel/multus-cni/types"
+	"github.com/onsi/ginkgo/extensions/table"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


pkg/pool-manager/pool_test.go
```diff
@@ -374,6 +377,46 @@ var _ = Describe("Pool", func() {
...
+				fmt.Printf("configMap Data %v\n", configMap.Data)
+				_, exist := configMap.Data[macAddressInConfigMapFormat]
+				Expect(exist).To(Equal(true), "should have an entry of the mac in the configmap")
+
+				By("check allocated pool is populated and set to AllocationStatusWaitingForPod status")
+				Expect(len(poolManager.macPoolMap)).To(Equal(1), "macPoolMap should hold the mac address waiting for approval")
+				Expect(poolManager.macPoolMap[allocatedMac]).To(Equal(AllocationStatusWaitingForPod), "macPoolMap's mac's status should be set to AllocationStatusWaitingForPod status")
+
+				By("mark the vm as allocated")
+				err = poolManager.MarkVMAsReady(newVM, log.WithName("fake-Reconcile"))
```
qinqon, line 408: Before this we have to call MarkVMAsReady when `macPoolMap` does NOT contain the mac address so we are sure that we don't add entries.

pkg/pool-manager/pool_test.go
```diff
@@ -375,6 +378,66 @@ var _ = Describe("Pool", func() {
...
```
qinqon, line 415: Let's move `MarkVMAsReady` to a nested `Context` with BeforeEach so we can test more stuff at `It`

pkg/pool-manager/pool_test.go
```diff
@@ -375,6 +378,66 @@ var _ = Describe("Pool", func() {
...
+				newVM = masqueradeVM.DeepCopy()
+				newVM.Name = "newVM"
+
+				By("Create a VM")
+				err := poolManager.AllocateVirtualMachineMac(newVM)
+				Expect(err).ToNot(HaveOccurred(), "should successfully  allocated macs")
+
+				allocatedMac = newVM.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress
+				By("check mac allocated as expected")
+				Expect(allocatedMac).To(Equal("02:00:00:00:00:01"), "should successfully allocate the first mac in the range")
```
qinqon, line 398: Move to `It`

pkg/pool-manager/pool_test.go
```diff
@@ -375,6 +378,66 @@ var _ = Describe("Pool", func() {
...
+
+				By("check configmap updated with mac allocated")
+				configMap, err := poolManager.kubeClient.CoreV1().ConfigMaps(poolManager.managerNamespace).Get(context.TODO(), names.WAITING_VMS_CONFIGMAP, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred(), "should successfully get configmap")
+
+				macAddressInConfigMapFormat := strings.Replace(allocatedMac, ":", "-", 5)
+				Expect(len(configMap.Data)).To(Equal(1), "configmap should hold the mac address waiting for approval")
+				fmt.Printf("configMap Data %v\n", configMap.Data)
+				_, exist := configMap.Data[macAddressInConfigMapFormat]
+				Expect(exist).To(Equal(true), "should have an entry of the mac in the configmap")
```
qinqon, line 408: Move this to `It`

pkg/pool-manager/pool_test.go
```diff
@@ -375,6 +378,66 @@ var _ = Describe("Pool", func() {
...
+
+				By("check allocated pool is populated and set to AllocationStatusWaitingForPod status")
+				Expect(len(poolManager.macPoolMap)).To(Equal(1), "macPoolMap should hold the mac address waiting for approval")
```
qinqon, line 411: In general use gomega `HasLen` or `BeEmpty` instead of using `len`

pkg/pool-manager/pool_test.go
RamLavi: do you mean in separate `It`?

qinqon: We can check this at an `It("should successfully allocate the first mac in the range")` after this `BeforeEach`

PR #189, opened 2020-06-11
add logs to document failing scenarios in kubemacpool startup.

RamLavi:

Currently in the main Run() method, we don't log error in case something fails.
We want to exit the Run() method should a irreconcilable error occur, and a proper log to be added in that case.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -17,7 +17,6 @@ package manager
 
 import (
 	"context"
-	"fmt"
 	"net"
 	"os"
 	"os/signal"
@@ -76,12 +75,12 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 	log.Info("Setting up client for manager")
 	k.config, err = config.GetConfig()
 	if err != nil {
-		return fmt.Errorf("unable to set up client config error %v", err)
+		return errors.Wrap(err, "unable to set up client config")
 	}
 
 	k.clientset, err = kubernetes.NewForConfig(k.config)
 	if err != nil {
-		return fmt.Errorf("unable to create a kubernetes client error %v", err)
+		return errors.Wrap(err, "unable to create a kubernetes client")
 	}
 
 	for k.continueToRunManager {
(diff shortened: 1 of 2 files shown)


alonSadan:
lgtm

phoracek:
```
cmd/manager/main.go:25:2: imported and not used: "github.com/pkg/errors"
```

pkg/manager/manager.go
```diff
@@ -92,13 +91,13 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 
 		err = kubevirt_api.AddToScheme(k.runtimeManager.GetScheme())
 		if err != nil {
-			errors.Wrap(err, "unable to register kubevirt scheme")
+			return errors.Wrap(err, "unable to register kubevirt scheme")
```
qinqon, line 94: This changes behaviour and it will end `Run` are we sure we want that ?

pkg/manager/manager.go
```diff
@@ -92,13 +91,13 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
...
 		}
 
 		isKubevirtInstalled := checkForKubevirt(k.clientset)
 		poolManager, err := poolmanager.NewPoolManager(k.clientset, rangeStart, rangeEnd, k.podNamespace, isKubevirtInstalled, k.waitingTime)
 		if err != nil {
-			errors.Wrap(err, "unable to create pool manager")
+			return errors.Wrap(err, "unable to create pool manager")
```
qinqon, line 100: ditto

pkg/manager/manager.go
```diff
@@ -112,18 +111,18 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 		log.Info("Setting up controllers")
 		err = controller.AddToManager(k.runtimeManager, poolManager)
 		if err != nil {
-			errors.Wrap(err, "unable to register controllers to the manager")
+			return errors.Wrap(err, "unable to register controllers to the manager")
```
qinqon, line 114: ditto

pkg/manager/manager.go
```diff
@@ -112,18 +111,18 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
...
 		}
 
 		log.Info("Setting up webhooks")
 		err = webhook.AddToManager(k.runtimeManager, poolManager)
 		if err != nil {
-			errors.Wrap(err, "unable to register webhooks to the manager")
+			return errors.Wrap(err, "unable to register webhooks to the manager")
```
qinqon, line 120: ditto

pkg/manager/manager.go
```diff
@@ -112,18 +111,18 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
...
 		}
 
 		err = k.runtimeManager.Start(k.restartChannel)
 		if err != nil {
-			errors.Wrap(err, "unable to run the manager")
+			log.Error(err, "unable to run the manager")
```
qinqon, line 125: Why don't we want to return here ?

pkg/manager/manager.go
RamLavi: it does change behavior, but I think the former behavior was a bug. 
I mean, how can kubemacpool work properly if kubevirt scheme is not added? or if pool is not allocated correctly?

still it's a change so let's consult with @phoracek

pkg/manager/manager.go
RamLavi: cases such as kubevirt was just applied and we want to restart the manager without restarting the pod

PR #190, opened 2020-06-11
remove old leader labels  after leader election has been won.

RamLavi:

In a case where an election is lost and the pod does require restart - we need to make sure that the we have only 1 leader pod label.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/manager/leaderelection.go
+++ b/pkg/manager/leaderelection.go
@@ -2,6 +2,7 @@ package manager
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -9,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 
 	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 	poolmanager "github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager"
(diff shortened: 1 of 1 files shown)


RamLavi:
/hold

test failing due to issue fixed in #188

pkg/manager/leaderelection.go
```diff
@@ -35,7 +40,23 @@ func (k *KubeMacPoolManager) waitToStartLeading(poolManger *poolmanager.PoolMana
...
+		return errors.Wrap(err, "failed cleaning leader label from pods")
+	}
+
+	err = k.AddLeaderLabelToElectedPod()
+	if err != nil {
+		return errors.Wrap(err, "failed adding leader label to elected pod")
+	}
+
+	return nil
+}
```
qinqon, line 55: Since we are iterating all the kubemacpool pods we can remove/add label at the same loop, so we can move the code from `removeLeaderLabelAllKubemacpoolPods` here and remove the leader label at all pods except  the one from `POD_NAME`.

pkg/manager/leaderelection.go
```diff
@@ -52,7 +73,41 @@ func (k *KubeMacPoolManager) AddLeaderLabelToElectedPod() error {
 		return errors.Wrap(err, "failed to update leader label to elected kubemacpool manager pod")
 	}
 
-	log.Info("marked this manager as leader for webhook", "podName", k.podName)
+	logger.V(1).Info("marked this manager as leader for webhook", "podName", k.podName)
+	return nil
+}
+
+func (k *KubeMacPoolManager) removeLeaderLabelFromAllKubemacpoolPods() error {
```
qinqon, line 80: If we continue with this approach, we can filter by label instead of namespace so we return the pod with leader label and remove it.

RamLavi: yes, but technically could be super wierd snecario where we have more than 1... I think doing it in 1 loop is better like you said in above comment

pkg/manager/leaderelection.go
RamLavi: yes totally

PR #203, opened 2020-06-22
Remove tests which involve changing MAC range.

yossisegev:
Changing the MAC range is not supported by the KubeMacPool anymore,
and the range is determined by the cluster-network-addons-operator.
Test cases which include changing the MAC range should now be removed.


phoracek:
/hold

These tests are not changing the MAC range. They are testing different ranges where for every single one of them we redeploy KMP with a different range.

We cannot run this test on production environment where KMP is predeployed and not owned by us, but it is perfectly fine on tier1 where we own the lifecycle.

yossisegev:
> These tests are not changing the MAC range. They are testing different ranges where for every single one of them we redeploy KMP with a different range.

Well, if that's what actually happens, then the automation is not implementing the test as it is intended. These test cases in Polarion explicitly include changing the range.

> We cannot run this test on production environment where KMP is predeployed and not owned by us, but it is perfectly fine on tier1 where we own the lifecycle.

Then in that case - we might leave these tests as active in Polarion, and label them as tier1-only.

phoracek:
> > These tests are not changing the MAC range. They are testing different ranges where for every single one of them we redeploy KMP with a different range.
> 
> Well, if that's what actually happens, then the automation is not implementing the test as it is intended. These test cases in Polarion explicitly include changing the range.

Polarion was mapping to existing U/S tests, so the dependency is reversed.

In this U/S repo, user is owner of the range and can change it by redeploying of KMP. On D/S, the range is not controlled by admin, nor can we redeploy the whole product just to change the range within a single test.

> 
> > We cannot run this test on production environment where KMP is predeployed and not owned by us, but it is perfectly fine on tier1 where we own the lifecycle.
> 
> Then in that case - we might leave these tests as active in Polarion, and label them as tier1-only.

But you still can't run them against D/S since there you are not able to control the range. If "tier-1 only" means "runs only on U/S", then it's fine.

yossisegev:
> > > We cannot run this test on production environment where KMP is predeployed and not owned by us, but it is perfectly fine on tier1 where we own the lifecycle.
> > 
> > 
> > Then in that case - we might leave these tests as active in Polarion, and label them as tier1-only.
> 
> But you still can't run them against D/S since there you are not able to control the range. If "tier-1 only" means "runs only on U/S", then it's fine.

Yes, that what it means. I wasn't very clear, sorry.
So I labeled these 3 tests in Polarion as "U/S only", and will abandon this PR.

RamLavi:
> So I labeled these 3 tests in Polarion as "U/S only", and will abandon this PR.

I will try to think of a way for the tier1 tests to work also on d/s.

PR #204, opened 2020-06-22
drop namespace annotations

phoracek:

Fixing BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1847594

We should not sneak into the restricted space.

**Release note**:

```release-note
Drop runtime level label from Namespace
```

### DIFF
--- a/config/default/manager/manager.yaml
+++ b/config/default/manager/manager.yaml
@@ -3,9 +3,9 @@ kind: Namespace
 metadata:
   labels:
     control-plane: mac-controller-manager
-    controller-tools.k8s.io: "1.0"
-    runlevel: "0"
-    openshift.io/run-level: "0"
+    # Make sure that KubeMacPool pods are never handled by the webhook to
+    # prevent dead-lock.
+    mutatepods.kubemacpool.io: ignore
   name: system
 ---
 apiVersion: v1
--- a/config/release/kubemacpool.yaml
+++ b/config/release/kubemacpool.yaml
@@ -3,9 +3,7 @@ kind: Namespace
 metadata:
   labels:
     control-plane: mac-controller-manager
-    controller-tools.k8s.io: "1.0"
-    openshift.io/run-level: "0"
-    runlevel: "0"
+    mutatepods.kubemacpool.io: ignore
   name: kubemacpool-system
 ---
 apiVersion: admissionregistration.k8s.io/v1beta1
--- a/config/test/kubemacpool.yaml
+++ b/config/test/kubemacpool.yaml
@@ -3,9 +3,7 @@ kind: Namespace
 metadata:
   labels:
     control-plane: mac-controller-manager
-    controller-tools.k8s.io: "1.0"
-    openshift.io/run-level: "0"
-    runlevel: "0"
+    mutatepods.kubemacpool.io: ignore
   name: kubemacpool-system
 ---
 apiVersion: admissionregistration.k8s.io/v1beta1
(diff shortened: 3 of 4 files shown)


RamLavi:
@phoracek 
right now it's ok to insert this because there is no way KMP would try to allocate its own pods  since their namespace in not opted in. 
But we need to come up with another way to do that should we remove the opt-in feature in the future.

phoracek:
We can rollback to what we used to do before this label. We had our own that we placed on our namepsace and used it for the anti-selector.

config/default/manager/manager.yaml
```diff
@@ -3,9 +3,8 @@ kind: Namespace
 metadata:
   labels:
     control-plane: mac-controller-manager
-    controller-tools.k8s.io: "1.0"
-    runlevel: "0"
-    openshift.io/run-level: "0"
+    mutatepods.kubemacpool.io: ignore
```
RamLavi, line 6: is it necessary to add a default label value?
if not, I would prefer not to add it.
If we do, we need  to change the documentation accordingly to use --overwrite :
```
hades01:cluster-network-addons-operator $ kubectl get namespace ram -oyaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"Namespace","metadata":{"annotations":{},"labels":{"mutatevirtualmachines.kubemacpool.io":"allocate"},"name":"ram"}}
  creationTimestamp: "2020-06-22T12:30:50Z"
  labels:
    mutatevirtualmachines.kubemacpool.io: allocate
  name: ram
  resourceVersion: "6583"
  selfLink: /api/v1/namespaces/ram
  uid: 7111f5df-2b14-487f-ab23-758ffa9fde21
spec:
  finalizers:
  - kubernetes
status:
  phase: Active
hades01:cluster-network-addons-operator $ kubectl label namespace example-namespace  mutatevirtualmachines.kubemacpool.io=bla
Error from server (NotFound): namespaces "example-namespace" not found
hades01:cluster-network-addons-operator $ kubectl label namespace ram  mutatevirtualmachines.kubemacpool.io=bla
error: 'mutatevirtualmachines.kubemacpool.io' already has a value (allocate), and --overwrite is false
hades01:cluster-network-addons-operator $ kubectl label namespace ram  mutatevirtualmachines.kubemacpool.io=bla --overwrite
namespace/ram labeled
```

phoracek: What do you suggest then?

RamLavi: What I wrote above is not correct, sorry.
This namespace should never be opted in, so it's ok to add this label.

RamLavi:
/lgtm
/approve

@phoracek please run `make generate generate-test generate-deploy` to make sure all manifests are up to date in order for the travis  to pass

config/default/manager/manager.yaml
phoracek: Added back

phoracek:
@RamLavi there was some reformatting introduced in a previous PR. Weird that it passed. Should be fixed now.

config/default/manager/manager.yaml
```diff
@@ -3,9 +3,9 @@ kind: Namespace
 metadata:
   labels:
     control-plane: mac-controller-manager
-    controller-tools.k8s.io: "1.0"
-    runlevel: "0"
-    openshift.io/run-level: "0"
+    # Make sure that KubeMacPool pods are never handled by the webhook to
+    # prevent dead-lock.
+    mutatepods.kubemacpool.io: ignore
```
RamLavi, line 8: maybe we should also add the label of virtualmachines with ignore value?

phoracek: The point here is to make sure we don't endup in a deadlock where a Pod that implements KMP is not stuck on creation waiting for itself. Unless we add management VMs implementing KMP, we don't need to ignore them.

RamLavi: you're right.

PR #207, opened 2020-06-23
change kubemacpool tests to work on CNAO based clusters

RamLavi:

Currently kubemacpool u/s tests running on cnao cluster encounter the operator reconcile that returns the range values to original set range. in order to allow tests to run without this reconcile, we also initiate cnao deployment restart (if needed) at the beginning of each test.

**Release note**:

```release-note
NONE
```


RamLavi:
won't work. CNAO will not let us change the mac range.

PR #209, opened 2020-06-25
fix leadership label removed after twice winning election

RamLavi:

Currently, due to faulty labeling logic, if a pod won leadership twice it will loose it's leadership label.
This commit fixes that, and also adds unit tests to prevent this from happening.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/manager/leaderelection.go
+++ b/pkg/manager/leaderelection.go
@@ -28,7 +28,7 @@ func (k *KubeMacPoolManager) waitToStartLeading(poolManger *poolmanager.PoolMana
 		return errors.Wrap(err, "failed to start pool manager routines")
 	}
 
-	err = k.UpdateLeaderLabel()
+	err = updateLeaderLabel(k.runtimeManager.GetClient(), k.podName, k.podNamespace)
 	if err != nil {
 		return errors.Wrap(err, "failed marking pod as leader")
 	}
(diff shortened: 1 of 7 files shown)


pkg/manager/leaderelection.go
```diff
@@ -57,11 +57,11 @@ func (k *KubeMacPoolManager) UpdateLeaderLabel() error {
 				return errors.Wrap(err, "failed to get kubemacpool manager pod")
 			}
 
-			_, exist := pod.Labels[names.LEADER_LABEL]
+			_, exist := pod.GetLabels()[names.LEADER_LABEL]
 			if pod.Name == k.podName && !exist {
```
qinqon, line 61: Adding label or delete it at already added/delete does not fail so we can simplify this by 
```golang
                     if pod.Name == k.podName {
				logger.V(1).Info("add the label to the elected leader", "Pod Name", pod.Name)
				pod.Labels[names.LEADER_LABEL] = "true"
			} else {
				logger.V(1).Info("deleting leader label from old leader", "Pod Name", pod.Name)
				delete(pod.Labels, names.LEADER_LABEL)
			}
```
We can before that is not problem with it at unit-test

RamLavi: the logic is simpler but we will update all pods even if there was no label change in them, shouldn't we avoid that?

qinqon: Yaml is going to be the same so they don't get updated also we call `Update` out of the condition.

pkg/manager/leaderelection.go
```diff
@@ -57,11 +57,11 @@ func (k *KubeMacPoolManager) UpdateLeaderLabel() error {
...
 				logger.V(1).Info("add the label to the elected leader", "Pod Name", pod.Name)
```
qinqon, line 62: can you remove `V(1)` here ? it does not happend very often and it helps debug d/s

pkg/manager/leaderelection.go
RamLavi: ack

RamLavi:
/hold

need to add unit test commit

pkg/manager/leaderelection.go
```diff
@@ -41,34 +41,34 @@ func (k *KubeMacPoolManager) waitToStartLeading(poolManger *poolmanager.PoolMana
 }
 
 // Adds the leader label to elected pod and removes it from all the other pods, if exists
-func (k *KubeMacPoolManager) UpdateLeaderLabel() error {
+func (k *KubeMacPoolManager) UpdateLeaderLabel(kubeClient client.Client) error {
```
qinqon, line 44: We don't need method receiver anymore since `k` is not used also this function can be private it's only used at this package.

RamLavi: well it's used by the test so I think I can't make it private

qinqon: Test is insde the package you can make it private

PR #210, opened 2020-06-25
Add leader election with pod restart e2e test

qinqon:
There is a bug that cleans up the leader label if the pod get restarted and
elected again as leader, to be sure the pod is elected again the test
set the replicas to 1

It copy paste some test packages from kubernetes-nmstate

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/functest.sh
+++ b/hack/functest.sh
@@ -4,4 +4,4 @@ set -xe
 
 source ./cluster/kubevirtci.sh
 
-KUBECONFIG=${KUBECONFIG:-$(kubevirtci::kubeconfig)} $GO test ./tests/... $E2E_TEST_ARGS -timeout=40m -ginkgo.v -test.v -race
+KUBECONFIG=${KUBECONFIG:-$(kubevirtci::kubeconfig)} $GO test ./tests $E2E_TEST_ARGS -timeout=40m -ginkgo.v -test.v -race
--- a/tests/environment/environment.go
+++ b/tests/environment/environment.go
@@ -1,0 +1,13 @@
+package environment
+
+import (
+	"os"
+)
+
+func GetVarWithDefault(name string, defaultValue string) string {
+	value := os.Getenv(name)
+	if len(value) == 0 {
+		value = defaultValue
+	}
+	return value
+}
--- a/tests/leader_test.go
+++ b/tests/leader_test.go
@@ -1,0 +1,49 @@
+package tests
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/k8snetworkplumbingwg/kubemacpool/tests/cmd"
+)
+
+var _ = Describe("Leader election mechanism", func() {
+	Context("when kubemacpool pod is re-started with 1 replica", func() {
+		BeforeEach(func() {
+			By("Changing replicas to 0 to start from the beginning")
+			err := changeManagerReplicas(0)
+			Expect(err).To(Succeed(), "should succeed changing replicas to 0")
+
+			By("Changing replicast to 1 to have just on pod with leader election")
+			err = changeManagerReplicas(1)
+			Expect(err).To(Succeed(), "should succeed changing replicas to 1")
+
+			By("Retrieving leader pod")
+			leaderPod, err := getLeaderPod()
+			Expect(err).To(Succeed(), "should succeed getting leader kubemacpool pod")
+
+			By("Sending SIGTERM signal to restart leader pod")
+			killCmd := []string{"exec", "-n", leaderPod.Namespace, leaderPod.Name, "--", "/bin/bash", "-c", "kill -s SIGTERM 1"}
+			output, err := cmd.Kubectl(killCmd...)
+			Expect(err).To(Succeed(), fmt.Sprintf("should succeed killing leader kubemacpool pod: %s", output))
+			waitManagerDeploymentReady()
+
+		})
+		AfterEach(func() {
+			err := changeManagerReplicas(0)
+			Expect(err).To(Succeed(), "should succeed changing replicas to 0")
+
+			err = changeManagerReplicas(2)
+			Expect(err).To(Succeed(), "should succeed changing replicas to 2")
+		})
+		It("should preserver leader label", func() {
+			Consistently(func() error {
+				_, err := getLeaderPod()
+				return err
+			}, 1*time.Minute, 1*time.Second).Should(Succeed(), "should succeed getting leader kubemacpool pod")
+		})
+	})
+})
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -21,11 +21,10 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/envtest"
-
 	kubevirtv1 "kubevirt.io/client-go/api/v1"
 	kubevirtutils "kubevirt.io/kubevirt/tools/vms-generator/utils"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
 
 	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 )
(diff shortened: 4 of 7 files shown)


tests/virtual_machines_test.go
```diff
@@ -603,6 +604,41 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+
+				By("Changing replicast to 1 to have just on pod with leader election")
+				err = changeManagerReplicas(1)
+				Expect(err).To(Succeed(), "should succeed changing replicas to 1")
+
+				By("Retrieving leader pod")
+				leaderPod, err := getLeaderPod()
+				Expect(err).To(Succeed(), "should succeed getting leader kubemacpool pod")
+
+				By("Seding SIGTERM signal to restart leader pod")
```
RamLavi, line 621: typo Seding-->Sending

tests/virtual_machines_test.go
```diff
@@ -603,6 +604,41 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+
+			})
+			AfterEach(func() {
+				err := changeManagerReplicas(0)
+				Expect(err).To(Succeed(), "should succeed changing replicas to 0")
+
+				err = changeManagerReplicas(2)
+				Expect(err).To(Succeed(), "should succeed changing replicas to 2")
+			})
+			FIt("should preserver leader label", func() {
```
RamLavi, line 635: please remove Force flag `F`

tests/tests.go
```diff
@@ -289,6 +298,30 @@ func ChangeManagerLeadership() {
...
+	Eventually(func() (corev1.ConditionStatus, error) {
+		managerDeployment, err := testClient.KubeClient.AppsV1().Deployments(managerNamespace).Get(context.TODO(), names.MANAGER_DEPLOYMENT, metav1.GetOptions{})
+		if err != nil {
+			return corev1.ConditionUnknown, err
+		}
+		pods, err := getKubemacpoolPods()
+		if err != nil {
+			return corev1.ConditionUnknown, err
+		}
+		if int32(len(pods.Items)) != *managerDeployment.Spec.Replicas {
```
RamLavi, line 312: why not just compare deployment.available to deployment.replicas? I mean like in the original implementation:
```
if managerDeployment.Status.ReadyReplicas != numOfReplica {
```

tests/tests.go
```diff
@@ -242,16 +241,26 @@ func getVmFailCleanupWaitTime() time.Duration {
 	return 0
 }
 
-func ChangeManagerLeadership() {
+func getLeaderPod() (corev1.Pod, error) {
 	filterByLeaderLabel := client.MatchingLabels{
 		"kubemacpool-leader": "true",
 	}
+	fmt.Println()
```
RamLavi, line 248: please remove

tests/tests.go
```diff
@@ -289,6 +298,30 @@ func ChangeManagerLeadership() {
 	}, 2*time.Minute, 3*time.Second).Should(Equal(corev1.ConditionTrue), "should have a leader manager pod with ready condition")
 }
 
+func waitManagerDeploymentReady() {
+	By(fmt.Sprintf("Waitting for manager deployment to be ready"))
```
RamLavi, line 302: if you're not using a parameter then you can drop the fmt.Sprintf()

RamLavi, line 302: typo Waitting-->Waiting

hack/functest.sh
```diff
@@ -4,4 +4,4 @@ set -xe
 
 source ./cluster/kubevirtci.sh
 
-KUBECONFIG=${KUBECONFIG:-$(kubevirtci::kubeconfig)} $GO test ./tests/... $E2E_TEST_ARGS -timeout=40m -ginkgo.v -test.v -race
+KUBECONFIG=${KUBECONFIG:-$(kubevirtci::kubeconfig)} $GO test ./tests $E2E_TEST_ARGS -timeout=40m -ginkgo.v -test.v -race
```
RamLavi, line 7: why do we drop the `...`? I thought it was like `-r`

tests/virtual_machines_test.go
```diff
@@ -603,6 +604,41 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+				err = changeManagerReplicas(1)
+				Expect(err).To(Succeed(), "should succeed changing replicas to 1")
+
+				By("Retrieving leader pod")
+				leaderPod, err := getLeaderPod()
+				Expect(err).To(Succeed(), "should succeed getting leader kubemacpool pod")
+
+				By("Seding SIGTERM signal to restart leader pod")
+				killCmd := []string{"exec", "-n", leaderPod.Namespace, leaderPod.Name, "--", "/bin/bash", "-c", "kill -s SIGTERM 1"}
+				output, err := cmd.Kubectl(killCmd...)
```
RamLavi, line 623: I don't like it cause it'll be tempting for future me not to use it even when not necessary. But I found no other way, so I guess it's ok. :)

phoracek:
The test passed on top of the new master with the fix in place.

tests/tests.go
qinqon: Doing changes from 2 to 1 was not working, since it has already 1 ready but the other pod is still terminating so it continues, checking number of pods is exactly what we need and with current implementation Deployment Available is usable too.

hack/functest.sh
qinqon: ginkgo disable live stream logs if it uses recursion and since we have just one test suite here is not needed.

tests/virtual_machines_test.go
qinqon: What do you mean ?

RamLavi: I mean it would be tempting to use "kubectl" cli commands instead of kubeclient instance..

RamLavi:
@qinqon from some reason once in like 10 runs, getLeaderPod() fails because waitManagerDeploymentReady() passes but there's no leader. it may be a timing thing.

qinqon:
running locally I get the following sometimes after restaring KMP
```
STEP: Retrieving leader pod
the server rejected our request for an unknown reason (get pods kubemacpool-mac-controller-manager-59b8fd9c79-mkn6f)
```
I think it has to retry on errors too.

tests/environment/environment.go
```diff
@@ -0,0 +1,13 @@
+package environment
+
+import (
+	"os"
+)
+
+func GetVarWithDefault(name string, defaultValue string) string {
+	value := os.Getenv(name)
```
ormergi, line 8: Shouldn't we handle or propagate the error `Getenv` return?

tests/tests.go
```diff
@@ -267,13 +275,13 @@ func ChangeManagerLeadership() {
...
 
 	By("Wait for the other pod to take over leadership")
-	Eventually(func() []corev1.Pod {
-		leaderPods = corev1.PodList{}
-		testClient.VirtClient.List(context.TODO(), &leaderPods, filterByLeaderLabel)
-		return leaderPods.Items
-	}, 2*time.Minute, 3*time.Second).Should(HaveLen(1), "should have just one leader manager pod after deleting previous leader")
+	leaderPod = corev1.Pod{}
+	Eventually(func() error {
+		var err error
```
ormergi, line 280: I think you can use `:=` instead of explicitly declare error variable 
```
Eventually(func() error {
    return _, err := getLeaderPod()
}
...
```

tests/leader_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/k8snetworkplumbingwg/kubemacpool/tests/cmd"
+)
+
+var _ = Describe("Leader election mechanism", func() {
+	FContext("when kubemacpool pod is re-started with 1 replica", func() {
```
oshoval, line 14: FContext , you want it ?

tests/tests.go
```diff
@@ -289,6 +297,30 @@ func ChangeManagerLeadership() {
...
+		if int32(len(pods.Items)) != *managerDeployment.Spec.Replicas {
+			return corev1.ConditionUnknown, nil
+		}
+		for _, condition := range managerDeployment.Status.Conditions {
+			if condition.Type == v1.DeploymentAvailable {
+				return condition.Status, nil
+			}
+		}
+		return corev1.ConditionUnknown, nil
+	}, 2*time.Minute, 3*time.Second).Should(Equal(corev1.ConditionTrue), "should have available deployment after chaing numbe of replicas")
```
ormergi, line 322: I think the `Eventually` error message should not mention replicas change in `waitManagerDeploymentReady` function scope, instead this function should return an error then handle it in the caller function.

tests/virtual_machines_test.go
ormergi: Wouldn't it be enough to delete the pod through kube-api go client?

tests/environment/environment.go
qinqon: There is no error at `Getenv` https://golang.org/pkg/os/#Getenv

tests/tests.go
qinqon: If I do that it will create a different instance of `leaderPod` and I want to have `leaderPod` ouside of `Eventually`

qinqon:
/hold 

There is something fishy here, other tests are failing, I think the refactoring replicas change at test is affecting them.

tests/virtual_machines_test.go
qinqon: I you delete the pod you have new pod so leader election label is not there, what we want to test is that leader election is ok if the container inside the pod is restarted.

PR #214, opened 2020-06-29
Remove pod range change in tests

RamLavi:

Currently during each start of kubemacpool tier1 test, we perform an initilazaion of the mac-range configmap.
This ability is not supported outside of the kubemacpool standalone repo, and so it makes it a problem to run tier1 tests on CNAO (for example).
We would like to change the test behavior to stop using range change.
In order to do that, we will remove the mac range change, and instead change tests the check depleted mac pools, to allocate filler vms that will deplete the mac pool. Also, we will change the range in kubemacpool repo to 16 macs in the test manifest.
This way, if some external repo like CNAO wants to run teir1 tests, they can will skip these tests using the common phrase "mac-pool is full".

**Release note**:

```release-note
NONE
```

### DIFF
--- a/config/test/manager_range_patch.yaml
+++ b/config/test/manager_range_patch.yaml
@@ -1,0 +1,5 @@
+[
+{"op": "replace",
+ "path": "/data/RANGE_END",
+ "value": "02:00:00:00:00:0F"}
+]
--- a/pkg/pool-manager/pool.go
+++ b/pkg/pool-manager/pool.go
@@ -18,7 +18,6 @@ package pool_manager
 
 import (
 	"context"
-	"encoding/binary"
 	"fmt"
 	"net"
 	"sync"
@@ -28,6 +27,8 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/utils"
 )
 
 const (
@@ -214,15 +215,21 @@ func (p *PoolManager) isInstanceOptedIn(namespaceName, mutatingWebhookConfigName
 	return false, nil
 }
 
-func GetMacPoolSize(rangeStart, rangeEnd net.HardwareAddr) (uint64, error) {
+func GetMacPoolSize(rangeStart, rangeEnd net.HardwareAddr) (int64, error) {
 	err := checkRange(rangeStart, rangeEnd)
 	if err != nil {
 		return 0, errors.Wrap(err, "mac Pool Size  is negative")
 	}
 
-	//making sure Uint64 receives an byte slice of size 8
-	startInt := binary.BigEndian.Uint64(append([]byte{0x0, 0x0}, rangeStart...))
-	endInt := binary.BigEndian.Uint64(append([]byte{0x0, 0x0}, rangeEnd...))
+	startInt, err := utils.ConvertHwAddrToInt64(rangeStart)
+	if err != nil {
+		return 0, errors.Wrap(err, "error converting rangeStart to int64")
+	}
+
+	endInt, err := utils.ConvertHwAddrToInt64(rangeEnd)
+	if err != nil {
+		return 0, errors.Wrap(err, "error converting rangeEnd to int64")
+	}
 
 	return endInt - startInt + 1, nil
 }
--- a/pkg/pool-manager/pool.go
+++ b/pkg/pool-manager/pool.go
@@ -18,6 +18,7 @@ package pool_manager
 
 import (
 	"context"
+	"encoding/binary"
 	"fmt"
 	"net"
 	"sync"
@@ -211,4 +212,17 @@ func (p *PoolManager) isInstanceOptedIn(namespaceName, mutatingWebhookConfigName
 	}
 
 	return false, nil
+}
+
+func GetMacPoolSize(rangeStart, rangeEnd net.HardwareAddr) (uint64, error) {
+	err := checkRange(rangeStart, rangeEnd)
+	if err != nil {
+		return 0, errors.Wrap(err, "mac Pool Size  is negative")
+	}
+
+	//making sure Uint64 receives an byte slice of size 8
+	startInt := binary.BigEndian.Uint64(append([]byte{0x0, 0x0}, rangeStart...))
+	endInt := binary.BigEndian.Uint64(append([]byte{0x0, 0x0}, rangeEnd...))
+
+	return endInt - startInt + 1, nil
 }
--- a/pkg/pool-manager/pool_test.go
+++ b/pkg/pool-manager/pool_test.go
@@ -110,26 +110,26 @@ var _ = Describe("Pool", func() {
 
 		table.DescribeTable("should check that a mac pool size is reported correctly", func(startMacAddr, endMacAddr string, expectedSize float64, needToSucceed bool) {
 			startMacAddrHW, err := net.ParseMAC(startMacAddr)
-			Expect(err).ToNot(HaveOccurred())
+			Expect(err).ToNot(HaveOccurred(), "Should succeed parsing startMacAddr")
 			endMacAddrHW, err := net.ParseMAC(endMacAddr)
-			Expect(err).ToNot(HaveOccurred())
+			Expect(err).ToNot(HaveOccurred(), "Should succeed parsing endMacAddr")
 			poolSize, err := GetMacPoolSize(startMacAddrHW, endMacAddrHW)
 			if needToSucceed {
-				Expect(err).ToNot(HaveOccurred())
-				Expect(float64(poolSize)).To(Equal(expectedSize))
+				Expect(err).ToNot(HaveOccurred(), "Should succeed getting Mac Pool size")
+				Expect(float64(poolSize)).To(Equal(expectedSize), "Should get the expected pool size value")
 			} else {
-				Expect(err).To(HaveOccurred())
+				Expect(err).To(HaveOccurred(), "Should fail getting Mac Pool size duu to invalid params")
 			}
 		},
-			table.Entry("Start: 40:00:00:00:00:00  End: 50:00:00:00:00:00", "40:00:00:00:00:00", "50:00:00:00:00:00", math.Pow(2, 11*4)+1, true),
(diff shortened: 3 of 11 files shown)


config/test/manager_range_patch.yaml
```diff
@@ -0,0 +1,5 @@
+[
+{"op": "replace",
+ "path": "/data/RANGE_END",
+ "value": "02:00:00:00:00:0F"}
```
qinqon, line 4: Isn't it redundant since it's alrady set to `02:00:00:00:00:0F` at `test/kubemacpool.yaml` ?

pkg/pool-manager/pool.go
```diff
@@ -212,3 +213,16 @@ func (p *PoolManager) isInstanceOptedIn(namespaceName, mutatingWebhookConfigName
...
 	return false, nil
 }
+
+func GetMacPoolSize(rangeStart, rangeEnd net.HardwareAddr) (uint64, error) {
+	err := checkRange(rangeStart, rangeEnd)
+	if err != nil {
+		return 0, errors.Wrap(err, "mac Pool Size  is negative")
+	}
+
+	//making sure Uint64 receives an byte slice of size 8
```
qinqon, line 223: Let's add a helper innter function with
```golang
addTwoBytesPadding = func(....
startInt := addTwoBytesPadding(rangeStart
endInt := addTwoBytesPadding(rangeEnd
```

pkg/pool-manager/pool_test.go
```diff
@@ -106,6 +107,30 @@ var _ = Describe("Pool", func() {
 			table.Entry("Invalid address: 01:FF:00:00:00:00, the first octet is not 02, 06, 0A or 0E", "01:FF:00:00:00:00", true),
 			table.Entry("Invalid address: FF:FF:00:00:00:00, the first octet is not 02, 06, 0A or 0E", "FF:FF:00:00:00:00", true),
 		)
+
+		table.DescribeTable("should check that a mac pool size is reported correctly", func(startMacAddr, endMacAddr string, expectedSize float64, needToSucceed bool) {
+			startMacAddrHW, err := net.ParseMAC(startMacAddr)
+			Expect(err).ToNot(HaveOccurred())
```
qinqon, line 113: Add "should..."

pkg/pool-manager/pool_test.go
```diff
@@ -106,6 +107,30 @@ var _ = Describe("Pool", func() {
...
+			endMacAddrHW, err := net.ParseMAC(endMacAddr)
+			Expect(err).ToNot(HaveOccurred())
```
qinqon, line 115: Ditto

pkg/pool-manager/pool_test.go
qinqon: Isn't this already check at `checkRange` ?

pkg/pool-manager/pool_test.go
```diff
@@ -106,6 +107,30 @@ var _ = Describe("Pool", func() {
...
+			table.Entry("Start: 02:00:00:00:00:00  End: 03:00:00:00:00:00", "02:00:00:00:00:00", "03:00:00:00:00:00", math.Pow(2, 10*4)+1, true),
+			table.Entry("Start: 02:00:00:00:00:00  End: 02:01:00:00:00:00", "02:00:00:00:00:00", "02:01:00:00:00:00", math.Pow(2, 8*4)+1, true),
+			table.Entry("Start: 02:00:00:00:00:00  End: 02:00:00:10:00:00", "02:00:00:00:00:00", "02:00:00:10:00:00", math.Pow(2, 5*4)+1, true),
+			table.Entry("Start: 02:00:00:00:00:10  End: 02:00:00:00:00:00", "02:00:00:00:00:00", "02:00:00:00:00:10", math.Pow(2, 1*4)+1, true),
+			table.Entry("Start: 00:00:00:00:00:01  End: 00:00:00:00:00:00", "00:00:00:00:00:01", "00:00:00:00:00:00", float64(0), false),
+			table.Entry("Start: 80:00:00:00:00:00  End: 00:00:00:00:00:00", "80:00:00:00:00:00", "00:00:00:00:00:00", float64(0), false),
+			table.Entry("Start: FF:FF:FF:FF:FF:FF  End: FF:FF:FF:FF:FF:FF", "FF:FF:FF:FF:FF:FF", "FF:FF:FF:FF:FF:FF", float64(0), false),
+			table.Entry("Start: 00:00:00:00:00:00  End: 00:00:00:00:00:00", "00:00:00:00:00:00", "00:00:00:00:00:00", float64(0), false),
+		)
```
qinqon, line 133: It needs something at the end of the description saying that it should fail or not.

tests/virtual_machines_test.go
```diff
@@ -298,48 +279,55 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 		})
 		Context("When trying to create a VM after all MAC addresses in range have been occupied", func() {
 			It("[test_id:2162]should return an error because no MAC address is available", func() {
-				err := initKubemacpoolParams("02:00:00:00:00:00", "02:00:00:00:00:01")
-				Expect(err).ToNot(HaveOccurred())
-
+				if shouldSkipTest() {
+					Skip("Skipping test on production platforms")
```
qinqon, line 283: I think we can call `Skip` inside `shouldSkipTest` so we don't need the if statement the function should be named then `skipTestIfNeeded` or the like, We can even call it at a `BeforeEach` so e can forget about it.

Also the Skip string is different at all the calls here.

tests/virtual_machines_test.go
```diff
@@ -627,6 +598,26 @@ func newNetwork(name string) kubevirtv1.Network {
 	}
 }
 
+// This function allocates vms with 1 NIC each, in order to fill the mac pool to the brim.
+func AllocateVmsUntilFull(macsToLeaveFree uint64) error {
+	maxPoolSize := getMacPoolSize()
+	Expect(maxPoolSize > macsToLeaveFree).To(BeTrue(), "max pool size must be greater than the number of macs we want to leave free")
```
qinqon, line 604: Use gomega `BeNumerically(">",` as a rule of thumb use the most specific gomega matcher always so you get proper expected/obtained messages.

tests/virtual_machines_test.go
```diff
@@ -627,6 +598,26 @@ func newNetwork(name string) kubevirtv1.Network {
 	}
 }
 
+// This function allocates vms with 1 NIC each, in order to fill the mac pool to the brim.
+func AllocateVmsUntilFull(macsToLeaveFree uint64) error {
```
qinqon, line 602: Well the name is not always true since we pass `macsToLeaveFree`

pkg/pool-manager/pool.go
```diff
@@ -212,3 +213,16 @@ func (p *PoolManager) isInstanceOptedIn(namespaceName, mutatingWebhookConfigName
...
+	startInt := binary.BigEndian.Uint64(append([]byte{0x0, 0x0}, rangeStart...))
```
ormergi, line 224: How about converting `rangeStart` and `rangEnd` to integers then calculate the range length,
that way you dont need mess with binary representation, padding, Endians etc...

RamLavi: That's basically what I did :) 
rangeStart/End are basically of type []byte. I found [this](https://stackoverflow.com/questions/11184336/how-to-convert-from-byte-to-int-in-go-programming) to convert it to a interger, but am open to other suggestions =)

tests/virtual_machines_test.go
```diff
@@ -298,13 +279,13 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 		})
 		Context("When trying to create a VM after all MAC addresses in range have been occupied", func() {
 			PIt("[test_id:2162]should return an error because no MAC address is available", func() {
-				err := initKubemacpoolParams("02:00:00:00:00:00", "02:00:00:00:00:01")
-				Expect(err).ToNot(HaveOccurred())
+				//err := initKubemacpoolParams("02:00:00:00:00:00", "02:00:00:00:00:01")
```
ormergi: Is there a reason that those lines are commented or can we remove them?

RamLavi: I saved them until the last commit and them removed them.

pkg/pool-manager/pool.go
ormergi: Look at this example form shading branch
https://github.com/k8snetworkplumbingwg/kubemacpool/pull/122/commits/1972f47a956d57049942092feda7e363299af1a5#diff-baa41b07e1d50dce50eab577f6b940d5R125

config/test/manager_range_patch.yaml
RamLavi: the change in `test/kubemacpool.yaml` is generated by this patch, using kustomize.
we want this change to only affect the test manifest.

pkg/pool-manager/pool.go
RamLavi: Like!

pkg/pool-manager/pool.go
RamLavi: if it's ok I'll use the function Or suggested: 1972f47#diff-baa41b07e1d50dce50eab577f6b940d5R125

pkg/pool-manager/pool_test.go
RamLavi: nope. `checkRange` only checks if one is not bigger than the other without parsing to int.

tests/virtual_machines_test.go
RamLavi: ok I moved it inside `skipTestIfNeeded`. we can't move it to `BeforeEach` though since we only want these 3 test to be skipped.

tests/virtual_machines_test.go
RamLavi: ack

RamLavi:
@qinqon and @ormergi had suggested to replace the `skipTestIfNeeded()` method with an external skip that will be added to CNAO & HCO clusters running kmp teir1 tests. Or using maybe some regex to catch these 3 tests, similar to how they do it on [kubevirt](https://github.com/kubevirt/kubevirt/blob/9e193c66d1b96b50483c113d4ae86fa9c28b77c9/automation/test.sh#L321).

@phoracek what do you think?

config/test/manager_range_patch.yaml
qinqon: I am stoopid, hehe ok .

RamLavi:
/hold

we may want to keep tes_id 2995

RamLavi:
unhold

PR #217, opened 2020-07-01
Fix external provider cluster up

RamLavi:

Currently external provider does not exclude deployment of private CNAO instance and kubevirt deployment.
We want to ignore apply of private CNAO component, since if run in external provider, they are already deployed by external cluster.
Also, we want to change kubevirt deployment to be optional by env var, should we want to run in hco cluster where kubevirt is already deployed.

**Release note**:

```release-note

```

### DIFF
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -19,43 +19,45 @@ set -ex
 source ./cluster/kubevirtci.sh
 CNAO_VERSIOV=0.35.0
 KUBEVIRT_VERSION=v0.29.0
+DEPLOY_KUBEVIRT=${DEPLOY_KUBEVIRT:-true}
 kubevirtci::install
 
-if [[ "$KUBEVIRT_PROVIDER" != external ]]; then
+if [[ "${KUBEVIRT_PROVIDER}" != external ]]; then
     $(kubevirtci::path)/cluster-up/up.sh
+
+    # Deploy CNAO
+    ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/namespace.yaml
+    ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/network-addons-config.crd.yaml
+    ./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/operator.yaml
+    ./cluster/kubectl.sh create -f ./hack/cna/cna-cr.yaml
+
+    # wait for cluster operator
+    ./cluster/kubectl.sh wait networkaddonsconfig cluster --for condition=Available --timeout=800s
 fi
 
-# Deploy CNA
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/namespace.yaml
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/network-addons-config.crd.yaml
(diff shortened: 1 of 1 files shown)


cluster/up.sh
```diff
@@ -19,43 +19,47 @@ set -ex
 source ./cluster/kubevirtci.sh
 CNAO_VERSIOV=0.35.0
 KUBEVIRT_VERSION=v0.29.0
+DEPLOY_KUBEVIRT=${DEPLOY_KUBEVIRT:-true}
 kubevirtci::install
 
-if [[ "$KUBEVIRT_PROVIDER" != external ]]; then
-    $(kubevirtci::path)/cluster-up/up.sh
-fi
+$(kubevirtci::path)/cluster-up/up.sh
```
ormergi, line 25: You should remove this line, in case `KUBEVIRT_PROVIDER=external` a cluster will be created

cluster/up.sh
```diff
@@ -19,43 +19,47 @@ set -ex
...
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/namespace.yaml
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/network-addons-config.crd.yaml
-./cluster/kubectl.sh create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${CNAO_VERSIOV}/operator.yaml
-./cluster/kubectl.sh create -f ./hack/cna/cna-cr.yaml
+if [[ "${KUBEVIRT_PROVIDER}" != external ]]; then
+    $(kubevirtci::path)/cluster-up/up.sh
 
-# wait for cluster operator
-./cluster/kubectl.sh wait networkaddonsconfig cluster --for condition=Available --timeout=800s
+    # Deploy CNAO
```
ormergi, line 30: Do we want to assume that external cluster should have CNAO deployment?

RamLavi: mmm... that's a valid question. I assumed so, I not really necessary I guess. what do you propose?

cluster/up.sh
RamLavi: yeah it's a duplicate sorry.

cluster/up.sh
RamLavi: ok I discussed this with @phoracek and in case of external provider we should not deploy this, as it is the responsibility of the external cluster itself

phoracek: I believe that cluster-up should be no-op on external cluster. We should be only interested in cluster-sync.

RamLavi: mmmm... see I thought we could use `make cluster-up` to deploy kubevirt on CNAO repo before running the tier 1 tests. 
Should we do this seperately then? @phoracek

phoracek: I would not want an external testing script to install something unexpected on my cluster. Explicitly installing KubeVirt on CNAO side would be preferable IMO. Or keeping a separate helper script on KMP, but executed explicitly by owner of the external cluster.

RamLavi: ack, 
I want to avoid having to bump the KUBEVIRT_VERSION in both the CNAO teir1 test script & KMP repo, so I think a separate script on KMP is the best choise here.

phoracek: Or just always fetch the latest: https://github.com/kubevirt/kubevirt/releases/latest

ormergi: kubevirt_deploy script should be everywhere you want to deploy kubevirt, this must be automated and separated from the main cluster deployment, we can get kubevirt version as a parameter so it will be easier to maintain.
IMO there is no problem having kubevirt_deploy in kubemacpool and cnao these are different projects

phoracek: Well that won't help with downloading the artefact directly but I'm sure there is a way to curl your way there.

RamLavi:
closing this PR in favor of using tier 1 script withot cluster-up, and also opening another PR to use kubevirt latest release version

PR #218, opened 2020-07-02
move to use latest z stream kubevirt release

RamLavi:

Currently we're using a fixed version of kubevirt.
Lets move to use the latest z stram  release kubevirt release of a fixed major, and benefit from the bug improvements and featuresthat the latest stable release has to offer.
master branch can use fixed major-minor of v0.31. 
I think it's not very risky considering kubevirt's role in kubemacpool operation.

**Release note**:

```release-note
use kubevirt latest release
```

### DIFF
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -14,11 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -ex
+set -ex pipefail
+
+function getLatestPatchVersion {
+  local major_minors=$1
+  curl -s https://api.github.com/repos/kubevirt/kubevirt/releases | grep .tag_name | grep ${major_minors} | sort -V | tail -1 | awk -F':' '{print $2}' | sed 's/,//' | xargs
+}
 
 source ./cluster/kubevirtci.sh
 CNAO_VERSIOV=0.35.0
-KUBEVIRT_VERSION=v0.29.0
+#use kubevirt latest z stream release
+
+KUBEVIRT_VERSION=$(getLatestPatchVersion v0.31)
 kubevirtci::install
 
 if [[ "$KUBEVIRT_PROVIDER" != external ]]; then


cluster/up.sh
```diff
@@ -18,7 +18,8 @@ set -ex
 
 source ./cluster/kubevirtci.sh
 CNAO_VERSIOV=0.35.0
-KUBEVIRT_VERSION=v0.29.0
+#use klubevirt latest release
```
phoracek, line 21: typo

cluster/up.sh
```diff
@@ -18,7 +18,8 @@ set -ex
 
 source ./cluster/kubevirtci.sh
 CNAO_VERSIOV=0.35.0
-KUBEVIRT_VERSION=v0.29.0
+#use kubevirt latest release
+KUBEVIRT_VERSION=$(curl -s https://api.github.com/repos/kubevirt/kubevirt/releases/latest | grep .tag_name | sort -V | tail -1 | awk -F':' '{print $2}' | sed 's/,//' | xargs)
```
qinqon, line 22: We have to be very careful here since latest is not always latest, for example the latest could be a z-stream update and comming from release branch and that not the latest from master, so I propose that we mark the major.minor jere and take the latest z stream for example right now we have the following versions at kubevirt:
- v0.31.0-rc.1
- v0.30.3

the release branch KMP `release-v0.14.0` has to take `v0.30.3` but master has to take `v0.31.0-rc.1` so ideally what we would have at KUBEVIRT_VERSION is 0.30,0.31 etc...

cluster/up.sh
```diff
@@ -18,7 +18,9 @@ set -ex
 
 source ./cluster/kubevirtci.sh
 CNAO_VERSIOV=0.35.0
-KUBEVIRT_VERSION=v0.29.0
+KUBEVIRT_FIXED_MAJOR =v0.31
```
qinqon, line 21: Name this KUBEVIRT_MAJOR_MINOR or KUBEVIRT_VERSION since FIXED does not means anything here.

RamLavi: ack you're right

cluster/up.sh
```diff
@@ -18,7 +18,9 @@ set -ex
 
 source ./cluster/kubevirtci.sh
 CNAO_VERSIOV=0.35.0
-KUBEVIRT_VERSION=v0.29.0
+KUBEVIRT_MAJOR_MINOR=v0.31
+#use kubevirt latest z stream release
+KUBEVIRT_VERSION=$(curl -s https://api.github.com/repos/kubevirt/kubevirt/releases | grep .tag_name | grep ${KUBEVIRT_MAJOR_MINOR} | sort -V | tail -1 | awk -F':' '{print $2}' | sed 's/,//' | xargs)
```
qinqon, line 23: Can we wrap this into a function ? so we use same for CNAO

cluster/up.sh
```diff
@@ -16,9 +16,16 @@
 
 set -ex
 
+function getLatestZStream {
+  KUBEVIRT_MAJOR_MINOR=v0.31
```
qinqon, line 20: Three comments here:
- Parameterize the function
```
local major_minor=$1
- `echo` is not needed
- also set `pipefail` -> `set -ex pipefail`
```

cluster/up.sh
```diff
@@ -16,9 +16,17 @@
 
 set -ex
 
+function getLatestZStream {
+  local major_minors=$1
+  set -ex pipefail
```
qinqon, line 21: This goes up in the script

PR #220, opened 2020-07-02
[WIP] Bump admissionregistration to v1

RamLavi:

Bump admissionregistration to v1.
We're currently using mutatingWebhookConfiguration but using v1beta1. Let's bump to v1

**Release note**:

```release-note
NONE
```

### DIFF
--- a/config/default/manager/manager.yaml
+++ b/config/default/manager/manager.yaml
@@ -135,7 +135,7 @@ spec:
   selector:
     app: kubemacpool
 ---
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
   # metadata.name will be kubemacpool-mutator (prefix added by kustomize)
@@ -150,6 +150,8 @@ webhooks:
         namespace: kubemacpool-system
         path: "/mutate-pods"
     failurePolicy: Fail
+    sideEffects: None
+    admissionReviewVersions: ["v1"]
     namespaceSelector:
       matchExpressions:
       - key: runlevel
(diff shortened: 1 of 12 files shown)


config/release/kubemacpool.yaml
```diff
@@ -5,13 +5,15 @@ metadata:
...
 ---
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
   labels: null
   name: kubemacpool-mutator
 webhooks:
-- clientConfig:
+- admissionReviewVersions:
```
RamLavi, line 14:
```
AdmissionReviewVersions is an ordered list of preferred AdmissionReview versions the Webhook expects. API server will try to use first version in the list which it supports. If none of the versions specified in this list supported by API server, validation will fail for this object. If a persisted webhook configuration specifies allowed versions and does not include any versions known to the API Server, calls to the webhook will fail and be subject to the failure policy.
```

If I understand this correctly, this needs to hold the **pod**'s supported version(since this webhook checks pods), right?

config/release/kubemacpool.yaml
```diff
@@ -43,7 +45,10 @@ webhooks:
     - CREATE
     resources:
     - pods
-- clientConfig:
+  sideEffects: None
+- admissionReviewVersions:
```
RamLavi, line 49: similarly, this needs to hold the **vm**'s supported version(since this webhook checks vm), right?

config/default/manager/manager.yaml
```diff
@@ -150,6 +150,8 @@ webhooks:
         namespace: kubemacpool-system
         path: "/mutate-pods"
     failurePolicy: Fail
+    sideEffects: None
```
qinqon, line 153: Are we sure this is none ?

qinqon (commented):
I think we are chaning the VMI some maybe sideEffects are not `None`

RamLavi:
/hold

we can't run it until controller-runtime issues a release with this [PR](https://github.com/kubernetes-sigs/controller-runtime/pull/1233), adding admission v1

PR #230, opened 2020-07-15
waitToStartLeading goroutine does not catch error.

RamLavi:
Let's add a wrapping function to catch the errs.
If an Error occur - also restart pod to initiate election.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -106,7 +106,13 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 		}
 		go k.waitForSignal()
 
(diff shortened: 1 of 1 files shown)


pkg/manager/manager.go
```diff
@@ -106,7 +106,13 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 		}
 		go k.waitForSignal()
 
-		go k.waitToStartLeading(poolManager)
+		go func() {
+			err := k.waitToStartLeading(poolManager)
```
phoracek, line 110: Remind me, why is this a goroutine? Shouldn't we block on `waitToStartLeading`?

PR #238, opened 2020-08-06
Set number of replicas to 1 and drop leader election

phoracek:

#### Drop the number of replicas to 1 and remove leader
    
We kept two replicas so there is one KMP manager running in the
background ready to take over the leadership when the original leader
dies. That should roughly lower the downtime on master failure from 4 to
2 minutes. However, we faced countless issues with the leadership
mechanism and thus decided to keep only a single replica. We should then
work on profiling and figure out what is taking so much time on the
startup without preemptive optimization.

#### Remove all leader election code
    
Now when we run in a single replica and upgrade strategy is "recreate"
we can drop all the leadership election code. The code was quite
complicated and regularly causing bugs.
    
I did some basic comparison between the current active-backup approach and proposed 1 pod:

* 1 pod: first start took 70 seconds; restart on the same host, able to serve in 8 seconds
* active-backup: first start took 90 seconds; when leader killed, able to serve in 20 seconds

**Release note**:

```release-note
Set number of replicas to 1 and drop leader election
```


config/default/manager/manager.yaml
```diff
@@ -31,7 +31,7 @@ metadata:
 spec:
   strategy:
     type: Recreate
-  replicas: 2
+  replicas: 1
```
qinqon, line 34: Maybe now we can go back to `RollingUpdate` ?

phoracek: That may be dangerous. We don't have locking and we don't want the old pod to run at the same time as the new one.

tests/virtual_machines_test.go
```diff
@@ -517,42 +517,6 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 			})
 		})
 
-		Context("When the leader is changed", func() {
```
RamLavi, line 520: Maybe it's  not directly relevant to this PR, but we should add another test, to check that kmp internal cache (of allocated macs) "migrates" safely (it doesn't really migrate but re-constructed, but still this is something we should make sure that happens).

PR #242, opened 2020-08-27
Ensure that the IMAGE_GIT_TAG has 8 chars at commit id

qinqon:
Depending on git version the `git describe --tags` has different commit id size
to ensure that is always the same we can specify `--abbrev=8` to ensure that
it uses 8 chars.

**Release note**:

```release-note
NONE
```


qinqon:

cluster-sync is working fine

PR #245, opened 2020-09-16
move 'test_id' tests to a new context, using the default opt-mode cluster config

RamLavi:
This PR moves polarion tests to a new context, using the default opt-mode config

- **move test_id related tests to a new context**
Currently all 'test_id' test are performed in opt-out context
There is no specific reason they were put there, since
they can run on either opt-mode, as long as they are running
on the same opt-mode configured by the cluster configuration (in order to prevent
the oprator deploying kubemacpool to reconcile back from the opt-mode configured by the test).
Let's therefor move the tests to a new context that will run using the
default opt-mode config that is set on the cluster.
Also this commit includes a small format fix in AllocateFillerVms

- **add BeforeEach to opt-In test**
Since opt-mode contexts may run in different order,
we need to make sure that opt-in context start with
a "fresh" namespace.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -489,13 +489,20 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 				Expect(err).ToNot(HaveOccurred(), "should set opt-mode to mutatingwebhookconfiguration")
 			})
 
-			It("should create a VM object without a MAC assigned", func() {
-				vm := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br", "")},
-					[]kubevirtv1.Network{newNetwork("br")})
+			Context("and kubemacpool is not opted-in on a namespace", func() {
+				BeforeEach(func() {
+					By("Removing any namespace labels to make sure that kubemacpool is not opted in on the namespace")
+					err := cleanNamespaceLabels(TestNamespace)
+					Expect(err).ToNot(HaveOccurred(), "should be able to remove the namespace labels")
+				})
+				It("should create a VM object without a MAC assigned", func() {
+					vm := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br", "")},
+						[]kubevirtv1.Network{newNetwork("br")})
 
-				err := testClient.VirtClient.Create(context.TODO(), vm)
-				Expect(err).ToNot(HaveOccurred(), "Should succeed creating the vm")
-				Expect(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress).Should(Equal(""), "should not allocated a mac to the opted-out vm")
+					err := testClient.VirtClient.Create(context.TODO(), vm)
+					Expect(err).ToNot(HaveOccurred(), "Should succeed creating the vm")
+					Expect(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress).Should(Equal(""), "should not allocated a mac to the opted-out vm")
+				})
 			})
 
 			Context("and kubemacpool is opted-in on a namespace", func() {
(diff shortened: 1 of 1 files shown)


tests/virtual_machines_test.go
```diff
@@ -317,320 +213,430 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+					})
+				})
+			})
+		})
+	})
+})
 
-					newVM1 := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br1", vm1.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress),
-						newInterface("br2", vm1.Spec.Template.Spec.Domain.Devices.Interfaces[1].MacAddress)}, []kubevirtv1.Network{newNetwork("br1"), newNetwork("br2")})
+func polarionTests() {
```
qinqon, line 243: We call `polarionTests` only at one place why do we need it?

RamLavi: we don't really. 
Done

alonSadan (commented):
it could be that it happens as it is now, and I missed something with the parentheses, but maybe it worth to re-arrange the tests order so that they will appear bu their context : first the default opt-mode then when we set it to opt-in the when we set it to opt out. something like that, will make it more readable I think.

other then that looks good!

RamLavi:
> it could be that it happens as it is now, and I missed something with the parentheses, but maybe it worth to re-arrange the tests order so that they will appear bu their context : first the default opt-mode then when we set it to opt-in the when we set it to opt out. something like that, will make it more readable I think.
> 
> other then that looks good!

it's already like that, the github compare scrambles it, but you can see it [here](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/29bedfd275c2f9ad42d620bb8ef8b08f5cc53dea/tests/virtual_machines_test.go)

alonSadan:
> > it could be that it happens as it is now, and I missed something with the parentheses, but maybe it worth to re-arrange the tests order so that they will appear bu their context : first the default opt-mode then when we set it to opt-in the when we set it to opt out. something like that, will make it more readable I think.
> > other then that looks good!
> 
> it's already like that, the github compare scrambles it, but you can see it [here](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/29bedfd275c2f9ad42d620bb8ef8b08f5cc53dea/tests/virtual_machines_test.go)

oh great! then
/lgtm

qinqon (changes requested):
PR comment an commit message: Please remove references to polarion an CNAO, talk about `test_id` tests and `cluster configuration`.

alonSadan (commented):
Tthe commit message says: "..we need to make sure that opt-in context start with..."
and the Context where the removal is made states:  " and kubemacpool is not opted-in on a namespace"
probably should be "opt-out" in the commit message. should it?

RamLavi:
> Tthe commit message says: "..we need to make sure that opt-in context start with..."
> and the Context where the removal is made states: " and kubemacpool is not opted-in on a namespace"
> probably should be "opt-out" in the commit message. should it?

the point is  to start the opt-in context "fresh", not assuming what was the context before we entered this context.
I'll reword the commit message

alonSadan:
> > Tthe commit message says: "..we need to make sure that opt-in context start with..."
> > and the Context where the removal is made states: " and kubemacpool is not opted-in on a namespace"
> > probably should be "opt-out" in the commit message. should it?
> 
> no. The test you mentioned is a negative test, meaning it's supposed to not allocate a mac - so we need to make sure the namespace is not opted in from last test.
> the point is to start the opt-in context "fresh", not assuming what was the context before we entered this context.
> 
> please let me know what do you think of the reworded commit message

understand. Thanks for explaining.
/lgtm

RamLavi:
/hold

wait for @qinqon re-review

PR #246, opened 2020-09-22
adjust test_id context to current opt-mode

RamLavi:
Currently Context: "When Running with default opt-mode configuration"
is only working on opt-out mode.
We need to adjust the test namespace according to
the currently configured opt-mode configuration

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"reflect"
 	"regexp"
 	"strconv"
 	"time"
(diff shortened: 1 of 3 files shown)


tests/virtual_machines_test.go
```diff
@@ -85,6 +85,18 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
 
 		Context("When Running with default opt-mode configuration", func() {
+			BeforeEach(func() {
+				By("Getting the current VM Opt-mode")
+				vmWebhookOptMode, err := getOptMode(vmNamespaceOptInLabel)
+				Expect(err).ToNot(HaveOccurred(), "Should succeed getting the mutating webhook opt-mode")
+
+				if vmWebhookOptMode == optInMode {
+					By("opting in the namespace")
+					err := addLabelsToNamespace(TestNamespace, map[string]string{vmNamespaceOptInLabel: "allocate"})
```
rhrazdil, line 95: Maybe we could make a constant for `allocate`, as it's used in two places

tests/tests.go
```diff
@@ -306,10 +308,38 @@ func addLabelsToNamespace(namespace string, labels map[string]string) error {
...
+// function checks what is the currently configured opt-mode configured in a specific webhook
+func getOptMode(webhookName string) (string, error) {
+	mutatingWebhook, err := testClient.KubeClient.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(context.TODO(), mutatingWebhookConfiguration, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+	for _, webhook := range mutatingWebhook.Webhooks {
+		if webhook.Name == webhookName {
+			for _, matchExpression := range webhook.NamespaceSelector.MatchExpressions {
+				optInLabelSelectorRequirement := metav1.LabelSelectorRequirement{Key: webhookName, Operator: "In", Values: []string{"allocate"}}
```
qinqon, line 320: Can we compare with content from patches ? https://github.com/k8snetworkplumbingwg/kubemacpool/blob/master/config/default/mutatepods_opt_in_patch.yaml

so we don't duplicate code.

qinqon (commented):
Just a comment we are ok if you thing is not worth it.

tests/tests.go
RamLavi: you mean I should parse `mutatevirtualmachines_opt_mode_patch.yaml` and extract the opt-mode from there
thing is - KMP repo runs using `test` folder  `config/test/mutatevirtualmachines_opt_mode_patch.yaml` while CNAO repo runs using `release` folder `config/release/mutatevirtualmachines_opt_mode_patch.yaml`

so it's not comparable without knowing what's the repo from which the test is run.. I can check by checking the namespace, but it's too messy so I think we should stick to the current approach

tests/virtual_machines_test.go
RamLavi: Done

tests/tests.go
qinqon: I would say that those patches belongs to a common place the don't differ from test and release, but maybe not at this patch.

RamLavi: agreed, I initially tried to put them in the pace to avoid duplicate patched but kustomize was stubborn, It didn't even eat a symbolic link pointing to a common file..
Let's try to tackle it in a different PR..

PR #253, opened 2020-11-26
Bump kubevirt to v0.34.2

RamLavi:
This PR bumps kubevirt to v0.34.2.

**Release note**:

```release-note
NONE
```


pkg/pool-manager/pool_test.go
```diff
@@ -289,6 +289,27 @@ var _ = Describe("Pool", func() {
 			Expect(exist).To(BeFalse())
 		})
 		Describe("Update vm object", func() {
+			It("should preserve disk.io configuration on update", func() {
```
qinqon, line 292: Is this related to the bumping ?

qinqon (commented):
Looks good, just a question

pkg/pool-manager/pool_test.go
RamLavi: yep.
when the namespace is opted-in, the webhook [gets](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/master/pkg/webhook/virtualmachine/virtualmachine.go#L132) the vm instance, but using a old version of kubvirt (0.29) that doesn't hold disk.io configuration.
so the disk.io conf is implicitly set to nil, and since it is set to `omitempty` then it disappears.
bumping the kubevirt to a version that has this conf fixes this, and this test makes sure it is indeed configured.

RamLavi:
/hold

need to understand if this PR is needed.

pkg/pool-manager/pool_test.go
RamLavi: not really. will issue in separate PR

PR #254, opened 2020-11-26
[release-v0.14.0] bump kubevirt to v0.34.2

RamLavi:
This PR bumps kubevirt to v0.34.2 on the stable branch, to support disk.io configuration.

**Release note**:

```release-note
NONE
```


cluster/up.sh
```diff
@@ -25,7 +25,7 @@ source ./cluster/kubevirtci.sh
 CNAO_VERSIOV=0.35.0
 #use kubevirt latest z stream release
 
-KUBEVIRT_VERSION=$(getLatestPatchVersion v0.30)
+KUBEVIRT_VERSION=$(getLatestPatchVersion v0.34)
```
phoracek, line 28: The bug https://bugzilla.redhat.com/show_bug.cgi?id=1901480 is for 2.6 which is yet to be branched out. This fix should be on the master.

In case we decide to fix the bug in a 2.5 z-stream, we would then needs a change similar to this (but I believe we should pin to v0.30.7)

RamLavi: closing the PR

PR #255, opened 2020-11-29
test 2995, Fix wrong expected error

RamLavi:
During test_id:2995 we allow getting an error saying that the mac
is still occupied until it is released by the controller.
The current error message is wrong, as the range is not full in this
specific test.
Fixed to the correct expected error.

**Release note**:

```release-note
NONE
```


ldean0558:
I'm getting this error in OCP 4.8.2 when I try to import a RHEL 7.9 VM from VMWare 6.7U3. This detail provides no details on how this was fixed, what version of OCP I need to run, etc. 

Likely I'm just not aware of the norms in this WG.

phoracek:
Hello @ldean0558. If you have an issue with OpenShift and expect a fix to arrive there, you should open a ticket/bugzilla. This repository is strictly for upstream and the only delivery we can help you with is the one done through this repository as described in the README.

PR #259, opened 2020-12-13
webhook Handle, patch only kubemacpool changes

RamLavi:
Currently kubemacpool webhook handlers patch the entire request object
(pod/virtualMachine). But sometimes there are changes made by other webhooks
or from internal kubernetes changes, that we don't want to set in our webhook.
In order to only add the wanted changes,
we change our webhook patch strategy to only patch the relevant changes.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/pod/pod.go
+++ b/pkg/webhook/pod/pod.go
@@ -22,6 +22,8 @@ import (
 	"net/http"
 
 	webhookserver "github.com/qinqon/kube-admission-webhook/pkg/webhook/server"
+	"gomodules.xyz/jsonpatch/v2"
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
@@ -54,6 +56,7 @@ func (a *podAnnotator) Handle(ctx context.Context, req admission.Request) admiss
 	if err != nil {
 		return admission.Errored(http.StatusBadRequest, err)
 	}
+	originalPod := pod.DeepCopy()
 
 	if pod.Annotations == nil {
 		pod.Annotations = map[string]string{}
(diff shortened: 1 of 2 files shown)


RamLavi:
/hold

let's merge this change only after we fixed the KMP issue causing test 2995 to be [flaky](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/689/pull-e2e-cluster-network-addons-operator-kubemacpool-functests/1336970017039716352) on CNAO tier1 tests, as I believe this change may reduce it's appearance.

pkg/webhook/pod/pod.go
```diff
@@ -65,13 +68,34 @@ func (a *podAnnotator) Handle(ctx context.Context, req admission.Request) admiss
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
 
-	marshaledPod, _ := json.Marshal(pod)
+	// admission.PatchResponse generates a Response containing patches.
+	return patchPodChanges(originalPod, pod)
+}
+
+// create jsonpatches only to changed caused by the kubemacpool webhook changes
+func patchPodChanges(originalPod, currentPod *corev1.Pod) admission.Response {
```
RamLavi, line 76: I think I can make a good reuse of `patchChange()` [function](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/259/files#diff-276495201a043c6a12276502c3cf1a1db02fd2c7c9f4323b20415c49b244e9dbR125), but not sure where I should place it...

pkg/webhook/pod/pod.go
```diff
@@ -65,13 +68,34 @@ func (a *podAnnotator) Handle(ctx context.Context, req admission.Request) admiss
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
 
-	marshaledPod, _ := json.Marshal(pod)
+	// admission.PatchResponse generates a Response containing patches.
+	return patchPodChanges(originalPod, pod)
```
phoracek, line 72: Since we know all the fields we intent to control, making diff of the whole objects seems redundant. Not only that, onl line 66 we change the pod in-place. AllocatePodMac is a black box and we just assume in patchPodChanges that those fields are the only ones that can change. This PR does the job, but it extends the existing code, even though part of it may not be needed anymore.

In order to make the code-base maintainable, could you consider changing AllocatePodMac to it does not mutate the pod? e.g. it could return a map of interfaces and their assigned MACs. We can generate the jsonPatch based on this information alone. That would limit the function interface to minimum, make it possible to test it and judge the behavior based on the function signature alone.

phoracek (changes requested):

pkg/webhook/pod/pod.go
RamLavi: Can you elaborate exactly what you suggest to change? Do you mean only on pod's handler or also vm handler? 

In pods handler we only change the annotation, but in virtualmachine handler we add the interface's macAddress and also add finalizer. to common denominator is of the last 2 is the vm spec itself. Alternatively I can return a struct of all the changes (interfaces & finalizers) but I don't think it would make it more maintainable.
That's why instead of returning only the changes, and constructing the patches from them, I preferred in `patchPodChanges` and `patchVMChanges` to return the copy of the changed vm/pod and then extract exactly the needed changes from it.

phoracek: I would not even return the changes, just which MAC should be allocated to each VM. This way you'd be able to test the reading of the VM, figuring out whether we handle the allocation correctly.

The next function will get this map and apply it on a specific VM - return patches. I'm not sure I follow why'd you return finalizers. If MACs are allocated, you need a finalizer. If not, you don't need any. This can be decided by the second function.

I will not fight for this. Although I believe that having smaller functions with limited signatures would make the maintainability and testability easier.

RamLavi: I see what you mean, but changing the logic in the way you suggest needs more careful examination in order to unharm the current logic. Can we further discuss and do the change you suggest in another PR?

phoracek: Fair enough

phoracek (approved):
/lgtm
/approve

PR #260, opened 2020-12-24
webhook, Enhance logs to better track KMP manager

RamLavi:
This PR adds a several log improvements in order to better track the manager's operation.
Also it fixes a small issue in test_id:2995 where the interface length was wrongly sampled.

- **webhook, Add VM interfaces to create/update webhook logs**

- **webhook, Add RequestID to patchVMChanges**
currently patchVMChanges's logs are not printed with the generated RequestID
Added RequestID to better track the webhook logs

- **webhook, Add resourceVersion to create/update logs**

- **ci, Add test test VMs to prints during test failure**

- **ci, Correct interface removal check in test_id:2995**
After removal of br01 interface we check if the inteface number has reduced to 1
But since the vm object is not refreshed it doesn't really reflect the reality
and in fact will always be correct.
Added a call to re-get the vm instance in order to correct to test.

**Release note**:

```release-note
NONE
```


RamLavi:
PR no longer relevant. closing

PR #261, opened 2020-12-27
opt-out, Fix opt-out mode not updating configMap

RamLavi:
currently vm controller reconcile checks opt-in case.
in case of vm creation in opt-out mode the vmWaitConfigMap is not cleared as
part of normal vm creation, and after 10 minutes will be removed from cache.
Added opt-mode check in vm reconcile
Added test in opt-in/out to make sure vmWaitConfigMap is working well.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/virtualmachine/virtualmachine_controller.go
+++ b/pkg/controller/virtualmachine/virtualmachine_controller.go
@@ -100,20 +100,18 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 	}
 
 	if instance.ObjectMeta.DeletionTimestamp.IsZero() {
-		instanceOptedIn, err := r.poolManager.IsVmInstanceOptedIn(request.Namespace)
+		vmShouldBeManaged, err := r.poolManager.IsNamespaceManaged(instance.GetNamespace())
 		if err != nil {
-			return reconcile.Result{}, errors.Wrap(err, "failed to check opt-in selection for vm")
+			return reconcile.Result{}, errors.Wrap(err, "Failed to check if vm is managed")
 		}
-
-		if !instanceOptedIn {
-			logger.Info("vm is opted-out from kubemacpool")
+		if !vmShouldBeManaged {
+			logger.Info("vm is not managed by kubemacpool")
 			return reconcile.Result{}, nil
 		}
 
 		logger.V(1).Info("vm create/update event")
 		// The object is not being deleted, so we can set the macs to allocated
 		err = r.poolManager.MarkVMAsReady(instance, logger)
-
 		if err != nil {
 			return reconcile.Result{}, errors.Wrap(err, "Failed to reconcile kubemacpool after virtual machine's creation event")
 		}
--- a/pkg/pool-manager/pool.go
+++ b/pkg/pool-manager/pool.go
@@ -20,9 +20,11 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"reflect"
 	"sync"
 
 	"github.com/pkg/errors"
+	"k8s.io/api/admissionregistration/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
@@ -53,6 +55,13 @@ type PoolManager struct {
 	isKubevirt       bool                         // bool if kubevirt virtualmachine crd exist in the cluster
 	waitTime         int                          // Duration in second to free macs of allocated vms that failed to start.
 }
+
+type OptMode string
+
+const (
+	OptInMode  OptMode = "Opt-in"
+	OptOutMode OptMode = "Opt-out"
+)
 
 type AllocationStatus string
 
@@ -186,31 +195,100 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
 	return currentMac
 }
 
-// Checks if the namespace of an instance is opted in for kubemacpool
-func (p *PoolManager) isInstanceOptedIn(namespaceName, mutatingWebhookConfigName, webhookName string) (bool, error) {
+// isNamespaceSelectorCompatibleWithOptModeLabel decides whether a namespace should be managed
+// by comparing the mutating-webhook's namespaceSelector (that defines the opt-mode)
+// and its compatibility to the given label in the namespace
+func (p *PoolManager) isNamespaceSelectorCompatibleWithOptModeLabel(namespaceName, mutatingWebhookConfigName, webhookName string, vmOptMode OptMode) (bool, error) {
+	isNamespaceManaged, err := isNamespaceManagedByDefault(vmOptMode)
+	if err != nil {
+		return false, errors.Wrap(err, "Failed to check if namespaces are managed by default by opt-mode")
+	}
+
 	ns, err := p.kubeClient.CoreV1().Namespaces().Get(context.TODO(), namespaceName, metav1.GetOptions{})
 	if err != nil {
-		return false, errors.Wrapf(err, "Failed to get Namespace %s", namespaceName)
+		return false, errors.Wrap(err, "Failed to get Namespace")
 	}
 	namespaceLabelMap := ns.GetLabels()
-	log.V(3).Info("namespaceName Labels", "namespaceName", namespaceName, "Labels", namespaceLabelMap)
+	log.V(1).Info("namespaceName Labels", "vm instance namespaceName", namespaceName, "Labels", namespaceLabelMap)
 
-	mutatingWebhookConfiguration, err := p.kubeClient.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), mutatingWebhookConfigName, metav1.GetOptions{})
+	webhook, err := p.lookupWebhookInMutatingWebhookConfig(mutatingWebhookConfigName, webhookName)
 	if err != nil {
-		return false, errors.Wrapf(err, "Failed to get mutatingWebhookConfig %s", mutatingWebhookConfigName)
+		return false, errors.Wrap(err, "Failed lookup webhook in MutatingWebhookConfig")
 	}
-
-	for _, webhook := range mutatingWebhookConfiguration.Webhooks {
-		if webhook.Name == webhookName {
-			if namespaceLabelSet := labels.Set(namespaceLabelMap); namespaceLabelSet != nil {
-				if webhookNamespaceLabelSelector, err := metav1.LabelSelectorAsSelector(webhook.NamespaceSelector); err == nil && webhookNamespaceLabelSelector.Matches(namespaceLabelSet) {
-					return true, nil
-				}
-			}
+	if namespaceLabelSet := labels.Set(namespaceLabelMap); namespaceLabelSet != nil {
+		isNamespaceManaged, err = isNamespaceManagedByWebhookNamespaceSelector(webhook, vmOptMode, namespaceLabelSet, isNamespaceManaged)
+		if err != nil {
+			return false, errors.Wrap(err, "Failed to check if namespace managed by webhook namespaceSelector")
 		}
 	}
 
-	return false, nil
+	return isNamespaceManaged, nil
+}
+
+func (p *PoolManager) lookupWebhookInMutatingWebhookConfig(mutatingWebhookConfigName, webhookName string) (*v1beta1.MutatingWebhook, error) {
+	mutatingWebhookConfiguration, err := p.kubeClient.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), mutatingWebhookConfigName, metav1.GetOptions{})
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to get mutatingWebhookConfig")
+	}
+	for _, webhook := range mutatingWebhookConfiguration.Webhooks {
+		if webhook.Name == webhookName {
+			return &webhook, nil
+		}
+	}
+	return nil, fmt.Errorf("webhook %s was not found on mutatingWebhookConfig %s", webhookName, mutatingWebhookConfigName)
+}
+
+// isNamespaceManagedByDefault checks if namespaces are managed by default by opt-mode
+func isNamespaceManagedByDefault(vmOptMode OptMode) (bool, error) {
+	switch vmOptMode {
+	case OptInMode:
+		return false, nil
+	case OptOutMode:
+		return true, nil
+	default:
+		return false, fmt.Errorf("opt-mode is not defined: %s", vmOptMode)
+	}
+}
+
+// isNamespaceManagedByWebhookNamespaceSelector checks if namespace managed by webhook namespaceSelector and opt-mode
+func isNamespaceManagedByWebhookNamespaceSelector(webhook *v1beta1.MutatingWebhook, vmOptMode OptMode, namespaceLabelSet labels.Set, defaultIsManaged bool) (bool, error) {
+	webhookNamespaceLabelSelector, err := metav1.LabelSelectorAsSelector(webhook.NamespaceSelector)
+	if err != nil {
+		return false, errors.Wrapf(err, "Failed to set webhook Namespace Label Selector for webhook %s", webhook.Name)
+	}
+
+	isMatch := webhookNamespaceLabelSelector.Matches(namespaceLabelSet)
+	log.V(1).Info("webhook NamespaceLabelSelectors", "webhookName", webhook.Name, "NamespaceLabelSelectors", webhookNamespaceLabelSelector)
+	if vmOptMode == OptInMode && isMatch {
+		// if we are in opt-in mode then we check that the namespace has the including label
+		return true, nil
+	} else if vmOptMode == OptOutMode && !isMatch {
+		// if we are in opt-out mode then we check that the namespace does not have the excluding label
+		return false, nil
+	}
+	return defaultIsManaged, nil
+}
+
+// getOptMode returns the configured opt-mode
+func (p *PoolManager) getOptMode(mutatingWebhookConfigName, webhookName string) (OptMode, error) {
+	webhook, err := p.lookupWebhookInMutatingWebhookConfig(mutatingWebhookConfigName, webhookName)
+	if err != nil {
+		return "", errors.Wrap(err, "Failed lookup webhook in MutatingWebhookConfig")
+	}
+	for _, expression := range webhook.NamespaceSelector.MatchExpressions {
+		if reflect.DeepEqual(expression, metav1.LabelSelectorRequirement{Key: webhookName, Operator: "In", Values: []string{"allocate"}}) {
+			return OptInMode, nil
+		} else if reflect.DeepEqual(expression, metav1.LabelSelectorRequirement{Key: webhookName, Operator: "NotIn", Values: []string{"ignore"}}) {
+			return OptOutMode, nil
+		}
+	}
+
+	// opt-in can technically also be defined with matchLabels
+	if value, ok := webhook.NamespaceSelector.MatchLabels[webhookName]; ok && value == "allocate" {
+		return OptInMode, nil
+	}
+
+	return "", fmt.Errorf("No Opt mode defined for webhook %s in mutatingWebhookConfig %s", webhookName, mutatingWebhookConfigName)
 }
 
 func GetMacPoolSize(rangeStart, rangeEnd net.HardwareAddr) (int64, error) {
--- a/pkg/pool-manager/pool_test.go
+++ b/pkg/pool-manager/pool_test.go
@@ -29,6 +29,7 @@ import (
 
 	multus "github.com/intel/multus-cni/types"
 	"github.com/onsi/ginkgo/extensions/table"
+	"k8s.io/api/admissionregistration/v1beta1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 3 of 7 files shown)


pkg/pool-manager/pool.go
```diff
@@ -213,6 +221,32 @@ func (p *PoolManager) isInstanceOptedIn(namespaceName, mutatingWebhookConfigName
...
+					return OptInMode, nil
+				} else if reflect.DeepEqual(expression, metav1.LabelSelectorRequirement{Key: webhookName, Operator: "NotIn", Values: []string{"ignore"}}) {
+					return OptOutMode, nil
+				}
+			}
+
+			// opt-in can technically also be defined with matchLabels
+			if value, ok := webhook.NamespaceSelector.MatchLabels[webhookName]; ok && value == "allocate" {
+					return OptInMode, nil
+			}
```
ormergi, line 243: Are you going to drop one of the method's to check the opt mod?

tests/virtual_machines_test.go
```diff
@@ -532,6 +532,14 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
 					Expect(err).ToNot(HaveOccurred(), "should successfully parse assigned mac")
 				})
+				It("should remove the mac from vmWaitConfigMap after vm created successfully", func() {
+					By("Checking vmWaitConfigMap is cleared")
+					Eventually(func() map[string]string {
+						data, err := getVmWaitConfigMapData()
+						Expect(err).ToNot(HaveOccurred(), "Should succeed getting vmWaitConfigMap")
+						return data
+					}, timeout, pollingInterval).Should(BeEmpty(), "vmWaitConfigMap should be empty after successful vm creation")
+				})
```
ormergi, line 542: I think you need to move vm creation to `BeforeEach` for this to work

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -100,14 +100,21 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
...
-			logger.Info("vm is opted-out from kubemacpool")
-			return reconcile.Result{}, nil
+			if err != nil {
+				return reconcile.Result{}, errors.Wrap(err, "failed to check opt-in selection for vm")
+			}
+
+			if !instanceOptedIn {
+				logger.Info("vm is opted-out from kubemacpool")
+				return reconcile.Result{}, nil
+			}
```
ormergi, line 117: Hey could you please elaborate on that, how is the behavior is changed?

RamLavi: `IsVmInstanceOptedIn()` is currently used to exit the controller if the vm is not "opted-in", i.e. the namespace of the vm doesn't have the right label.

In Opt-out mode however, we want to opposite behavior, i.e. that the controller will **not** exit by default.

This basically means that `IsVmInstanceOptedIn()` needs to only be checked if we are indeed in `opt-in mode`. So I added a check for that

pkg/pool-manager/pool.go
RamLavi: I locally removed the code added in `virtualmachine_controller.go` and re-ran the tests to make sure the test the added tests fail. 
would you like me to do it on the PR so you can see that?

tests/virtual_machines_test.go
RamLavi: totally right.

pkg/pool-manager/pool.go
```diff
@@ -213,6 +221,32 @@ func (p *PoolManager) isInstanceOptedIn(namespaceName, mutatingWebhookConfigName
 	return false, nil
 }
 
+func (p *PoolManager) getOptMode(mutatingWebhookConfigName, webhookName string) (OptMode, error) {
```
alonSadan, line 224: maybe you can pass this function no arguments and define `mutatingWebhookConfigName` and `webhookName` as constants.   Then you could export this function (change to GetOptMode) and frop the wrapper `GetVmOptMode` in virtualmachine_pool.go

tests/virtual_machines_test.go
```diff
@@ -518,19 +518,35 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 			})
 
 			Context("and kubemacpool is opted-in on a namespace", func() {
+				var (
+					vm *kubevirtv1.VirtualMachine
+				)
 				BeforeEach(func() {
-					By("opting out the namespace")
```
alonSadan, line 522: also in the opt out namespace, if the namespace wasn't opt out, kubemacpool should allocate the mac right?
Maybe this check, is needed as well?

tests/virtual_machines_test.go
```diff
@@ -518,19 +518,35 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
-					_, err = net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
-					Expect(err).ToNot(HaveOccurred(), "should successfully parse assigned mac")
+						By("Create VM")
+						err := testClient.VirtClient.Create(context.TODO(), vm)
+						Expect(err).ToNot(HaveOccurred(), "should success creating the vm")
+					})
+					It("should create a VM object with a MAC assigned", func() {
+						_, err := net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
+						Expect(err).ToNot(HaveOccurred(), "should successfully parse assigned mac")
+					})
```
alonSadan, line 541: nit: maybe new-line here

tests/virtual_machines_test.go
```diff
@@ -570,6 +586,14 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 					err := testClient.VirtClient.Create(context.TODO(), vm)
 					Expect(err).ToNot(HaveOccurred(), "should success creating the vm")
 				})
+				It("should remove the mac from vmWaitConfigMap after vm created successfully", func() {
```
alonSadan, line 589: nit: maybe new-line here

alonSadan (commented):
Thanks Ram. Added some notes

pkg/pool-manager/pool.go
RamLavi: not sure I follow.
The reason I used a generic `getOptMode()`, is that there are currently 2 webhooks: for vms, and for pods.
In this PR I'm only addressing the vm Opt mode, but in the future we may adress the same for pods, so using a generic function seems fitting.

tests/virtual_machines_test.go
RamLavi: Adding a new line after `It()` is inconsistent with all the other tests in this module.
why add it here?

tests/virtual_machines_test.go
RamLavi: Adding a new line at the end of a test is inconsistent with all the other tests in this module.
can you specify why do you think a new line is needed?

tests/virtual_machines_test.go
RamLavi: the opt-x wording is confusing, so let me clarify
in opt-in-mode - opting a namespace in, means it's "in the game", i.e. controlled by KMP

in other words, by opting in the namespace (as in the this BeforeEach), I expect that:
1. the mac will be allocated
2.  the mac in configmap will eventually get removed
so I think this is right. do you agree?

tests/virtual_machines_test.go
alonSadan: right. my bad

tests/virtual_machines_test.go
alonSadan: right. my bad

pkg/pool-manager/pool.go
alonSadan: Does `mutatingWebhookConfigName` `webhookName string` will have different values on the pods webhook?
why those names have to passed as arguments?

tests/virtual_machines_test.go
alonSadan: Thanks for explaining. A lot clearer now. Fill free to ignore this note then. Sorry for the noise.

pkg/pool-manager/pool.go
RamLavi: yes, they have the same mutatingWebhookConfigName, but different webhookName

tests/virtual_machines_test.go
RamLavi: no no, I think I'll try to make it clearer thanks to you.

pkg/pool-manager/pool.go
alonSadan: got it. Thanks!

RamLavi:
Following @alonSadan 's comments, I've decided to do some refactoring to the PR. 
now it combines the behavior in all opt-mode cases.
Also added unit test.

pkg/pool-manager/pool.go
```diff
@@ -186,31 +194,78 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+		// in OptOutMode namespaces are managed by default
+		isNamespaceManaged = true
+	} else {
+		return false, fmt.Errorf("opt-mode is not defined: %s", vmOptMode)
+	}
+
 	ns, err := p.kubeClient.CoreV1().Namespaces().Get(context.TODO(), namespaceName, metav1.GetOptions{})
 	if err != nil {
-		return false, errors.Wrapf(err, "Failed to get Namespace %s", namespaceName)
+		return isNamespaceManaged, errors.Wrapf(err, "Failed to get Namespace %s", namespaceName)
```
EdDev, line 212: This patter is contravention is general: On failure, the caller is usually supposed to ignore the value (non error type ones). The method, usually returns the "init value" and attempts to express that it should not be examined.

In this case, it will be strange to see it returning `true` and error at the same time. It make me wander if the caller is suppose to examine it even on errors (which is again uncommon).

pkg/pool-manager/pool.go
```diff
@@ -186,31 +194,78 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
 				}
+
+				log.V(1).Info("webhook NamespaceLabelSelectors", "webhookName", webhookName, "NamespaceLabelSelectors", webhookNamespaceLabelSelector)
+				if vmOptMode == OptInMode && webhookNamespaceLabelSelector.Matches(namespaceLabelSet) {
+					// if we are in opt-in mode then we check that the namespace has the including label
+					isNamespaceManaged = true
+				} else if vmOptMode == OptOutMode && !webhookNamespaceLabelSelector.Matches(namespaceLabelSet) {
+					// if we are in opt-out mode then we check that the namespace does not have the excluding label
+					isNamespaceManaged = false
+				}
```
EdDev, line 237: This looks like a different check for the namespace-manahed. If before it was checked based on the  opt-mode defaults, not it is checked against the label selector.
So, it will be nice to extract it to its own function, naming it per what it does so it will be easy to read without entering the low details.

In addition, the boolean is updated here on every webhook, each one overwriting the previous. Was this the intention (it is not "any" nor "all" operation)?
Maybe you depend on entering this area only once per the webhook name?

pkg/pool-manager/pool.go
```diff
@@ -186,31 +194,78 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+}
+
+func (p *PoolManager) getOptMode(mutatingWebhookConfigName, webhookName string) (OptMode, error) {
+	mutatingWebhookConfiguration, err := p.kubeClient.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), mutatingWebhookConfigName, metav1.GetOptions{})
+	if err != nil {
+		return "", errors.Wrapf(err, "Failed to get mutatingWebhookConfig %s", mutatingWebhookConfigName)
+	}
+
+	for _, webhook := range mutatingWebhookConfiguration.Webhooks {
+		if webhook.Name == webhookName {
```
EdDev, line 252: Can there be multiple webhooks with the same name?

pkg/pool-manager/pool_test.go
```diff
@@ -589,4 +590,182 @@ var _ = Describe("Pool", func() {
...
+			},
+		}
+		var poolManager *PoolManager
+		BeforeEach(func() {
+			poolManager = createPoolManager("02:00:00:00:00:00", "02:00:00:00:00:01", optOutMutatingWebhookConfiguration, optInMutatingWebhookConfiguration, namespaceWithIncludingLabel, namespaceWithExcludingLabel, namespaceWithNoLabels, namespaceWithIrrelevantLabels)
+		})
+		table.DescribeTable("Should return the expected namespace acceptance outcome according to the opt-mode",
+			func(n *compareNamespaceSelectorToOptModeParams) {
+				isNamespaceManaged, err := poolManager.compareNamespaceSelectorToOptMode(n.namespaceName, n.mutatingWebhookConfigurationName, webhookName, n.optMode)
+				Expect(err).ToNot(HaveOccurred(), "compareNamespaceSelectorToOptMode should not return error")
```
EdDev, line 702: There are multiple exit points with error. I think it is not checked anywhere.

tests/virtual_machines_test.go
```diff
@@ -518,19 +518,35 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 			})
 
 			Context("and kubemacpool is opted-in on a namespace", func() {
+				var (
+					vm *kubevirtv1.VirtualMachine
+				)
```
EdDev, line 523: nit: Why 3 lines?

tests/virtual_machines_test.go
```diff
@@ -518,19 +518,35 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
 				BeforeEach(func() {
-					By("opting out the namespace")
+					By("opting in the namespace")
 					err := addLabelsToNamespace(TestNamespace, map[string]string{vmNamespaceOptInLabel: "allocate"})
```
EdDev, line 526: nit: Checking `Succeed` is nicer.

tests/virtual_machines_test.go
```diff
@@ -518,19 +518,35 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
-					Expect(err).ToNot(HaveOccurred(), "Should succeed creating the vm")
-					_, err = net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
-					Expect(err).ToNot(HaveOccurred(), "should successfully parse assigned mac")
+						By("Create VM")
+						err := testClient.VirtClient.Create(context.TODO(), vm)
+						Expect(err).ToNot(HaveOccurred(), "should success creating the vm")
+					})
+					It("should create a VM object with a MAC assigned", func() {
+						_, err := net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
+						Expect(err).ToNot(HaveOccurred(), "should successfully parse assigned mac")
```
EdDev, line 540: nit: Here is a shorter version...
`Expect(net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)).NotTo(BeEmpty(), "...")`

In case of an error, it will explode.

EdDev (commented):
In addition the the inline comments, I noticed the usage of errors wrap.
Two general comments about it, not strongly related to this PR:

1. Consider using the Go stdlib instead of the external one.
2. Using `Wrap` exposes inner errors to the outer layers. It means that you explicitly want to maintain these errors as formal API/s from the packages that throw them.
If you do that, you a. need to document these errors OR b. assert on them in tests.
If there are no checks on these errors, there is little reasoning to wrap them.

pkg/pool-manager/pool.go
```diff
@@ -186,31 +194,78 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
 	return currentMac
 }
 
-// Checks if the namespace of an instance is opted in for kubemacpool
-func (p *PoolManager) isInstanceOptedIn(namespaceName, mutatingWebhookConfigName, webhookName string) (bool, error) {
+// Checks if the namespace of an instance needs to managed by kubemacpool
```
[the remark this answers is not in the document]
RamLavi: Done

pkg/pool-manager/pool.go
```diff
@@ -186,31 +194,78 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+	var isNamespaceManaged bool
+	if vmOptMode == OptInMode {
+		// in OptInMode namespaces are not managed by default
+		isNamespaceManaged = false
+	} else if vmOptMode == OptOutMode {
+		// in OptOutMode namespaces are managed by default
+		isNamespaceManaged = true
+	} else {
+		return false, fmt.Errorf("opt-mode is not defined: %s", vmOptMode)
+	}
```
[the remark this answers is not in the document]
RamLavi: Done

pkg/pool-manager/pool.go
RamLavi: I agree. In case of failure I would prefer to not handle the namespace.

pkg/pool-manager/pool.go
RamLavi: > This looks like a different check for the namespace-manahed. If before it was checked based on the opt-mode defaults, not it is checked against the label selector.
> So, it will be nice to extract it to its own function, naming it per what it does so it will be easy to read without entering the low details.

I agree

> 
> In addition, the boolean is updated here on every webhook, each one overwriting the previous. Was this the intention (it is not "any" nor "all" operation)?
> Maybe you depend on entering this area only once per the webhook name?

IIUC the boolean will only be updated if it matches the name, which is unique, so it should only happen once (if any).

pkg/pool-manager/pool.go
RamLavi: my first instinct was to say no, but then I checked, and it's actually technically possible, as there is no validating-webhook in the mutatingWebhookConfiguration instance that checks for duplicate names in the webhook list. So.. that's good to know =)
However, in our case it's still not possible: KMP's mutatingWebhookConfiguration is not meant for user configuration, and even if someone tries to change it manually, it will be reconciled back by the CNAO operator. So I think it's still valid to treat this field as unique and pick by it.

pkg/pool-manager/pool_test.go
RamLavi: you are right. 
Added the error scenarios.
Done

tests/virtual_machines_test.go
RamLavi: no particular reason. just keeping conformity with other [tests](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/master/tests/virtual_machines_test.go#L561) that defined vm instances.
We can of course decide to change it, but if so then let's do it in another PR to also change in the other tests together.

tests/virtual_machines_test.go
RamLavi: Like! but again this should be fixed in a broader scope for all tests.. will issue a PR

tests/virtual_machines_test.go
RamLavi: the comment is not in the correct place (I guess github is to blame) but I guess you meant this line:
```
Expect(err).ToNot(HaveOccurred(), "should success creating the vm")
```

fixed. Done

RamLavi:
@EdDev review fixes [here](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/30c72407dd47edd40a2dbd619b150c1a7db06439..dba452c233abd27c72226c8908acd85bee82744b)

RamLavi:
> In addition the the inline comments, I noticed the usage of errors wrap.
> Two general comments about it, not strongly related to this PR:
> 
> 1. Consider using the Go stdlib instead of the external one.
> 2. Using `Wrap` exposes inner errors to the outer layers. It means that you explicitly want to maintain these errors as formal API/s from the packages that throw them.
>    If you do that, you a. need to document these errors OR b. assert on them in tests.
>    If there are no checks on these errors, there is little reasoning to wrap them.

Although `github.com/pkg/errors` is an external library I'm a big advocate for it, as it makes the debugging job much easier. Indeed, whenever an error is thrown by KMP manager I try to document it in the tests by using gomega's error matchers.

pkg/pool-manager/pool.go
```diff
@@ -186,31 +195,97 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+	log.V(1).Info("namespaceName Labels", "vm instance namespaceName", namespaceName, "Labels", namespaceLabelMap)
 
 	mutatingWebhookConfiguration, err := p.kubeClient.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Get(context.TODO(), mutatingWebhookConfigName, metav1.GetOptions{})
 	if err != nil {
-		return false, errors.Wrapf(err, "Failed to get mutatingWebhookConfig %s", mutatingWebhookConfigName)
+		return false, errors.Wrap(err, "Failed to get mutatingWebhookConfig")
 	}
 
 	for _, webhook := range mutatingWebhookConfiguration.Webhooks {
 		if webhook.Name == webhookName {
```
EdDev, line 220: Based on the previous answer to my question, I understand this is expected to occur only once, therefore once it is matched there is no point to go over the rest of the webhooks. (I think this is what originally happens)

It will also be nice to express that this name is expected only once by splitting the lookup from the operation on it:
```
webhook = lookupWebhook(webhookName) // replacing the for-loop
if namespaceLabelSet := labels.Set(namespaceLabelMap); namespaceLabelSet != nil {
    return isNamespaceManagedByWebhookNamespaceSelector(webhook, vmOptMode, namespaceLabelSet, isNamespaceManaged)
}
```

pkg/pool-manager/pool.go
EdDev: See previous comment, you can use a lookup function here to express there is only one webhook name.
This is the 2nd place you can use it.

pkg/pool-manager/pool_test.go
```diff
@@ -589,4 +590,223 @@ var _ = Describe("Pool", func() {
 			})
 		})
 	})
+
+	type compareNamespaceSelectorToOptModeParams struct {
+		mutatingWebhookConfigurationName string
+		namespaceName                    string
+		optMode                          OptMode
+		ErrorTextExpected                string
```
EdDev, line 598: I would not insist on this, but IMO it is wrong to assert on error text.
You should assert on the actual error object. The implication is that the errors are either pre-defined using `New` or you use a specific type check that is exposed.
Then the assertion is against a specific instance (if you exposed the error using New) or against a specific type (in case you expose the error type).

tests/virtual_machines_test.go
EdDev: No idea how it landed on this line indeed. But it was meant for all occurrences where the return value is only an error, which there are a few of those at the moment.

Then you can do `Expect(foo()).To(Succeed())`

pkg/pool-manager/pool.go
```diff
@@ -186,31 +195,97 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
 	return currentMac
 }
 
-// Checks if the namespace of an instance is opted in for kubemacpool
-func (p *PoolManager) isInstanceOptedIn(namespaceName, mutatingWebhookConfigName, webhookName string) (bool, error) {
+// compareNamespaceSelectorToOptMode decides whether a namespace should be managed
+// by comparing the mutating-webhook's namespaceSelector (that defines the opt-mode)
+// and its compatibility to the given label in the namespace
+func (p *PoolManager) compareNamespaceSelectorToOptMode(namespaceName, mutatingWebhookConfigName, webhookName string, vmOptMode OptMode) (bool, error) {
```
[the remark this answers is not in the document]
RamLavi: totally. I'm bad with naming, `isNamespaceSelectorCompatibleWithOptModeLabel` is better imo

pkg/controller/virtualmachine/virtualmachine_controller.go
```diff
@@ -100,14 +100,9 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
...
 	if instance.ObjectMeta.DeletionTimestamp.IsZero() {
-		instanceOptedIn, err := r.poolManager.IsVmInstanceOptedIn(request.Namespace)
-		if err != nil {
-			return reconcile.Result{}, errors.Wrap(err, "failed to check opt-in selection for vm")
-		}
-
-		if !instanceOptedIn {
-			logger.Info("vm is opted-out from kubemacpool")
-			return reconcile.Result{}, nil
+		vmShouldBeManaged, err := r.poolManager.IsNamespaceManaged(instance.GetNamespace())
```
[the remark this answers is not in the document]
RamLavi: Done

pkg/pool-manager/pool.go
RamLavi: I agree. Done

pkg/pool-manager/pool.go
```diff
@@ -186,31 +195,97 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
 
-	return false, nil
+	return isNamespaceManaged, nil
+}
+
+// isNamespaceManagedByDefault checks if namespaces are managed by default by opt-mode
+func isNamespaceManagedByDefault(vmOptMode OptMode) (bool, error) {
+	switch vmOptMode {
+	case OptInMode:
+		// in OptInMode namespaces are not managed by default
```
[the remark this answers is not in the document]
RamLavi: Done

pkg/pool-manager/pool.go
```diff
@@ -186,31 +195,97 @@ func getNextMac(currentMac net.HardwareAddr) net.HardwareAddr {
...
+
+// isNamespaceManagedByWebhookNamespaceSelector checks if namespace managed by webhook namespaceSelector and opt-mode
+func isNamespaceManagedByWebhookNamespaceSelector(webhook v1beta1.MutatingWebhook, vmOptMode OptMode, namespaceLabelSet labels.Set, defaultIsManaged bool) (bool, error) {
+	webhookNamespaceLabelSelector, err := metav1.LabelSelectorAsSelector(webhook.NamespaceSelector)
+	if err != nil {
+		return false, errors.Wrapf(err, "Failed to set webhook Namespace Label Selector for webhook %s", webhook.Name)
+	}
+
+	log.V(1).Info("webhook NamespaceLabelSelectors", "webhookName", webhook.Name, "NamespaceLabelSelectors", webhookNamespaceLabelSelector)
+	if vmOptMode == OptInMode && webhookNamespaceLabelSelector.Matches(namespaceLabelSet) {
```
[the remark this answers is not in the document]
RamLavi: Done

pkg/pool-manager/pool.go
RamLavi: Done

tests/virtual_machines_test.go
RamLavi: ack. this is true for all the tests, will do in a separate PR

pkg/pool-manager/pool_test.go
RamLavi: I think this method does not go hand in hand with the `errors.wrap()` method, where all the errors are concatenated with `:` between them. That's why I prefer to match the error text.

I changed the test to use gomega's matchError that supports error text matching.

RamLavi:
@EdDev Fixes2 [here](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/df2f08cfa33f3b85accf9ebfc60f962ec52cff61..dcacf46c84d84f1c04c4633ff8c8641bb80374f4)

pkg/pool-manager/pool_test.go
EdDev: It is the opposite, the `Wrap` is exposing the instance and type object of the inner error, not the text. Text exposure does not need to use wrapping, a simple `%v` does the job.
You can read about it [here](https://blog.golang.org/go1.13-errors).

EdDev (approved):
Thanks!

EdDev:
Trying again.
/lgtm

Issue #264, opened 2021-01-21
Add kubevirt.io/v1 apiVersion to the webhook rules


coboluxx:
**What happened**:
kubemacpool does not provide a mac to kubevirt machiens deployed with apiVersion kubevirt.io/v1

**What you expected to happen**:

mac address assigned.

**How to reproduce it (as minimally and precisely as possible)**:

Create a new machine with with the apiVersion kubevirt.io/v1 (kubevirt v0.37.0)

**Anything else we need to know?**:

The webhook rules need an update to include v1

```yaml
rules:
  - apiGroups:
    - kubevirt.io
    apiVersions:
    - v1alpha3
    operations:
    - CREATE
    - UPDATE
    resources:
    - virtualmachines
```

RamLavi:
Hey @coboluxx, thanks for bringing this to my attention :) .
I'm currently waiting for [v0.37.1](v0.37.1) to be released, as it adds a [patch](https://github.com/kubevirt/kubevirt/pull/4842) needed in order successfully add kubevirt schema in KMP.

RamLavi:
@coboluxx please see if you are able to create virtualMachines with kubevirt/v1.
If all is well, I will close issue

PR #265, opened 2021-01-27
Bump kubevirt 0.37.1

RamLavi:
This PR bumps kubevirt to 0.37.1 and updates code and manifests to kubevirt/v1

PR Solves  #264 

**Release note**:

```release-note
bump kubevirt apiVersion to v1
```


pkg/manager/manager.go
```diff
@@ -92,6 +92,7 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 			return errors.Wrap(err, "unable to set up manager")
 		}
 
+		os.Setenv("KUBEVIRT_CLIENT_GO_SCHEME_REGISTRATION_VERSION", "v1")
```
davidvossel, line 95: you'll need to set this environment variable before executing the binary. So set it on the pod's container spec, or using an entry point shell script. 

setting it within the binary is too late as the init function that interprets the env var will have already executed.

RamLavi: I see, thanks!

config/test/kubemacpool.yaml
```diff
@@ -71,7 +71,7 @@ webhooks:
   - apiGroups:
     - kubevirt.io
     apiVersions:
-    - v1alpha3
+    - v1
```
qinqon, line 74: This add a hard dep on kubevirt >= 0.37 maybe we should document it ?

qinqon (changes requested):

qinqon:
@RamLavi can you fix DCO ? 

/lgtm
/approve

RamLavi:
> @RamLavi can you fix DCO ?

fixed. can you re-lgtm?

PR #267, opened 2021-01-31
Refactor vm tests

RamLavi:
This PR cleans the vm tests a little bit.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -103,16 +103,14 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 						vm := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br", "")}, []kubevirtv1.Network{newNetwork("br")})
 						err := testClient.VirtClient.Create(context.TODO(), vm)
 						Expect(err).ToNot(HaveOccurred())
-						_, err = net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)
-						Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


ormergi (approved):

pkg/pool-manager/pool_test.go
```diff
@@ -51,13 +51,13 @@ var _ = Describe("Pool", func() {
 	createPoolManager := func(startMacAddr, endMacAddr string, fakeObjectsForClient ...runtime.Object) *PoolManager {
 		fakeClient := fake.NewSimpleClientset(fakeObjectsForClient...)
 		startPoolRangeEnv, err := net.ParseMAC(startMacAddr)
-		Expect(err).ToNot(HaveOccurred(), "should successfully parse starting mac address range")
+		Expect(err).To(Succeed(), "should successfully parse starting mac address range")
```
qinqon, line 54: Humm, feels better to use "Success" when you pass the cmd to the Expect in cases when there is just the error returned, it will be read like "this command shoud succeed"
```golang
Expect( net.ParseMAC(startMacAddr)).To(Succeed(), "should successfully parse starting mac address range")
```

But in the case that we store the error at an "err" variable feels better to talk about error ocurring, since an error can occurr or don't occurr but but they don't succeed

qinqon (changes requested):
Let's use the "Succeed" version only when we pass the command to the Expect.

pkg/pool-manager/pool_test.go
RamLavi: I agree. will remove.

qinqon: I mean that if the function return just error then we can use Succeed otherwise we use HaveOcurred
```golang
func Foo() error...
func Bar() string, error...
Expect(Foo()).To(Succeed())
bar, err := Bar()
Expect(err).ToNot(HaveOcurred())
```

qinqon:
/lgtm
/approve

Let's use implicit error handling.

PR #276, opened 2021-02-24
consider opt-mode when inserting vms/pod from the cluster to macPoolMap

RamLavi:
usually Kubemacpool handles only managed pods/vms according to the opt-mode. 
However there are corner cases (for example when initializing the macPoolMap for pods/vms) where we don't consider if the pods/vms are managed.
This PR addresses this issue to only handle managed instances.

- **opt-mode, generalize pods and vms to use the same opt-mode check**
Currently pod and vm handlers use slightly different functions to
determine if the pod/vm namespace is handled by kubemacpool according to
the opt-mode.
Aligned them to use the same function

- **cluster vm sweep, fix to only use managed vms**
When kubemacpool performs cluster sweeps for virtual machines (during
macPoolMap initalization for example), it needs to only consider the vms
from managed namespaces.
Moved vm macPoolMap initialization to a function initMacMapFromCluster
Fixed forEachVmInterfaceInClusterRunFunction function to only go over
managed vms.
Added e2e tests since unit test currently don't support vm instances

- **pod-init, init macPoolMap only insert managed pods**
currently during MacPoolMap initialization of pods, all pods are taken
into account,
Fixed to only use pods on managed namespaces.
Introduced unit tests with opt-out modes for pods and vms in order to
correctly initialize the macPoolMap.
Added pod macPoolMap init unit test

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/pod/pod_controller.go
+++ b/pkg/controller/pod/pod_controller.go
@@ -80,12 +80,12 @@ func (r *ReconcilePolicy) Reconcile(request reconcile.Request) (reconcile.Result
 	logger := log.WithName("Reconcile").WithValues("podName", request.Name, "podNamespace", request.Namespace)
 
 	logger.V(1).Info("got a pod event in the controller")
-	instanceOptedIn, err := r.poolManager.IsPodInstanceOptedIn(request.Namespace)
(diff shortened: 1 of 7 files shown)


qinqon (approved):
All good

PR #277, opened 2021-03-24
webhook, Include metadata changes to kubemacpool TS update scenario

RamLavi:
This PR adds more scenarios where the kuemacpool TS annotation is updated, 
but also making sue we do not cause an update loop.
Also, it fixes a problem with the way annotations where patched in the webhook handlers.

- **webhook, vm, Include metadata changes to kubemacpool TS update scenario**
Currently, only vm spec changes in the cause the kubemacpool
TransactionTimestampAnnotation to be updated. In order to optimize
updating of the mac to be faster, we also include metadata changes like
Annotation and namespace updates.
We do not include the TransactionTimestampAnnotation itself when
checking for changes, to avoid an internal update loop.

- **webhook, Fix Patching of annotations**
Currently update of the annotation patch causes other existing
annotations to disappear, initiating a update loop with kubevirt trying
to add his annotation.
To avoid this, we move to patch the entire annotation field, similar to
how it is done in kubevirt[[1]](https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-api/webhooks/mutating-webhook/mutators/migration-create-mutator.go#L68).

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/pod/pod.go
+++ b/pkg/webhook/pod/pod.go
@@ -75,12 +75,12 @@ func (a *podAnnotator) Handle(ctx context.Context, req admission.Request) admiss
 
 // create jsonpatches only to changed caused by the kubemacpool webhook changes
 func patchPodChanges(originalPod, currentPod *corev1.Pod) admission.Response {
-	var kubemapcoolJsonPatches []jsonpatch.Operation
+	kubemapcoolJsonPatches := []jsonpatch.Operation{}
 
 	currentNetworkAnnotation := currentPod.GetAnnotations()[pool_manager.NetworksAnnotation]
 	originalPodNetworkAnnotation := originalPod.GetAnnotations()[pool_manager.NetworksAnnotation]
 	if originalPodNetworkAnnotation != currentNetworkAnnotation {
-		annotationPatch := jsonpatch.NewPatch("add", "/metadata/annotations", map[string]string{pool_manager.NetworksAnnotation: currentNetworkAnnotation})
+		annotationPatch := jsonpatch.NewPatch("replace", "/metadata/annotations", currentPod.GetAnnotations())
 		kubemapcoolJsonPatches = append(kubemapcoolJsonPatches, annotationPatch)
 	}
 
--- a/pkg/webhook/virtualmachine/virtualmachine.go
+++ b/pkg/webhook/virtualmachine/virtualmachine.go
@@ -98,13 +98,13 @@ func (a *virtualMachineAnnotator) Handle(ctx context.Context, req admission.Requ
 // create jsonpatches only to changed caused by the kubemacpool webhook changes
 func patchVMChanges(originalVirtualMachine, currentVirtualMachine *kubevirt.VirtualMachine, parentLogger logr.Logger) admission.Response {
 	logger := parentLogger.WithName("patchVMChanges")
-	var kubemapcoolJsonPatches []jsonpatch.Operation
+	kubemapcoolJsonPatches := []jsonpatch.Operation{}
 
 	if !pool_manager.IsVirtualMachineDeletionInProgress(currentVirtualMachine) {
 		originalTransactionTSString := originalVirtualMachine.GetAnnotations()[pool_manager.TransactionTimestampAnnotation]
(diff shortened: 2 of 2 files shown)


pkg/webhook/pod/pod.go
```diff
@@ -75,12 +75,12 @@ func (a *podAnnotator) Handle(ctx context.Context, req admission.Request) admiss
 
 // create jsonpatches only to changed caused by the kubemacpool webhook changes
 func patchPodChanges(originalPod, currentPod *corev1.Pod) admission.Response {
-	var kubemapcoolJsonPatches []jsonpatch.Operation
+	kubemapcoolJsonPatches := []jsonpatch.Operation{}
 
 	currentNetworkAnnotation := currentPod.GetAnnotations()[pool_manager.NetworksAnnotation]
 	originalPodNetworkAnnotation := originalPod.GetAnnotations()[pool_manager.NetworksAnnotation]
 	if originalPodNetworkAnnotation != currentNetworkAnnotation {
-		annotationPatch := jsonpatch.NewPatch("add", "/metadata/annotations", map[string]string{pool_manager.NetworksAnnotation: currentNetworkAnnotation})
```
ormergi, line 83: I think it override other annotations due to missing `/-` at the patch path section:
```
jsonpatch.NewPatch("add", "/metadata/annotations", ...)
jsonpatch.NewPatch("add", "/metadata/annotations/-", ...)
                                                ^^^
```

pkg/webhook/virtualmachine/virtualmachine.go
```diff
@@ -204,13 +204,46 @@ func (a *virtualMachineAnnotator) mutateUpdateVirtualMachinesFn(virtualMachine *
...
+
+// isVirtualMachineChanged checks if the vm spec changed in this webhook update request.
 func isVirtualMachineSpecChanged(previousVirtualMachine, virtualMachine *kubevirt.VirtualMachine) bool {
 	return !reflect.DeepEqual(previousVirtualMachine.Spec, virtualMachine.Spec)
 }
 
+// isVirtualMachineMetadataChanged checks if non-automatically generated metadata fields changed in this webhook
+// update request.
+func isVirtualMachineMetadataChanged(previousVirtualMachine, virtualMachine *kubevirt.VirtualMachine) bool {
+	if previousVirtualMachine.GetNamespace() != virtualMachine.GetNamespace() {
```
ormergi, line 225: Correct me if I am wrong, if the exact same VM will be create on different namespace (using the same yaml file but create it on a different namespace) it will be recognized as VM to update right?
I think we should teat such VM as another VM in the cluster.

pkg/webhook/pod/pod.go
RamLavi: interesting, will try that!

pkg/webhook/virtualmachine/virtualmachine.go
rhrazdil: Speaking of namespace, how could a namespace of a VM change? AFAIK namespace is not mutable for any k8s resource

RamLavi: Not sure I follow.

RamLavi: ahh I think I get you ... a valid point. I'll remove this check then

RamLavi: Done

rhrazdil:
/lgtm
Adding hold so @ormergi can re-visit
/hold

ormergi:
Thanks

/lgtm
/hold

pkg/webhook/pod/pod.go
RamLavi: @ormergi I tried using that locally, but am getting:
```
Error from server: error when creating "myvm_2iface_no_ns.yaml": admission webhook "virtualmachine-validator.kubevirt.io" denied the request: json: cannot unmarshal object into Go struct field ObjectMeta.metadata.annotations of type string
```

so I guess we'll do it the kubevirt way

PR #279, opened 2021-03-25
Move cert-manager execution to a new pod

qinqon:
Currectly the cert-manager from kube-admission-webhook is running
whithing the webhook server itself, this means that we need to create a
placeholder secret and also a very specific readiness probe has to be
used to ensure that secret is properly rotated at start up. This change
introduces a new pod runing the cert-manager code so we don't need to
create the placeholder secret (the webhook pod will not start until the
secret is not created) and we don't need a readiness probess since
secret will be correct form the beginning.

Also start up is faster since webhook pods start at the very moment the
secret is created by cert-manager pod.

**Release note**:

```release-note
Move cert manager code to a different pod
```


config/default/manager/manager.yaml
```diff
@@ -107,26 +99,68 @@ spec:
...
+        imagePullPolicy: Always
+        name: manager
+        env:
+          - name: RUN_CERT_MANAGER
+            value: ""
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: POD_NAME
```
RamLavi, line 151: will this have the same POD_NAME as KMP manager?

config/test/kubemacpool.yaml
```diff
@@ -229,6 +219,59 @@ spec:
...
+  template:
+    metadata:
+      labels:
+        app: kubemacpool
+        control-plane: cert-manager
+        controller-tools.k8s.io: "1.0"
+    spec:
+      containers:
+      - args:
+        - --v=production
```
RamLavi, line 245: I think you should add an entry to the [kustomize](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/master/config/test/kustomization.yaml#L9) file in tests folder so that this will be changed to `debug` like the `kubemacpool-mac-controller-manager` pod.

RamLavi (changes requested):

RamLavi:
in ci, do we want to also restart the cert-pod in `restartKubemacpoolManagerPods`?

qinqon:
[force-push](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/692e12022fedadddbfb5705795641d131239e775..e1db75cfaaa4c1ad22c3f7b425b59f22c84900c4): Forgot to call `make generate`

config/default/manager/manager.yaml
qinqon: no, the manifests has different names.

qinqon:
Addressed @RamLavi comments and added a readiness probe since certs are ok but we have to be sure that HTTP server is up and running.

cmd/manager/main.go
```diff
@@ -62,69 +65,101 @@ func main() {
...
+			os.Exit(1)
+		}
+
+		log.Info("starting manager")
+		if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
+			log.Error(err, "problem running manager")
+			os.Exit(1)
+		}
+
+	} else {
```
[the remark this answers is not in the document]
qinqon: Done

phoracek:
> Addressed @RamLavi comments and added a readiness probe since certs are ok but we have to be sure that HTTP server is up and running.

Do we need to add the readiness probe to nmstate too?

qinqon:
> > Addressed @RamLavi comments and added a readiness probe since certs are ok but we have to be sure that HTTP server is up and running.
> 
> Do we need to add the readiness probe to nmstate too?

Yep :-/

config/test/kubemacpool.yaml
qinqon: Done

qinqon:
> in ci, do we want to also restart the cert-pod in restartKubemacpoolManagerPods?
> Can you also add a test to check scenarios where the cert-pod is deleted?

If we restart the cert-manager at `restartKubemacpoolManagerPods` we will be testing cert-pod deletion too right ?

RamLavi:
> If we restart the cert-manager at `restartKubemacpoolManagerPods` we will be testing cert-pod deletion too right ?

I think so it would suffice, unless there is some other scenario worth testing you can think of.

RamLavi (approved):
/lgtm
@qinqon feel free to approve if you feel there are no test scenarios to add

PR #282, opened 2021-04-05
Reduce wait-time to 5 min

RamLavi:
This PR reduces kubemacpool's wait-time to 5 minutes, and adjusts tests accordingly to not be affected by webhook rejects.

- **reduce wait-time to 5 minutes**
currently if a webhook change is rejected then after wait-time
the kubemacpool goroutine will refresh the mac-entry after wait-time
according to what is set in the vm instance.
Reducing this value to 5 minutes, to allow quicker response time.

- **ci, extend test eventually timeout**
in tests that check webhook changes, we extend the timeout to ensure that
if the last webhook change was rejected, the mac entry will eventually get
refreshed with the persisted value.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/config/default/manager/manager.yaml
+++ b/config/default/manager/manager.yaml
@@ -63,7 +63,7 @@ spec:
         - /manager
         args:
           - "--v=production"
-          - "--wait-time=600"
+          - "--wait-time=300"
         image: quay.io/kubevirt/kubemacpool:latest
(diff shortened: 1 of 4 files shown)


qinqon:
I think we are ok with 5 min 
/Lgtm
/Approve

PR #284, opened 2021-04-06
ci, add deployment print upon changeReplicas failure

RamLavi:
This PR adds a print of deployment in case of timeout of change replicas 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 	"reflect"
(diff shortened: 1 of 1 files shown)


tests/tests.go
```diff
@@ -286,22 +292,25 @@ func changeReplicas(managerName string, numOfReplica int32) error {
...
+		}
 
 		if managerDeployment.Status.ReadyReplicas != numOfReplica {
 			return false
 		}
 
 		return true
-
-	}, 2*time.Minute, 3*time.Second).Should(BeTrue(), "failed to change kubemacpool deployment number of replicas")
+	}, 2*time.Minute, 3*time.Second).Should(BeTrue(), fmt.Sprintf("failed to change kubemacpool deployment number of replicas.\n deployment:\n%v", string(indentedDeployment)))
```
qinqon, line 313: To/Should cláusula at Ginkgo works like a Sprintf you can just pass the fmt string and the expansion values.

qinqon (changes requested):

tests/tests.go
RamLavi: cool! thanks. Done

PR #285, opened 2021-04-12
Remove "app" from matchingLabels

qinqon:
When Deployments are upgraded they cannot change matchingLabels since
they are inmutable, this change remove the problematic label from
matchingLabels.

**Release note**:

```release-note
NONE
```


RamLavi (changes requested):
don't we want to also add a unique label for cert-pod?

qinqon:
> don't we want to also add a unique label for cert-pod?

We have already "control-plane: cert-manager", the problem was that after adding "commonLabel" to kustomize.yaml it also add that to deployment labelMatcher, so just removing that is enough.

PR #286, opened 2021-04-19
Delete pods to restart KMP

qinqon:
Changing Deployment replicas to restart KMP does not work well with CNAO
tier1 tests since CNAO will Reconcile the Deployment after changing the
replicas so it's not always working. This change replace the restart
mechanism with KMP pods deletiong and deployment condition check.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -21,6 +21,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


tests/tests.go
```diff
@@ -315,6 +306,40 @@ func changeReplicas(managerName string, numOfReplica int32) error {
...
+				if condition.Type == v1.DeploymentAvailable {
+					return condition.Status == conditionStatus
+				}
+			}
+
+			return true
+		}, 2*time.Minute, time.Second/10).Should(BeTrue(), "Failed waiting readiness at for deployment:\n%v", string(deploymentName))
+	}
+
+	deploymentConditionAvailability(corev1.ConditionFalse, 2*time.Minute, time.Second/10)
```
RamLavi, line 338: What if the deployment is ready before getting to this line?

RamLavi (commented):
I'm concerned that now that KMP pods are ready faster, that polling the deployment for readiness will result with flakiness...

tests/tests.go
qinqon: restarting pods takes enouhg time to reach this, or this is what I have see on multiple runs

I can go back to check replicas but we can face the same problem and enmascarade a false Readiness.

RamLavi: How do you feel about waiting for the "old" pods to be cleared by specific-name, before changing the deployment replicas back to 1?

qinqon: Ahh that's a good one :-) will do thanks!

tests/tests.go
```diff
@@ -315,6 +306,40 @@ func changeReplicas(managerName string, numOfReplica int32) error {
 	return nil
 }
 
+func deletePodsFromDeployment(deploymentName string) error {
```
RamLavi, line 309: maybe `restartPodsFromDeployment` is more exact here?

tests/tests.go
qinqon: Done

tests/tests.go
qinqon: Done

RamLavi (approved):
/lgtm
/approve

PR #292, opened 2021-04-29
fix relaod if controller-manager.Start() returns

RamLavi:
This PR fixes a bug where if kuevirt is ready after KMP, then KMP will not relaod and as a result the VM controller will not be initiated.

- ***fix kubemacpool relaod if controller-manager returns***
currently kubemacpool is waiting on signals in order to reload the
manager, simulating a kubemacpool restart.
An example is if kubevirt is not ready by the time kubemacpol is up.
In the example - when kubevirt is ready, a channel is closed to signal
kubemacpool to restart.
This commit fixes the waitForSignal method to use the new context based
cancel function.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -43,12 +43,13 @@ type KubeMacPoolManager struct {
 	config                   *rest.Config
 	metricsAddr              string
 	continueToRunManager     bool
-	kubevirtInstalledChannel chan struct{}   // This channel is close after we found kubevirt to reload the manager
-	stopSignalChannel        chan os.Signal  // stop channel signal
-	podNamespace             string          // manager pod namespace
-	podName                  string          // manager pod name
-	waitingTime              int             // Duration in second to free macs of allocated vms that failed to start.
-	runtimeManager           manager.Manager // Delegated controller-runtime manager
+	kubevirtInstalledChannel chan struct{}      // This channel is close after we found kubevirt to reload the manager
+	stopSignalChannel        chan os.Signal     // stop channel signal
+	cancel                   context.CancelFunc // cancel() closes the controller-runtime context internal channel
+	podNamespace             string             // manager pod namespace
+	podName                  string             // manager pod name
+	waitingTime              int                // Duration in second to free macs of allocated vms that failed to start.
+	runtimeManager           manager.Manager    // Delegated controller-runtime manager
 }
 
 func NewKubeMacPoolManager(podNamespace, podName, metricsAddr string, waitingTime int) *KubeMacPoolManager {
@@ -97,6 +98,8 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 			return errors.Wrap(err, "unable to create pool manager")
 		}
 
+		var ctx context.Context
+		ctx, k.cancel = context.WithCancel(context.Background())
 		if !isKubevirtInstalled {
 			log.Info("kubevirt was not found in the cluster start a watching process")
 			go k.waitForKubevirt()
@@ -120,7 +123,7 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 			return errors.Wrap(err, "failed to start pool manager routines")
 		}
 
-		err = k.runtimeManager.Start(context.Background())
+		err = k.runtimeManager.Start(ctx)
 		if err != nil {
 			log.Error(err, "unable to run the manager")
 		}
@@ -164,6 +167,8 @@ func (k *KubeMacPoolManager) waitForKubevirt() {
 
 // wait for the any interrupt to stop the manager and clean the webhook.
 func (k *KubeMacPoolManager) waitForSignal() {
+	defer k.cancel()
+
 	select {
 	// This channel is a system interrupt this will stop the container
 	case <-k.stopSignalChannel:


pkg/manager/manager.go
```diff
@@ -174,6 +177,7 @@ func (k *KubeMacPoolManager) waitForSignal() {
 	case <-k.kubevirtInstalledChannel:
 		log.Info("found kubevirt restarting the manager")
 	}
+	k.cancel()
```
qinqon, line 180: can we defer this at the begging of the function ?

rhrazdil (commented):
Hello Ram, I'm just going through the loop trying to understand the whole process.

rhrazdil:
> Hello Ram, I'm just going through the loop trying to understand the whole process.

duh... Submitted by mistake, I can't comment on lines that are not being changed, so I'll try to formulate
my questions this way. Don't take this as a review really and feel free to respond whenever convenient.


```
		if !isKubevirtInstalled {
			log.Info("kubevirt was not found in the cluster start a watching process")
			go k.waitForKubevirt()
		}
		go k.waitForSignal()
```
so here we spawn two goroutines, `k.waitForKubevirt` and `k.waitForSignal`. 
IIUC `waitForSignal` blocks until either `k.stopSignalChannel`  or  `k.kubevirtInstalledChannel` receive
and `waitForKubevirt` runs for 5 seconds and then finishes.

In the meantime the main thread continues all the way to 
```
		// restart channels
		k.kubevirtInstalledChannel = make(chan struct{})
```

Shouldn't we also wait for `waitForKubevirt` to finish?

Unless `k.runtimeManager.Start(ctx)` does somethiing I don't know abount, it seems to me that
`waitForKubevirt` goroutine could be running multiple times

RamLavi:
> > Hello Ram, I'm just going through the loop trying to understand the whole process.
> 
> duh... Submitted by mistake, I can't comment on lines that are not being changed, so I'll try to formulate
> my questions this way. Don't take this as a review really and feel free to respond whenever convenient.
> 
> ```
> 		if !isKubevirtInstalled {
> 			log.Info("kubevirt was not found in the cluster start a watching process")
> 			go k.waitForKubevirt()
> 		}
> 		go k.waitForSignal()
> ```
> 
> so here we spawn two goroutines, `k.waitForKubevirt` and `k.waitForSignal`.
> IIUC `waitForSignal` blocks until either `k.stopSignalChannel` or `k.kubevirtInstalledChannel` receive
> and `waitForKubevirt` runs for 5 seconds and then finishes.

it actually does not block at all (perhaps a change of name is in order, as I misunderstood it the same way the first time :) ). 
but anyway - it just waits for kubevirt, and then signals the pod to "restart" - this is for both VM & pod InitiMacPools and init controllers to start properly.

> 
> In the meantime the main thread continues all the way to
> 
> ```
> 		// restart channels
> 		k.kubevirtInstalledChannel = make(chan struct{})
> ```
> 
> Shouldn't we also wait for `waitForKubevirt` to finish?

Actually k.runtimeManager.Start(ctx) is "blocking", in the sense that most of KMP's life we spend there.
we only reach the `// restart channels` if we force the `k.cancel()` method or something went wrong during the run of controller-runtime.

> 
> Unless `k.runtimeManager.Start(ctx)` does somethiing I don't know abount, it seems to me that
> `waitForKubevirt` goroutine could be running multiple times

IIUC, since after `waitForKubevirt` finds kubevirt, we "Restart" KMP, we shouldn't be running the `waitForKubevirt` goroutine multiple times.. Well at least not after this PR (that's what it essentially fixes).

RamLavi:
log that shows the restart of the main loop when kubevirt is found:
```
2021-05-02T11:23:13.344Z	DEBUG	manager	kubevirt exist in the cluster	{"kubevirtExist": true}
2021-05-02T11:23:13.344Z	INFO	manager	found kubevirt restarting the manager
2021-05-02T11:23:13.344Z	INFO	controller-runtime.webhook	shutting down webhook server
2021-05-02T11:23:13.344Z	INFO	controller-runtime.manager.controller.pod-controller	Stopping workers
2021-05-02T11:23:13.844Z	INFO	manager	Setting up Manager
I0502 11:23:15.795356       1 request.go:655] Throttling request took 1.049079896s, request: GET:https://10.96.0.1:443/apis/networking.k8s.io/v1beta1?timeout=32s
2021-05-02T11:23:16.955Z	INFO	controller-runtime.metrics	metrics server is starting to listen	{"addr": ":8080"}
2021-05-02T11:23:16.993Z	INFO	manager	Setting up controllers
2021-05-02T11:23:16.993Z	INFO	controller-runtime.injectors-warning	Injectors are deprecated, and will be removed in v0.10.x
2021-05-02T11:23:16.993Z	INFO	controller-runtime.injectors-warning	Injectors are deprecated, and will be removed in v0.10.x
2021-05-02T11:23:16.993Z	INFO	controller-runtime.injectors-warning	Injectors are deprecated, and will be removed in v0.10.x
2021-05-02T11:23:16.993Z	INFO	controller-runtime.injectors-warning	Injectors are deprecated, and will be removed in v0.10.x
2021-05-02T11:23:16.993Z	INFO	manager	Setting up webhooks

```

rhrazdil:
Hello @RamLavi, Thanks for explaining! 

> Actually k.runtimeManager.Start(ctx) is "blocking", in the sense that most of KMP's life we spend there.
we only reach the // restart channels if we force the k.cancel() method or something went wrong during the run of controller-runtime.

Oh right now I see how it works. Now using the context makes sense to me.
Regarding the context, there's one thing that occurs to me.
Culd it possibly happen that `waitForKubevirt` finds kubevirt and sends signal to `waitForSignal`, which calls `k.cancel()`
**before**
```
ctx, k.cancel = context.WithCancel(context.Background())
``` 
is reached? IIUC, that would cause panic at runtime

pkg/manager/manager.go
```diff
@@ -164,6 +167,8 @@ func (k *KubeMacPoolManager) waitForKubevirt() {
 
 // wait for the any interrupt to stop the manager and clean the webhook.
 func (k *KubeMacPoolManager) waitForSignal() {
+	defer k.cancel()
```
oshoval, line 170: waitForSignal will be alive until one of the conditions is met isn't it ?
(because its a select without a default case)
If so, this code is prone to create some dangling go routines
maybe better to close them ?
since the loop 
https://github.com/k8snetworkplumbingwg/kubemacpool/pull/292/files#diff-d500fbd6a2aa620607ca5e2a7c3ac4f1a4c82309d1a549561e92abfcb18f2f0eR84
creates them

sorry if i am wrong

pkg/manager/manager.go
```diff
@@ -120,7 +123,7 @@ func (k *KubeMacPoolManager) Run(rangeStart, rangeEnd net.HardwareAddr) error {
 			return errors.Wrap(err, "failed to start pool manager routines")
 		}
 
-		err = k.runtimeManager.Start(context.Background())
+		err = k.runtimeManager.Start(ctx)
```
oshoval, line 126: is it fine that there is no return on line 128  (right column of the numbers) ?

general speaking - maybe better to explicitly close channels for example on line 132
when creating a new one

RamLavi:
> Hello @RamLavi, Thanks for explaining!
> 
> > Actually k.runtimeManager.Start(ctx) is "blocking", in the sense that most of KMP's life we spend there.
> > we only reach the // restart channels if we force the k.cancel() method or something went wrong during the run of controller-runtime.
> 
> Oh right now I see how it works. Now using the context makes sense to me.
> Regarding the context, there's one thing that occurs to me.
> Culd it possibly happen that `waitForKubevirt` finds kubevirt and sends signal to `waitForSignal`, which calls `k.cancel()`
> **before**
> 
> ```
> ctx, k.cancel = context.WithCancel(context.Background())
> ```
> 
> is reached? IIUC, that would cause panic at runtime

I don't think this could happen now (it could with last week's changes, but I already fixed it yesterday).
see, `ctx, k.cancel = context.WithCancel(context.Background())` is called before `go k.waitForKubevirt()`, and once kubevirt is deployed, we will not get into `waitForKubevirt` again, even if kubemacpool gets reset (assuming that kubevirt is not crashing as well, and then kubemacpool is the least of the cluster's concerns :) )

pkg/manager/manager.go
RamLavi: I didn't find any such things. This is my reasoning:
- `waitForSignal` is supposed to be alive as long as kubemacpool is properly working, so normally there's nothing dangling. 

that leaves the two cases:
- In the case of kubevirt is pending then found by `checkForKubevirt`, then the `waitForKubevirt` goroutine will signal `waitForSignal` and close first, then `waitForSignal` goroutine will close, initiating the defer that will only then initiate the restart of the loop, and the goroutines will be recreated.. didn't find dangling goroutines here.. 

- the other case is that KMP gets a stop signal interrupt. in this scenario, assuming that kubevirt is already up, we only have `waitForSignal` running. then the signal notified to `waitForSignal` which breaks the loop and closes. didn't find dangling goroutines here as well.. 

IIUC, in the remote chance that both cases happen simultaniously, then we might have a dangling `waitForKubevirt` goroutine, but that's highly unlikely.. 

what do you think?

pkg/manager/manager.go
RamLavi: when KMP runtimeManager fails, it does a "soft" restart, where we simply return to the runtimeManager loop. This is why if `k.runtimeManager.Start(ctx)` returns, we don't return and simply reloop (if it would return, the the main function will finish, and the pod will be restarted by the deployment, which is more like a "hard" reset.

As to explicitly close channels.. I was under the impression that closing channels is the way channels signal. so not sure closing them is the right way to handle them.. @qinqon is there a better way to handle this?

rhrazdil:
> I don't think this could happen now (it could with last week's changes, but I already fixed it yesterday).
see, ctx, k.cancel = context.WithCancel(context.Background()) is called before go k.waitForKubevirt(), and once kubevirt is deployed, we will not get into waitForKubevirt again, even if kubemacpool gets reset (assuming that kubevirt is not crashing as well, and then kubemacpool is the least of the cluster's concerns :) )

Ah! I missed this [change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/a9eb53cf20d57d3e145cfbcd81abe83bc78e11b4..ef6aa33440b36e0f0a309d6a8b7eda1d5ced47c8), this is exactly what I had in mind.

oshoval:
```
		if !isKubevirtInstalled {
			log.Info("kubevirt was not found in the cluster start a watching process")
			go k.waitForKubevirt()
		}
		go k.waitForSignal()
```

Aint a race can happen ? 
since the first go routine signals, and the 2nd wait for it
the signal might be missing ?

pkg/manager/manager.go
oshoval: Sorry i am not familiar with it enough
better that we discuss offline if possible, meanwhile lets not block the PR
thanks

rhrazdil: I think Or has a point in case `Start` returns an error. IIUC, the [docs](https://github.com/kubernetes-sigs/controller-runtime/blob/0d84ce43073cefb813563dd7586c943fa2c97783/pkg/manager/manager.go#L86) say that Start returns in two cases:
- an error occured
- context was canceled

If `Start` returns an error, then the goroutines are still running. If `Start` returns error in each iteration for some reason, then I think we may be spawning new goroutines in each iteration.
But this behaviour is not introduced by this PR

RamLavi:
> ```
> 		if !isKubevirtInstalled {
> 			log.Info("kubevirt was not found in the cluster start a watching process")
> 			go k.waitForKubevirt()
> 		}
> 		go k.waitForSignal()
> ```
> 
> Aint a race can happen ?
> since the first go routine signals, and the 2nd wait for it
> the signal might be missing ?

since the channel is already created, it would just wait for it

> unless it will be catch by the next iteration or something, which will work, but maybe better to fix it on a follow PR.
> I am not into the details, so maybe i miss stuff.

no, next loop run we would have different channels.

> 
> EDIT, maybe since the event is in the channel, it will be fine, because it can happen even if the select
> started afterwards

I think you're right in this case.

pkg/manager/manager.go
RamLavi: ahh I see what you mean.. I will open an issue about it, to fix later, as indeed it's not introduced in this PR

Thanks @oshoval @rhrazdil for clarifying :)

RamLavi: https://github.com/k8snetworkplumbingwg/kubemacpool/issues/293

rhrazdil:
Thank you Ram for being so diligent with responses to all my questions!
/lgtm

PR #299, opened 2021-05-13
WIP Add component label template

oshoval:
In order to allow CNAO to add component label
when bumping KMP version, add label template
to the KMP kustomization manifest.

```release-note
None
```


oshoval:
/cc @RamLavi 

even if its WIP, lets please ok to test it and run sanity ?

oshoval:
we will update the kustomization file via CNAO instead, since its affect also the KMP itself, and shouldnt

PR #300, opened 2021-05-13
initVirtualMachineMap, Limit vmList request to avoid memory max limit

RamLavi:
limit the memory consumption when kubemacpool initiates the vm mac map
by limiting the request to chunks of 100 vms.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/pool-manager/virtualmachine_pool.go
+++ b/pkg/pool-manager/virtualmachine_pool.go
@@ -270,61 +270,87 @@ func (p *PoolManager) initMacMapFromCluster(parentLogger logr.Logger) error {
 	return nil
 }
 
-// forEachManagedVmInterfaceInClusterRunFunction gets all the macs from all the supported interfaces in all the managed cluster vms, and runs
-// a function f on it
-func (p *PoolManager) forEachManagedVmInterfaceInClusterRunFunction(f func(vmFullName string, iface kubevirt.Interface, networks map[string]kubevirt.Network) error) error {
-	logger := log.WithName("forEachManagedVmInterfaceInClusterRunFunction")
-	var result = p.kubeClient.ExtensionsV1beta1().RESTClient().Get().RequestURI("apis/kubevirt.io/v1/virtualmachines").Do(context.TODO())
-	if result.Error() != nil {
-		return result.Error()
-	}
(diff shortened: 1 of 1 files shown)


RamLavi:
@qinqon should we backport to 4.8 stable branch? I'm leaning to yes

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -270,61 +270,86 @@ func (p *PoolManager) initMacMapFromCluster(parentLogger logr.Logger) error {
...
-	err := result.Into(vms)
-	if err != nil {
-		return err
-	}
+// paginateVmsWithLimit performs a vm list request with pagination, to limit the amount of vms received at a time
+// and prevent taking too much memory.
+func (p *PoolManager) paginateVmsWithLimit(limit int64, vmsFunc func(pods *kubevirt.VirtualMachineList) error) error {
+	continueFlag := ""
+	for {
+		var result = p.kubeClient.ExtensionsV1beta1().RESTClient().Get().RequestURI(fmt.Sprintf("apis/kubevirt.io/v1/virtualmachines?limit=%v&continue=%v", limit, continueFlag)).Do(context.TODO())
```
qinqon, line 278: This feels a little fragile, we don't have the limit/continue option at kubevirt client right ?

qinqon (commented):
Just one question

pkg/pool-manager/virtualmachine_pool.go
RamLavi: This is the external api support [way](https://kubernetes.io/docs/reference/using-api/api-concepts/#retrieving-large-results-sets-in-chunks)..
I think once we'll incorporate the kubevirt client way then we'll be able to use the ListOptions ...

qinqon: Ack, can we add a `//FIXME` note about it ?

RamLavi: totally, [Done](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/1064582b3be4f87a570056d764f1af7099fc7536..5bccedba8228960095a047dc893efbf87b2dc465)

PR #312, opened 2021-06-06
finalizer, remove kubemacpool finalizer

RamLavi:
we no longer need to keep the vm instance until kubemacpool has freed
the vm macs from the pool since we now store the vm-name in the macpool
mac entry.
Hence, we should remove the finalizer from the vm creation.
For backwards compatibility, we still remove the finalizer if it exists
when deleting the vm.

**Release note**:

```release-note
NONE
```


RamLavi:
/hold
I still need to debug and see what causes the fails..

PR #322, opened 2021-08-05
metrics, Add kubevirt_kmp_duplicate_macs gauge

oshoval:
kubevirt_kmp_duplicate_macs counts how many duplicate macs
were found upon kubemacpool init.

Additional changes:
Improve log in case of Mac Address duplicates in order to ease mitigation.

```release-note
None
```

### DIFF
--- a/pkg/pool-manager/virtualmachine_pool.go
+++ b/pkg/pool-manager/virtualmachine_pool.go
@@ -32,6 +32,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	kubevirt "kubevirt.io/client-go/api/v1"
+
+	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/gauges"
 )
 
 func (p *PoolManager) AllocateVirtualMachineMac(virtualMachine *kubevirt.VirtualMachine, transactionTimestamp *time.Time, parentLogger logr.Logger) error {
@@ -259,6 +261,9 @@ func (p *PoolManager) initMacMapFromCluster(parentLogger logr.Logger) error {
 
 		if iface.MacAddress != "" {
 			if err := p.allocateRequestedVirtualMachineInterfaceMac(vmFullName, iface, parentLogger); err != nil {
+				if strings.Contains(err.Error(), "failed to allocate requested mac address") {
+					gauges.DuplicateMacGauge.Inc()
+				}
 				// Dont return an error here if we can't allocate a mac for a configured vm
 				parentLogger.Error(err, "Invalid/Duplicate mac address for virtual machine",
 					"virtualMachineFullName", vmFullName,
--- a/pkg/pool-manager/virtualmachine_pool.go
+++ b/pkg/pool-manager/virtualmachine_pool.go
@@ -212,7 +212,8 @@ func (p *PoolManager) allocateRequestedVirtualMachineInterfaceMac(vmFullName str
 	if macEntry, exist := p.macPoolMap[requestedMac]; exist {
 		if !macAlreadyBelongsToVmAndInterface(vmFullName, iface.Name, macEntry) {
 			err := fmt.Errorf("failed to allocate requested mac address")
-			logger.Error(err, "mac address already allocated")
+			logger.Error(err, fmt.Sprintf("mac address %s already allocated to %s, %s, conflict with: %s, %s",
+				iface.MacAddress, macEntry.instanceName, macEntry.macInstanceKey, vmFullName, iface.Name))
 
 			return err
 		}
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -11,12 +11,18 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 	pool_manager "github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager"
+	"github.com/k8snetworkplumbingwg/kubemacpool/tests/kubectl"
+
 	kubevirtv1 "kubevirt.io/client-go/api/v1"
 )
 
@@ -270,6 +276,7 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 					})
 				})
 			})
+
 			Context("and when restarting kubeMacPool and trying to create a VM with the same manually configured MAC as an older VM", func() {
 				It("[test_id:2179]should return an error because the MAC address is taken by the older VM", func() {
 					vm1 := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br1", "02:00:ff:ff:ff:ff")}, []kubevirtv1.Network{newNetwork("br1")})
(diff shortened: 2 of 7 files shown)


oshoval:
@RamLavi 
its still WIP, but i need ok-to-test please
want to run once all the tests and then will make it draft

oshoval:
/cc @RamLavi 
still testing it e2e with CRC (smoke test passed), but basically ready for review, thanks

Related
https://github.com/kubevirt/monitoring/pull/23
https://github.com/kubevirt/cluster-network-addons-operator/pull/952

tests/virtual_machines_test.go
```diff
@@ -270,6 +275,42 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 					})
 				})
 			})
+
+			Context("and creating two VMs with same mac, one on opt-out namespace", func() {
```
RamLavi, line 279: This test is running under `Context("When Running with default opt-mode configuration"` 
right now that means it's opt-out, but we can't rely on that. 
Since your tests is running assuming that we're running in opt-out, you need to move it under `Context("When all the VMs are serviced by default (opt-out mode)"`

tests/virtual_machines_test.go
```diff
@@ -11,12 +11,17 @@ import (
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/k8snetworkplumbingwg/kubemacpool/pkg/names"
 	pool_manager "github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager"
+	"github.com/k8snetworkplumbingwg/kubemacpool/tests/kubectl"
 	kubevirtv1 "kubevirt.io/client-go/api/v1"
```
RamLavi, line 25: kubevirt should be in a separate import group (org imports)

tests/virtual_machines_test.go
```diff
@@ -270,6 +275,42 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+
+					err = addLabelsToNamespace(OtherTestNamespace, map[string]string{vmNamespaceOptInLabel: "ignore"})
+					Expect(err).ToNot(HaveOccurred(), "should be able to add the namespace labels")
+
+					vm2 := CreateVmObject(OtherTestNamespace, false, []kubevirtv1.Interface{newInterface("br1", macAddress)}, []kubevirtv1.Network{newNetwork("br1")})
+					err = testClient.VirtClient.Create(context.TODO(), vm2)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(net.ParseMAC(vm2.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)).ToNot(BeEmpty(), "Should successfully parse mac address")
+
+					err = addLabelsToNamespace(OtherTestNamespace, map[string]string{vmNamespaceOptInLabel: "allocate"})
```
RamLavi, line 296: this is not needed in opt-out mode, where all vms are managed by default
in order to turn OtherTestNamespace from unmanaged to managed, you need to remove the "ignore" label

tests/virtual_machines_test.go
```diff
@@ -270,6 +275,42 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+					Expect(err).ToNot(HaveOccurred())
+					Expect(net.ParseMAC(vm1.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)).ToNot(BeEmpty(), "Should successfully parse mac address")
+
+					err = addLabelsToNamespace(OtherTestNamespace, map[string]string{vmNamespaceOptInLabel: "ignore"})
+					Expect(err).ToNot(HaveOccurred(), "should be able to add the namespace labels")
+
+					vm2 := CreateVmObject(OtherTestNamespace, false, []kubevirtv1.Interface{newInterface("br1", macAddress)}, []kubevirtv1.Network{newNetwork("br1")})
+					err = testClient.VirtClient.Create(context.TODO(), vm2)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(net.ParseMAC(vm2.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)).ToNot(BeEmpty(), "Should successfully parse mac address")
```
RamLavi, line 294: up until this line (including) - should be in BeforeEach(), as it sets the initial environment of the test, and not the test itself..

RamLavi (changes requested):
First pass

RamLavi:
I also think we need to add a service to expose kubemacpool as an endpoint.

oshoval:
> I also think we need to add a service to expose kubemacpool as an endpoint.
> when CNAO deployed it it'll add the serviceMonitor label to it as well.

I checked and there is no need to,
if we add it, it will have the metrics twice (check for example go_threads for the kmp pod and you will see it appear twice if you add a service),
just the container port was needed

oshoval:
Addressed comments

tests/virtual_machines_test.go
```diff
@@ -879,6 +886,43 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+					macAddress := "02:00:ff:ff:ff:ff"
+
+					vm1 := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br1", macAddress)}, []kubevirtv1.Network{newNetwork("br1")})
+					err := testClient.VirtClient.Create(context.TODO(), vm1)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(net.ParseMAC(vm1.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)).ToNot(BeEmpty(), "Should successfully parse mac address")
+
+					err = addLabelsToNamespace(OtherTestNamespace, map[string]string{vmNamespaceOptInLabel: "ignore"})
+					Expect(err).ToNot(HaveOccurred(), "should be able to add the namespace labels")
+
```
RamLavi, line 901: add `By("Adding a vm with a same mac Address %s on the unmanaged namespace")`

tests/virtual_machines_test.go
```diff
@@ -879,6 +886,43 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+					vm2 := CreateVmObject(OtherTestNamespace, false, []kubevirtv1.Interface{newInterface("br1", macAddress)}, []kubevirtv1.Network{newNetwork("br1")})
+					err = testClient.VirtClient.Create(context.TODO(), vm2)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(net.ParseMAC(vm2.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)).ToNot(BeEmpty(), "Should successfully parse mac address")
+				})
+
+				It("should detect duplicate macs gauge after cleaning the opt-out label and restarting kubemacpool", func() {
+					err := cleanNamespaceLabels(OtherTestNamespace)
```
RamLavi, line 909: add `By("moving the unmanaged namespace to be managed")`

tests/virtual_machines_test.go
```diff
@@ -879,6 +886,43 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+					Expect(err).ToNot(HaveOccurred(), "should be able to remove the namespace labels")
+
```
RamLavi, line 911: add `By("restarting Kubemacpool to check if there are duplicates mac in the managed namespaces")`

RamLavi (changes requested):
some nits to make the test more readable

oshoval:
Addressed comments

RamLavi:
can we also test the mitigation part of the alert? 
suggestion:
```
Context("and creating two VMs with same mac, one on unmanaged (opted out) namespace", func() {
  Context("and moving the unmanaged namespace to be mananged, then restarting kubemacpool", func() {
    It("should detect duplicate macs gauge", func() {
    })
    Context("and mac duplication is remove then restarting kubemacpool", func() {
      It("should not detect duplicate macs gauge", func() {
      })
    })
  })
})
```

RamLavi:
/hold
until we branch to KMP stable branch

oshoval:
> can we also test the mitigation part of the alert?
> suggestion:
> 
> ```
> Context("and creating two VMs with same mac, one on unmanaged (opted out) namespace", func() {
>   Context("and moving the unmanaged namespace to be mananged, then restarting kubemacpool", func() {
>     It("should detect duplicate macs gauge", func() {
>     })
>     Context("and mac duplication is remove then restarting kubemacpool", func() {
>       It("should not detect duplicate macs gauge", func() {
>       })
>     })
>   })
> })
> ```

Thought about it as well before but reached the following:
We can but i don't think its needed, because the kmp is restarted, and if it was restarting with value different than 0, then there the test won't equal 1 after one conflict.
If you still think its needed i will add it.

RamLavi:
> > can we also test the mitigation part of the alert?
> > suggestion:
> > ```
> > Context("and creating two VMs with same mac, one on unmanaged (opted out) namespace", func() {
> >   Context("and moving the unmanaged namespace to be mananged, then restarting kubemacpool", func() {
> >     It("should detect duplicate macs gauge", func() {
> >     })
> >     Context("and mac duplication is remove then restarting kubemacpool", func() {
> >       It("should not detect duplicate macs gauge", func() {
> >       })
> >     })
> >   })
> > })
> > ```
> 
> Thought about it as well before but reached the following:
> We can but i don't think its needed, because the kmp is restarted, and if it was restarting with value different than 0, then there the test won't equal 1 after one conflict.
> If you still think its needed i will add it.

I think we should. this will also act as a documentation to how to mitigate such issues.

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -259,6 +261,9 @@ func (p *PoolManager) initMacMapFromCluster(parentLogger logr.Logger) error {
 
 		if iface.MacAddress != "" {
 			if err := p.allocateRequestedVirtualMachineInterfaceMac(vmFullName, iface, parentLogger); err != nil {
+				if strings.Contains(err.Error(), "failed to allocate requested mac address") {
+					gauges.DuplicateMacGauge.Inc()
```
RamLavi, line 265: question. if we have a duplicate mac in vms vm1,vm2, then won't KMP count 2 duplicates? (on on vm1 with vm2 and the other vm2 with vm1)

tests/virtual_machines_test.go
```diff
@@ -879,6 +886,75 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+					Expect(err).ToNot(HaveOccurred())
+					Expect(net.ParseMAC(conflictingVM.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)).ToNot(BeEmpty(), "Should successfully parse mac address")
+				})
+
+				It("should detect duplicate macs gauge after cleaning the opt-out label and restarting kubemacpool", func() {
+					By("moving the unmanaged namespace to be managed")
+					err := cleanNamespaceLabels(OtherTestNamespace)
+					Expect(err).ToNot(HaveOccurred(), "should be able to remove the namespace labels")
+
+					By("restarting Kubemacpool to check if there are duplicates mac in the managed namespaces")
```
RamLavi, line 916:
```suggestion
					By("restarting Kubemacpool to check if there are duplicate macs in the managed namespaces")
```

tests/virtual_machines_test.go
```diff
@@ -879,6 +886,75 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+					err = initKubemacpoolParams()
```
RamLavi, line 917: This part: 
```
By("moving the unmanaged namespace to be managed")
					err := cleanNamespaceLabels(OtherTestNamespace)
					Expect(err).ToNot(HaveOccurred(), "should be able to remove the namespace labels")

					By("restarting Kubemacpool to check if there are duplicates mac in the managed namespaces")
					err = initKubemacpoolParams()
```
should be on BeforeEach, as it should be done for this and also for the next context

RamLavi (changes requested):

pkg/pool-manager/virtualmachine_pool.go
oshoval: The first one will enter the table without conflict, the 2nd won't enter and its the only one that will be counted

tests/virtual_machines_test.go
oshoval: But the 2nd context (and only the 2nd) must delete the VM before the initKubemacpoolParams
so just the clean i can move to the before each if needed
make sense ?

oshoval:
Addressed comments

oshoval:
Merged to one test

oshoval:
Thanks, add hold if you want please until stable is out (or its out already? and then we need approve ACK)

oshoval:
@RamLavi 
Thanks
can you please create a release so it will be part of CNAO ?

PR #326, opened 2021-08-19
dryRun, Fix logic conditions of dryRun

oshoval:
DryRun pointer is optional, so it might be nil.
Fix the condition to support this case.

```release-note
None
```

### DIFF
--- a/pkg/webhook/pod/pod.go
+++ b/pkg/webhook/pod/pod.go
@@ -60,7 +60,7 @@ func (a *podAnnotator) Handle(ctx context.Context, req admission.Request) admiss
 		pod.Annotations = map[string]string{}
 	}
 
-	if req.DryRun != nil && *req.DryRun != true {
+	if req.DryRun == nil || *req.DryRun == false {
 		transactionTimestamp := pool_manager.CreateTransactionTimestamp()
 		log.V(1).Info("got a create pod event", "podName", pod.Name, "podNamespace", pod.Namespace, "transactionTimestamp", transactionTimestamp)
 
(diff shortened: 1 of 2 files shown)


phoracek (commented):
Nice! Or, do I understand it correctly that this code did not work with dry run enabled? Would it make sense to write a test exposing that issue (if it existed)?

oshoval:
> Nice! Or, do I understand it correctly that this code did not work with dry run enabled? Would it make sense to write a test exposing that issue (if it existed)?

It worked with dryRun enabled but without supporting fully dryRun - this what i am fixing in other PR:
To support dryRun that does everything beside the persistent actions.

It didn't support the case where the dryRun field was nil (as its optional) - Tests I will cover on the dryRun fixing PR if its ok (this what was discussed with Ram), we might need old kubectl for it, will see.

oshoval:
can you please add ok-to-test ?

RamLavi:
/ok-to-test
/lgtm
I think this is good, I don't mind if you add tests for it in the PR that will deal with the dry-Run fix.

AFAIR the current tests already check the dry-Run=false option as they are the default value

oshoval:
While writing the test of https://github.com/k8snetworkplumbingwg/kubemacpool/pull/327
which fix dry run functionality, to include the whole response without creating side effects,
we saw that controller runtime supports only "All" CreationOption which act as server,
where the default is not dryRun of course, but even then the pointer is not nil (else it wouldn't work with current code).

I don't know how we can simulate the case where the pointer is nil,
and anyhow i prefer to not cause binding between the PRs, as they are not connected
If you have an idea how to test it please elaborate,
in anycase, the current tests will check the good flow as Ram said as well above, 
since the pointer will not be nil but the value is false, so the new condition will hit as well.

PR #327, opened 2021-08-19
dryRun, Adapt VM Mac Address allocation to support dryRun

oshoval:
In case of "server" ("All" when used by controller runtime) dryRun, 
we do want to return an updated server response, 
which allocates Mac Address, but without doing side effects.
Fix the flow to support it.
    
Note that the Mac pool head is changed, so calling twice dryRun
won't yield the same Mac Address in the response.
It is needed because in case there are multiple Mac Addresses in
the same request, we should not give the same Mac even on dryRun,
hence the mac pool head should be updated.

```release-note
None
```

### DIFF
--- a/pkg/pool-manager/pod_pool.go
+++ b/pkg/pool-manager/pod_pool.go
@@ -31,7 +31,7 @@ import (
 
 const tempPodName = "tempPodName"
 
-func (p *PoolManager) AllocatePodMac(pod *corev1.Pod) error {
+func (p *PoolManager) AllocatePodMac(pod *corev1.Pod, isNotDryRun bool) error {
 	p.poolMutex.Lock()
 	defer p.poolMutex.Unlock()
 
@@ -74,15 +74,15 @@ func (p *PoolManager) AllocatePodMac(pod *corev1.Pod) error {
 	networkList := []*multus.NetworkSelectionElement{}
 	for _, network := range networks {
 		if network.MacRequest != "" {
-			if err := p.allocatePodRequestedMac(network, pod); err != nil {
-				p.revertAllocationOnPod(podFullName, newAllocations)
+			if err := p.allocatePodRequestedMac(network, pod, isNotDryRun); err != nil {
+				p.revertAllocationOnPod(podFullName, newAllocations, isNotDryRun)
 				return err
 			}
 			newAllocations[network.Name] = network.MacRequest
 		} else {
-			macAddr, err := p.allocatePodFromPool(network, pod)
+			macAddr, err := p.allocatePodFromPool(network, pod, isNotDryRun)
 			if err != nil {
-				p.revertAllocationOnPod(podFullName, newAllocations)
+				p.revertAllocationOnPod(podFullName, newAllocations, isNotDryRun)
 				return err
 			}
 
@@ -129,7 +129,7 @@ func (p *PoolManager) ReleaseAllPodMacs(podFullName string) error {
 	return nil
 }
 
-func (p *PoolManager) allocatePodRequestedMac(network *multus.NetworkSelectionElement, pod *corev1.Pod) error {
+func (p *PoolManager) allocatePodRequestedMac(network *multus.NetworkSelectionElement, pod *corev1.Pod, isNotDryRun bool) error {
 	requestedMac := network.MacRequest
 
 	if _, err := net.ParseMAC(requestedMac); err != nil {
@@ -141,7 +141,9 @@ func (p *PoolManager) allocatePodRequestedMac(network *multus.NetworkSelectionEl
 		// the pod name may have not been updated in the webhook context yet,
 		// so we use a temp pod name and update it the the controller
 		podFullName = tempPodName
-		p.macPoolMap.createOrUpdateEntry(requestedMac, podFullName, network.Name)
+		if isNotDryRun {
+			p.macPoolMap.createOrUpdateEntry(requestedMac, podFullName, network.Name)
+		}
 		return nil
 	}
 	if macEntry, exist := p.macPoolMap[requestedMac]; exist {
@@ -153,8 +155,9 @@ func (p *PoolManager) allocatePodRequestedMac(network *multus.NetworkSelectionEl
 		}
 	}
 
-	p.macPoolMap.createOrUpdateEntry(requestedMac, podFullName, network.Name)
-
+	if isNotDryRun {
+		p.macPoolMap.createOrUpdateEntry(requestedMac, podFullName, network.Name)
+	}
 	log.Info("requested mac was allocated for pod",
 		"requestedMap", requestedMac,
 		"podFullName", podFullName)
@@ -162,7 +165,7 @@ func (p *PoolManager) allocatePodRequestedMac(network *multus.NetworkSelectionEl
 	return nil
 }
 
-func (p *PoolManager) allocatePodFromPool(network *multus.NetworkSelectionElement, pod *corev1.Pod) (string, error) {
+func (p *PoolManager) allocatePodFromPool(network *multus.NetworkSelectionElement, pod *corev1.Pod, isNotDryRun bool) (string, error) {
 	macAddr, err := p.getFreeMac()
 	if err != nil {
 		return "", err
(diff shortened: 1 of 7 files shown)


oshoval:
/cc @RamLavi 

ready for review
please add ok-to-test

tests/virtual_machines_test.go
```diff
@@ -100,6 +100,30 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 				}
 			})
 
+			Context("and the client tries to assign the same MAC address for two different vm, one created with dryRun=All mode", func() {
```
RamLavi, line 103:
```suggestion
			Context("and applying a dryRun vm create request", func() {
```

tests/virtual_machines_test.go
```diff
@@ -100,6 +100,30 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+				It("should success creating a vm with the same mac allocated to a dry run", func() {
+					By("Creating a vm with dry run mode")
```
RamLavi, line 105: will move the creation of the vm dryRun to `BeforeEach()`
The test should be about the pool allocation and not the dryRun creation

tests/virtual_machines_test.go
```diff
@@ -100,6 +100,30 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+					vm := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br", "")}, []kubevirtv1.Network{newNetwork("br")})
+					createOptions := &client.CreateOptions{}
+					client.DryRunAll.ApplyToCreate(createOptions)
+
+					err := testClient.VirtClient.Create(context.TODO(), vm, createOptions)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(net.ParseMAC(vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress)).ToNot(BeEmpty(), "Should successfully parse mac address")
+
+					vmDryRunOverlap := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("brDryRunOverlap", "")}, []kubevirtv1.Network{newNetwork("brDryRunOverlap")})
```
RamLavi, line 114: This should be the actual test, and we should add a `By("creating a vm with the same mac to make sure the dryRun assigned mac is not occupied on the macpool")`

tests/virtual_machines_test.go
```diff
@@ -100,6 +100,30 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+					macAddress := vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress
+					vmDryRunOverlap.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress = macAddress
```
RamLavi, line 116: you can assign the dryRun Mac in the `CreateVmObject()` itself
```
vmDryRunOverlap := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("brDryRunOverlap", macAddress)}, []kubevirtv1.Network{newNetwork("brDryRunOverlap")})
```

RamLavi (changes requested):

RamLavi:
>Note that the Mac pool head is changed, so calling twice dryRun
won't yield the same Mac Address in the response.
It is needed because in case there are multiple Mac Addresses in
the same request, we should not give the same Mac even on dryRun,
hence the mac pool head should be updated.

I agree with this line of thought. repeated dry Run requests will not issue the same MAC-Addresses, as the pool moves on to the next mac. there is no restriction on KMP to give macs in a specific order, so there is no issue here.

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -110,12 +117,12 @@ func (p *PoolManager) ReleaseAllVirtualMachineMacs(vmFullName string, parentLogg
 	return nil
 }
 
-func (p *PoolManager) UpdateMacAddressesForVirtualMachine(previousVirtualMachine, virtualMachine *kubevirt.VirtualMachine, transactionTimestamp *time.Time, parentLogger logr.Logger) error {
+func (p *PoolManager) UpdateMacAddressesForVirtualMachine(previousVirtualMachine, virtualMachine *kubevirt.VirtualMachine, transactionTimestamp *time.Time, isPersistent bool, parentLogger logr.Logger) error {
```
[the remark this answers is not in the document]
oshoval: all over right ?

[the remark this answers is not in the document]
RamLavi: yes

oshoval:
addressed comments

pkg/pool-manager/pod_pool.go
```diff
@@ -173,12 +176,15 @@ func (p *PoolManager) allocatePodFromPool(network *multus.NetworkSelectionElemen
 		// the pod name may have not been updated in the webhook context yet,
 		// so we use a temp pod name and update it the the controller
 		podFullName = tempPodName
-		p.macPoolMap.createOrUpdateEntry(macAddr.String(), podFullName, network.Name)
+		if isNotDryRun {
```
oshoval, line 179: i can move isNotDryRun also as a parameter to createOrUpdateEntry which will return without doing anything, if needed

pkg/pool-manager/pool_test.go
```diff
@@ -355,7 +355,7 @@ var _ = Describe("Pool", func() {
 			newVM.Name = "newVM"
```
RamLavi, line 355: can you please update the terminology of the dryRun options?
```
In case of "server/All" dryRun, we do want to return an updated...
```

pkg/pool-manager/pool_test.go
```diff
@@ -684,7 +684,7 @@ var _ = Describe("Pool", func() {
 
 				By("Create a VM")
 				transactionTimestamp = updateTransactionTimestamp(0)
-				err := poolManager.AllocateVirtualMachineMac(newVM, &transactionTimestamp, logger)
+				err := poolManager.AllocateVirtualMachineMac(newVM, &transactionTimestamp, true, logger)
```
RamLavi, line 687: can you add one unit test with false?

RamLavi (changes requested):
loogs good, added small nits

pkg/pool-manager/pool_test.go
oshoval: updated PR desc

oshoval:
addressed comments

pkg/pool-manager/pool_test.go
```diff
@@ -706,6 +706,21 @@ var _ = Describe("Pool", func() {
...
+					newVMDryRun := masqueradeVM.DeepCopy()
+					newVMDryRun.Name = "newVMDryRun"
+
+					By("Create a dry run VM")
+					transactionTimestamp = updateTransactionTimestamp(0)
+					err := poolManager.AllocateVirtualMachineMac(newVM, &transactionTimestamp, false, logger)
+					Expect(err).ToNot(HaveOccurred(), "should successfully create dry run VM")
+				})
+				It("should not storage the dry run mac in the mac pool", func() {
+					Expect(poolManager.macPoolMap).To(HaveLen(1), "macPoolMap should hold only the mac address waiting for approval")
```
RamLavi, line 720: can we compare the entire `poolManager.macPoolMap` before and after?

oshoval:
```
Expect(poolManager.macPoolMap).To(HaveLen(1), "macPoolMap should hold only the mac address waiting for approval")
Expect(poolManager.macPoolMap[allocatedMac]).To(Equal(expectedMacEntry), "macPoolMap's mac's entry should be as expected")
```
We basically comparing the actual versus the expecter as a whole,
since there is 1 entry and we compare the value of that,
same as other tests are doing, and they compare what the real VM which is created did to the pool

i don't think its needed
if you insist let me know please with how you want to do that, to add those two lines also in the beginning of the before each ?

oshoval:
address comments

RamLavi (approved):
/lgtm
/approve

PR #330, opened 2021-08-24
kubectl, Support different cluster root directory

oshoval:
In order to support CNAO testing of kubemacpool e2e,
add an env var, that allows using different cluster root directory,
because when CNAO deploys the cluster, it is created on CNAO directory
but the kmp tests are running from a temp kmp folder that
doesn't have the cluster metadata.

```release-note
None
```

### DIFF
--- a/tests/kubectl/kubectl.go
+++ b/tests/kubectl/kubectl.go
@@ -2,16 +2,24 @@ package kubectl
 
 import (
 	"bytes"
+	"os"
 	"os/exec"
 )
 
 func Kubectl(command ...string) (string, string, error) {
 	var stdout, stderr bytes.Buffer
 	cmd := exec.Command("./cluster/kubectl.sh", command...)
-	cmd.Dir = ".."
+	cmd.Dir = getClusterRootDirectory()
 	cmd.Stderr = &stderr
 	cmd.Stdout = &stdout
 	err := cmd.Run()
 	return stdout.String(), stderr.String(), err
 }
 
+func getClusterRootDirectory() string {
+	dir, found := os.LookupEnv("CLUSTER_ROOT_DIRECTORY")
+	if !found {
+		return ".."
+	}
+	return dir
+}


tests/kubectl/kubectl.go
```diff
@@ -2,16 +2,24 @@ package kubectl
...
 	"bytes"
+	"os"
 	"os/exec"
 )
 
 func Kubectl(command ...string) (string, string, error) {
 	var stdout, stderr bytes.Buffer
 	cmd := exec.Command("./cluster/kubectl.sh", command...)
-	cmd.Dir = ".."
+	cmd.Dir = getClusterRootDirectory()
```
RamLavi, line 12: Let's parameterize the entire command (similar to [here](https://github.com/nmstate/kubernetes-nmstate/blob/794179d34d0b09697707a5927789b80f1c218129/test/cmd/kubectl.go#L7))
suggestion:
```
kubectl := environment.GetVarWithDefault("KUBECTL", "../cluster/kubectl.sh")
cmd := exec.Command(kubectl, command...)
...
```

then export the KUBECTL entire command like in knmstate automation [here](https://github.com/kubevirt/cluster-network-addons-operator/blob/main/automation/check-patch.e2e-nmstate-functests.sh#L40)

RamLavi (changes requested):
right now the path is different, but next time it might be something else like the script name.
Let's parameterize the entire kubectl command

tests/kubectl/kubectl.go
oshoval: thanks but what vars you set on KMP itself (when you run it on kmp itself and not with CNAO cluster) ?
i think it won't work on kmp itself out of the box, because the cmd.Dir is missing, unless you add another var that solves it

oshoval:
as we spoke offline, we are falling back to the current solution as the alternatives are problematic,

can you please add ok-to-test ?

PR #346, opened 2022-01-13
metrics: Use https instead of http

oshoval:
KMP uses controller runtime which creates an http
end point for /metrics.
In order to support https, we deploy rbac-proxy container
in the same pod of KMP.
The new container exposes https end point,
and connects internally to the http end point.
    
It is self signed, and protected with rbac rules.

Will work offline with Oren Cohen in order to use sha instead of label for the rbac-proxy image.
It is needed for downstream.

**Release note**:
```release-note
None
```

### DIFF
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -23,6 +23,7 @@ function getLatestPatchVersion {
 
 source ./cluster/cluster.sh
 CNAO_VERSION=v0.58.0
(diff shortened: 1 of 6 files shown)


oshoval:
/cc @qinqon 
can you please review ?

qinqon (approved):
/lgtm
/approve

PR #349, opened 2022-01-19
DNM: Use controller-runtime client

qinqon:

**Release note**:

```release-note

```


Makefile
```diff
@@ -98,7 +100,9 @@ container: manager
 docker-push:
 	docker push ${REGISTRY}/${IMG}:${IMAGE_TAG}
 	docker tag ${REGISTRY}/${IMG}:${IMAGE_TAG} ${REGISTRY}/${IMG}:${IMAGE_GIT_TAG}
+	docker tag ${REGISTRY}/${IMG}:${IMAGE_TAG} ${REGISTRY}/${IMG}:partial
```
RamLavi, line 103: what is this used for?

pkg/pool-manager/pool.go
```diff
@@ -52,10 +53,11 @@ var log = logf.Log.WithName("PoolManager")
 var now = func() time.Time { return time.Now() }
 
 type PoolManager struct {
-	kubeClient       kubernetes.Interface // kubernetes client
-	rangeStart       net.HardwareAddr     // fist mac in range
-	rangeEnd         net.HardwareAddr     // last mac in range
-	currentMac       net.HardwareAddr     // last given mac
+	cachedKubeClient client.Client
+	kubeClient       client.Client
```
RamLavi, line 57: why are we moving to client.Client?
I'm not against, just want to understand why

RamLavi (commented):
just some questions

Makefile
qinqon: I used this to test at different cluster, we can remove it.

pkg/pool-manager/pool.go
qinqon: to keep code that call client the same independently if it's using the cached versin or not.

An alternative is to use the cached client but with Selectors so we specify what we want to cache and what not.

PR #353, opened 2022-01-19
metrics: Use https instead of http

oshoval:

KMP uses controller runtime which creates an http
end point for /metrics.
In order to support https, we deploy rbac-proxy container
in the same pod of KMP.
The new container exposes https end point,
and connects internally to the http end point.

It is self signed, and protected with rbac rules.

CNAO will parameterize the kube-rbac-proxy image url, so we will able
to consume it on D/S.
See please https://github.com/kubevirt/cluster-network-addons-operator/pull/1144

**Release note**:
```release-note
None
```

### DIFF
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -23,6 +23,7 @@ function getLatestPatchVersion {
 
 source ./cluster/cluster.sh
 CNAO_VERSION=v0.58.0
+export KUBEVIRT_DEPLOY_PROMETHEUS=true
 
 #use kubevirt latest z stream release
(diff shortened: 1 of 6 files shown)


oshoval:
/hold 

on hold until the kube-rbac-proxy image will be adapted to D/S

oshoval:
/hold

just need to sync with Oren about it before we merge, because he needs to adapt the D/S scripts to use the new image var
see https://github.com/kubevirt/cluster-network-addons-operator/pull/1144

PR #358, opened 2022-03-03
vm-pool, Reject VMs with duplicate interfaces

RamLavi:
When applying a VM with multiple interfaces with the same name,
the VM is rejected but kubemacpool is mishandling the MACs,
causing a side effect where that MAC is no longer usable (it
is "taken" by the ghost VM).
This commit fixes this by checking and rejecting VMs with
duplicate interface names on kmp webhook context.

**Release note**:

```release-note
Kubemacpool webhook behavior is changed: virtualMachine with duplicate interface names will be rejected from now on.
```

### DIFF
--- a/pkg/pool-manager/virtualmachine_pool.go
+++ b/pkg/pool-manager/virtualmachine_pool.go
@@ -47,6 +47,13 @@ func (p *PoolManager) AllocateVirtualMachineMac(virtualMachine *kubevirt.Virtual
 		return nil
 	}
 
+	// We can't allow for duplicate interfaces names, as interface.Name is macMap's key.
+	if isNotDryRun {
+		if err := checkVmForInterfaceDuplication(virtualMachine); err != nil {
+			return err
+		}
+	}
+
 	vmFullName := VmNamespaced(virtualMachine)
 	if len(getVirtualMachineNetworks(virtualMachine)) == 0 {
 		logger.Info("no networks found for virtual machine, skipping mac allocation",
@@ -126,6 +133,13 @@ func (p *PoolManager) UpdateMacAddressesForVirtualMachine(previousVirtualMachine
 	}
 	defer p.poolMutex.Unlock()
 
+	// We can't allow for duplicate interfaces names, as interface.Name is macMap's key.
+	if isNotDryRun {
+		if err := checkVmForInterfaceDuplication(virtualMachine); err != nil {
+			return err
+		}
+	}
+
 	currentInterfaces := getVirtualMachineInterfaces(previousVirtualMachine)
 	requestInterfaces := getVirtualMachineInterfaces(virtualMachine)
 	logger.V(1).Info("data before update", "macPoolMap", p.macPoolMap, "currentInterfaces", currentInterfaces, "requestInterfaces", requestInterfaces)
@@ -197,6 +211,19 @@ func getVirtualMachineInterfaces(virtualMachine *kubevirt.VirtualMachine) []kube
 
 func getVirtualMachineNetworks(virtualMachine *kubevirt.VirtualMachine) []kubevirt.Network {
 	return virtualMachine.Spec.Template.Spec.Networks
+}
+
+func checkVmForInterfaceDuplication(virtualMachine *kubevirt.VirtualMachine) error {
+	vmInterfaces := getVirtualMachineInterfaces(virtualMachine)
+
+	tmpInterfaceMap := map[string]struct{}{}
+	for _, vmInterface := range vmInterfaces {
+		if _, exists := tmpInterfaceMap[vmInterface.Name]; exists {
+			return fmt.Errorf("Failed to mutate virtual machine: every interface must have a unique name")
+		}
+		tmpInterfaceMap[vmInterface.Name] = struct{}{}
+	}
+	return nil
 }
 
 func (p *PoolManager) allocateFromPoolForVirtualMachine(vmFullName string, iface kubevirt.Interface, isNotDryRun bool, parentLogger logr.Logger) (string, error) {
--- a/pkg/pool-manager/pool_test.go
+++ b/pkg/pool-manager/pool_test.go
@@ -348,9 +348,26 @@ var _ = Describe("Pool", func() {
 						Devices: kubevirt.Devices{
 							Interfaces: []kubevirt.Interface{masqueradeInterface, multusBridgeInterface}}},
 					Networks: []kubevirt.Network{podNetwork, multusNetwork}}}}}
+		duplicateInterfacesVM := kubevirt.VirtualMachine{ObjectMeta: metav1.ObjectMeta{Namespace: "default"}, Spec: kubevirt.VirtualMachineSpec{
+			Template: &kubevirt.VirtualMachineInstanceTemplateSpec{
+				Spec: kubevirt.VirtualMachineInstanceSpec{
+					Domain: kubevirt.DomainSpec{
+						Devices: kubevirt.Devices{
+							Interfaces: []kubevirt.Interface{masqueradeInterface, multusBridgeInterface, multusBridgeInterface}}},
+					Networks: []kubevirt.Network{podNetwork, multusNetwork}}}}}
 		updateTransactionTimestamp := func(secondsPassed time.Duration) time.Time {
 			return time.Now().Add(secondsPassed * time.Second)
 		}
+		It("should reject allocation if there are interfaces with the same name", func() {
+			poolManager := createPoolManager("02:00:00:00:00:00", "02:00:00:00:00:02")
+			newVM := duplicateInterfacesVM.DeepCopy()
+			newVM.Name = "duplicateInterfacesVM"
+
+			transactionTimestamp := updateTransactionTimestamp(0)
+			err := poolManager.AllocateVirtualMachineMac(newVM, &transactionTimestamp, true, logger)
+			Expect(err).To(HaveOccurred(), "Should reject an allocation of a vm with duplicate interface names")
+			Expect(poolManager.macPoolMap).To(BeEmpty(), "Should not allocate macs if there are duplicate interfaces")
+		})
 		It("should not allocate a new mac for bridge interface on pod network", func() {
 			poolManager := createPoolManager("02:00:00:00:00:00", "02:00:00:00:00:02")
 			newVM := sampleVM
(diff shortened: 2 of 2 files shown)


RamLavi:
/hold
need to check pods module as well

RamLavi:
/hold cancel
pods are not relevant for this issue

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -199,6 +213,19 @@ func getVirtualMachineNetworks(virtualMachine *kubevirt.VirtualMachine) []kubevi
...
+
+	tmpInterfaceMap := map[string]string{}
+	for _, vmInterface := range vmInterfaces {
+		if _, exists := tmpInterfaceMap[vmInterface.Name]; exists {
+			return false
+		}
+		tmpInterfaceMap[vmInterface.Name] = ""
+	}
+	return true
+}
```
qinqon, line 227:
```suggestion
func checkVmForInterfaceDuplication(virtualMachine *kubevirt.VirtualMachine) error {
	vmInterfaces := getVirtualMachineInterfaces(virtualMachine)

	tmpInterfaceMap := map[string]string{}
	for _, vmInterface := range vmInterfaces {
		if _, exists := tmpInterfaceMap[vmInterface.Name]; exists {
			return fmt.Errorf("Failed to mutate virtual machine: every interface must have a unique name")
		}
		tmpInterfaceMap[vmInterface.Name] = ""
	}
	return nil
}
```

This way on the two places we call it we don't have to duplicate the error

```golang
if err := checkVmForInterfaceDuplication(virtualMachine) ; err != nil {
    return err
}
		}
```

qinqon (changes requested):

maiqueb (commented):
Some comments.

pkg/pool-manager/virtualmachine_pool.go
RamLavi: Done

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -47,6 +47,13 @@ func (p *PoolManager) AllocateVirtualMachineMac(virtualMachine *kubevirt.Virtual
 		return nil
 	}
 
+	if isNotDryRun {
+		// We can't allow for duplicate interfaces names, as interface.Name is macMap's key.
+		if !checkVmForInterfaceDuplication(virtualMachine) {
```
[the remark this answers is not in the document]
RamLavi: Done

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -126,6 +133,13 @@ func (p *PoolManager) UpdateMacAddressesForVirtualMachine(previousVirtualMachine
 	}
 	defer p.poolMutex.Unlock()
 
+	if isNotDryRun {
+		// We can't allow for duplicate interfaces names, as interface.Name is macMap's key.
+		if !checkVmForInterfaceDuplication(virtualMachine) {
```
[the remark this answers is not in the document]
RamLavi: Done

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -199,6 +213,19 @@ func getVirtualMachineNetworks(virtualMachine *kubevirt.VirtualMachine) []kubevi
 	return virtualMachine.Spec.Template.Spec.Networks
 }
 
+func checkVmForInterfaceDuplication(virtualMachine *kubevirt.VirtualMachine) bool {
+	vmInterfaces := getVirtualMachineInterfaces(virtualMachine)
+
+	tmpInterfaceMap := map[string]string{}
```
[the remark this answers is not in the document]
RamLavi: Done

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -47,6 +47,11 @@ func (p *PoolManager) AllocateVirtualMachineMac(virtualMachine *kubevirt.Virtual
 		return nil
 	}
 
+	// We can't allow for duplicate interfaces names, as interface.Name is macMap's key.
+	if err := checkVmForInterfaceDuplication(virtualMachine); isNotDryRun && err != nil {
```
qinqon, line 51: We don't want to even run `checkVmForInterfaceDuplication` if we are on dry mode right ?

qinqon (changes requested):

pkg/pool-manager/virtualmachine_pool.go
maiqueb: yeah, with inlining I meant that we should do something like `if !isDryRun && ...`

qinqon: @RamLavi can you fix it ?

RamLavi: I don't think it's a working syntax. if we want to not run `checkVmForInterfaceDuplication` if not needed  then either we do but don't return the err from `checkVmForInterfaceDuplication`:
```
if isNotDryRun && checkVmForInterfaceDuplication(virtualMachine) != nil {
    ...
}
```
or do  it in a 2 lines like initially done
```
if isNotDryRun {
    if err := checkVmForInterfaceDuplication(virtualMachine); err != nil {
        return err
    }
}
```
do you have another option?

maiqueb: Yes, the latter (without in-lining) seems preferable. Sorry for dragging you into this.

RamLavi: no worries, better it be as clean as possible :) 

even better, moved the flag inside the function on @qinqon 's suggestion. what do you think?

qinqon (approved):
/lgtm

pkg/pool-manager/virtualmachine_pool.go
maiqueb: I think boolean arguments suck. The caller knows what to do: call the function, or avoid doing so, based on the value of the boolean.

No point passing it around. 

My 2 cents :)

qinqon: Let's not inline the boolean check then and do as @RamLavi proposed

RamLavi: [Done](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/1ea79aa610967f9f4b42500bcfc7377e7ed1dafc..53f7cda95cc5b39942d4b6662a256682ec13ff49)

maiqueb:
I wonder about the approach altogether; afaiu, how the mutating webhook handles failures depend on its configuration - specifically, the value of its [`failurePolicy`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy). 

IIUC, this approach will only work if the failure policy is `fail`; since we own the webhook, that **should not be an issue**. Never the less, this might deserve a call out in the release notes: this approach is tied to the `fail` failure policy - which is the mutating webhook's default.

maiqueb:
> I wonder about the approach altogether; afaiu, how the mutating webhook handles failures depend on its configuration - specifically, the value of its [`failurePolicy`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy).
> 
> IIUC, this approach will only work if the failure policy is `fail`; since we own the webhook, that **should not be an issue**. Never the less, this might deserve a call out in the release notes: this approach is tied to the `fail` failure policy - which is the mutating webhook's default.

/hold 

The release notes should be updated indicating what is being fixed, and under what conditions - i.e. the `MutatingWebhookConfiguration` cannot be updated / must feature the `fail` failure policy.

RamLavi:
> I wonder about the approach altogether; afaiu, how the mutating webhook handles failures depend on its configuration - specifically, the value of its [`failurePolicy`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy).
> 
> IIUC, this approach will only work if the failure policy is `fail`; since we own the webhook, that **should not be an issue**. Never the less, this might deserve a call out in the release notes: this approach is tied to the `fail` failure policy - which is the mutating webhook's default.

Well this webhook implementation has always relied on the fact that the `failurePolicy` is `fail`. nothing changed in that regard. Moreover, the check we are adding is already being done, only later, in the webhook validation phase.
So, considering what you say, we have 2 options here:
1. we can fail the VM on our webhook, saying we cannot deal with duplicate interface names, due to our current implementation (this is what we do in this PR)
2. we can ignore the VM on our webhook (without assigning MACs, thus avoiding the side effect), letting the VM fail in the validation part later on. 

IMO I'm not too excited about option 2. I prefer to fail instead of ignoring the issue.. 
what do you think @maiqueb @qinqon ?

qinqon:
> > I wonder about the approach altogether; afaiu, how the mutating webhook handles failures depend on its configuration - specifically, the value of its [`failurePolicy`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy).
> > IIUC, this approach will only work if the failure policy is `fail`; since we own the webhook, that **should not be an issue**. Never the less, this might deserve a call out in the release notes: this approach is tied to the `fail` failure policy - which is the mutating webhook's default.
> 
> Well this webhook implementation has always relied on the fact that the `failurePolicy` is `fail`. nothing changed in that regard. Moreover, the check we are adding is already being done, only later, in the webhook validation phase. So, considering what you say, we have 2 options here:
> 
> 1. we can fail the VM on our webhook, saying we cannot deal with duplicate interface names, due to our current implementation (this is what we do in this PR)
> 2. we can ignore the VM on our webhook (without assigning MACs, thus avoiding the side effect), letting the VM fail in the validation part later on.
> 
> IMO I'm not too excited about option 2. I prefer to fail instead of ignoring the issue.. what do you think @maiqueb @qinqon ?

Fail fast is always better.

maiqueb:
> > > I wonder about the approach altogether; afaiu, how the mutating webhook handles failures depend on its configuration - specifically, the value of its [`failurePolicy`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy).
> > > IIUC, this approach will only work if the failure policy is `fail`; since we own the webhook, that **should not be an issue**. Never the less, this might deserve a call out in the release notes: this approach is tied to the `fail` failure policy - which is the mutating webhook's default.
> > 
> > 
> > Well this webhook implementation has always relied on the fact that the `failurePolicy` is `fail`. nothing changed in that regard. Moreover, the check we are adding is already being done, only later, in the webhook validation phase. So, considering what you say, we have 2 options here:

Right. Which means the mutating webhook mutates, and the validating webhook validates. So far so good. 

> > 
> > 1. we can fail the VM on our webhook, saying we cannot deal with duplicate interface names, due to our current implementation (this is what we do in this PR)

This is making the mutating webhook validate. Which is something that - imo - belongs on the validating webhook. I won't pretend that I am versed enough in webhook design patterns to claim that validating stuff in the mutating webhook is an anti-pattern, but it surely feels semantically weird. That's why I advocated out of band for the reconcile loop to sort this inconsistency.

> > 2. we can ignore the VM on our webhook (without assigning MACs, thus avoiding the side effect), letting the VM fail in the validation part later on.

And let eventually the reconcile loop correct the state. IMO, this is more aligned with the kubernetes design model.
> > 
> > IMO I'm not too excited about option 2. I prefer to fail instead of ignoring the issue.. what do you think @maiqueb @qinqon ?
> 
> Fail fast is always better.

...
Now the key thing is (quoting @RamLavi) :
> > Well this webhook implementation has always relied on the fact that the `failurePolicy` is `fail`. nothing changed in that regard. 

Right, meaning whatever implicit rules I'm trying to uphold were already broken. No point arguing about them now; your proposed fix seems more in-line with the current design of this component.

Please detail the fix in the release notes - afaiu, you're changing the behavior of the webhook. Furthermore ... shouldn't you also delete this particular validation from the validating webhook ? It will be redundant now.

maiqueb (approved):
Thank you for your patience.

PR #361, opened 2022-03-14
webhook: Configure TLS Min Version and Cipher Suites using environment variables

qinqon:

This PR add a pair of envinonment variables to tune the TLS min version and ciper suites, it's needed in case users want to enforce different levels of TLS security.

The environment variables introduced are:
- `TLS_CIPHERS`: Comma separated list of cipher suites names from https://pkg.go.dev/crypto/tls#pkg-constants
- `MIN_TLS_VERSION`: It can be "1.0", "1.1", "1.2" or "1.3"

Depends on: 
- https://github.com/qinqon/kube-admission-webhook/pull/65

**Release note**:

```release-note
Expose TLS min versions with TLS_MIN_VERSION
Expose TLS CipherSuites with TLS_CIPHERS
```


qinqon:
/hold

It's using a not merged PR at kube-admission-webhook
https://github.com/qinqon/kube-admission-webhook/pull/65

pkg/webhook/webhook.go
```diff
@@ -74,3 +75,7 @@ func cipherSuites() []string {
 	}
 	return strings.Split(cipherSuitesEnv, ",")
 }
+
+func tlsMinVersion() string {
```
rhrazdil, line 79: When os.Getenv("TLS_MIN_VERSION") is empty, the server will default to "1.0", maybe worth a comment, maybe not.

pkg/webhook/webhook.go
```diff
@@ -60,3 +66,11 @@ func AddToManager(mgr manager.Manager, poolManager *pool_manager.PoolManager) er
 	}
 	return nil
 }
+
+func cipherSuites() []string {
+	cipherSuitesEnv := os.Getenv("TLS_CIPHERS")
```
rhrazdil, line 72: What happens if TLS_CIPHERS happens to be empty?

pkg/webhook/webhook.go
qinqon: Adding comment

pkg/webhook/webhook.go
qinqon: Then the decision is delegated to go tls package.

rhrazdil (approved):

qinqon:
@RamLavi can you ack this ?

PR #365, opened 2022-03-20
unit tests: Add make check

oshoval:
In order to detect that we don't miss updating files,
with make generate, add make check to the unit tests job.

**Release note**:

```release-note
None
```

### DIFF
--- a/automation/check-patch.unit-test.sh
+++ b/automation/check-patch.unit-test.sh
@@ -11,6 +11,7 @@ main() {
     trap teardown EXIT SIGINT SIGTERM SIGSTOP
 
     make all
+    make check
     make test
 }
 


RamLavi:
can you push a tmp commit the messes with the fmt check for example?

oshoval:
> can you push a tmp commit the messes with the fmt check for example?

checked locally that `make check` catch it, but np will test here as well

oshoval:
reverted the manifests PR and it found it, removing the commit
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/k8snetworkplumbingwg_kubemacpool/365/pull-kubemacpool-unit-test/1505526700451041280#1:build-log.txt%3A126

RamLavi:
Thanks @oshoval for checking this!
/lgtm
/approve

PR #367, opened 2022-04-06
vendor, Bump prometheus/client_golang to v1.11.1

RamLavi:
This PR bumps prometheus/client_golang to v1.11.1

**Release note**:

```release-note
NONE
```


go.mod
```diff
@@ -103,6 +103,7 @@ replace (
 	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
 	github.com/openshift/api => github.com/openshift/api v3.9.0+incompatible
 	github.com/openshift/client-go => github.com/openshift/client-go v3.9.0+incompatible
+	github.com/prometheus/client_golang => github.com/prometheus/client_golang v1.11.1
```
oshoval, line 106: please change also here
https://github.com/k8snetworkplumbingwg/kubemacpool/pull/367/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R12

oshoval (commented):
Thanks, 
If there is a reason please add it to the PR desc

go.mod
RamLavi: I don't follow, can you be specific what change is needed? (AFAIU go tidy go vendor should generate the change..)

oshoval: you can pin the version on the require as well i think ? not just on the replace (it is fine to keep it on the replace as well)
so if you can update the version of `github.com/prometheus/client_golang v1.11.0` go.mod line 12 as well ?

RamLavi: ah.. why is it better to do so?

oshoval: just feels more straight forward imho, instead of have two different versions in the file when we don't have to
unless you are against

qinqon:
Can we enforce it at `require` instead of `replace` ?

RamLavi:
> Can we enforce it at `require` instead of `replace` ?

I think we can, but won't it render this change implicit? 
I think we need to make sure that if someone wants to use an older version in some future PR, then make vendor should alert him.. but I won't insist on it.. what do you think?

go.mod
qinqon: except go.mod all the files are generated, so this is fine.

qinqon:
> 

Let's go with require if replace is not needed.

RamLavi:
> > 
> 
> Let's go with require if replace is not needed.

done

go.sum
```diff
@@ -744,8 +744,9 @@ github.com/prometheus/client_golang v0.9.3/go.mod h1:/TN21ttK/J9q6uSwhBd54HahCDf
 github.com/prometheus/client_golang v1.0.0/go.mod h1:db9x61etRT2tGnBNRi70OPL5FsnadC4Ky3P0J6CfImo=
 github.com/prometheus/client_golang v1.1.0/go.mod h1:I1FGZT9+L76gKKOs5djB6ezCbFQP1xR9D75/vuwEF3g=
 github.com/prometheus/client_golang v1.7.1/go.mod h1:PY5Wy2awLA44sXw4AOSfFBetzPP4j5+D6mVACh+pe2M=
```
phoracek, line 746: Seems like we have to use `replace`

phoracek (changes requested):

go.sum
RamLavi: right!
[looks](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/3d69f4aad74da5c05628ca2f782830cc95e18db0/go.sum#L738) better now. [Done](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/e0f54533f2da93f89ae48dc77cd9236849e60396..3d69f4aad74da5c05628ca2f782830cc95e18db0)

oshoval: wonder if the combination of require after `go mod tidy --compact` would fix it as well ?
I think it did fix for me one time

RamLavi: is <VERSION> the golang version? how is it related? this bump is of an external library. could you explain?

oshoval: yes, golang version
see here please https://github.com/nmstate/kubernetes-nmstate/blob/main/Makefile#L197

It might be not related, but i think it might worth to check it, for example if each go version had the specific version hash that it included, and once you tidy per specific version it cleans them all
I might saw it happen locally one time, so maybe worth to check, as it is pretty easy to do please

RamLavi: I think it's only relevant to standard golang libraries, but can't hurt to check

PR #368, opened 2022-05-16
oci: Support podman

oshoval:
Support (rootfull) podman OCI
Auto detect running OCI

Bump kubevirtci to have podman support

**Release note**:
```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@ REPO ?= kubevirt
 IMAGE_TAG ?= latest
 IMAGE_GIT_TAG ?= $(shell git describe --abbrev=8 --tags)
 IMG ?= $(REPO)/kubemacpool
+OCI_BIN ?= $(shell if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi)
+TLS_SETTING := $(if $(filter $(OCI_BIN),podman),--tls-verify=false,)
 
 BIN_DIR = $(CURDIR)/build/_output/bin/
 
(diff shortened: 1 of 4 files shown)


oshoval:
Hi @qinqon 
Can you take a look please ?

Makefile
```diff
@@ -4,6 +4,8 @@ REPO ?= kubevirt
 IMAGE_TAG ?= latest
 IMAGE_GIT_TAG ?= $(shell git describe --abbrev=8 --tags)
 IMG ?= $(REPO)/kubemacpool
+OCI_BIN ?= $(shell if hash podman 2>/dev/null; then echo podman; elif hash docker 2>/dev/null; then echo docker; fi)
```
qinqon, line 7: Why don't we use `determine_cri_bin` here ?

qinqon (changes requested):

Makefile
oshoval: because it is in hack/common.sh, so we will need to make this file standalone runnable (or source it unless it does more things)

btw i am checking both here and determine_cri_bin to be simpler, logic like
`export OCI_BIN ?= $(shell if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi)`

oshoval:
Made the logic simpler

Makefile
qinqon: ack, let's keep it simple.

oshoval:
Thanks, 
we need to add colors to the lgtm approve labels ;]

PR #370, opened 2022-07-19
kubevirtci: Introduce bump script

oshoval:
Built such that automation can use it:
In case fatal happens return error code.
In case all is good, determine if there was a change
according non empty stdout.

**Release note**:
```release-note
None
```

### DIFF
--- a/hack/bump-kubevirtci.sh
+++ b/hack/bump-kubevirtci.sh
@@ -1,0 +1,8 @@
+#!/bin/bash -e
+
+KUBEVIRTCI_TAG=$(curl -L -Ss https://storage.googleapis.com/kubevirt-prow/release/kubevirt/kubevirtci/latest)
+[[ ${#KUBEVIRTCI_TAG} != "18" ]] && echo "error getting KUBEVIRTCI_TAG" && exit 1
+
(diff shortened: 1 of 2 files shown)


hack/bump-kubevirtci.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -e
+
+KUBEVIRTCI_TAG=$(curl -L -Ss https://storage.googleapis.com/kubevirt-prow/release/kubevirt/kubevirtci/latest)
+[[ ${#KUBEVIRTCI_TAG} != "18" ]] && echo "error getting KUBEVIRTCI_TAG" && exit 1
```
rhrazdil, line 4: Could you assign the magic number to a constant to make clear what is the purpose?
eg. EXPECTED_KUBEVIRTCI_TAG_LENGTH

oshoval:
Thanks Radim
@RamLavi 
Can you take a look please ?

oshoval:
> Also, do you plan to use it from other repos like project-infra?
we don't going to do automation at the moment if thats what you mean

oshoval:
Added makefile target

RamLavi (approved):
/lgtm
/approve

PR #383, opened 2022-12-13
makeFile: Remove vendored go

RamLavi:
Since we have the go.mod "go" directive it is safe enough to ensure that the golang compiler used has the needed capabilities. This change remove the infra to use a vendored golang tool chain.

**Release note**:

```release-note
NONE
```


RamLavi:
/hold
seems like the lane image doesn't have go.

qinqon:
/lgtm
/approve
we need to install go from automation or prow-jobs

Makefile
```diff
@@ -11,34 +11,25 @@ BIN_DIR = $(CURDIR)/build/_output/bin/
...
 all: generate
 
-$(GO):
-	hack/install-go.sh $(BIN_DIR) > /dev/null
-
 # Run tests
-test: $(GO)
-	$(GO) test ./pkg/... ./cmd/... -coverprofile cover.out
+test:
+	go test ./pkg/... ./cmd/... -coverprofile cover.out
```
oshoval, line 29: doesnt it means that running it locally might override the go version that the user already have ?
it isn't good if so

hack/install-go-gvm.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -e
+
+install_gvm() {
```
oshoval, line 3: why gvm is better than the regular install we had ?

RamLavi:
closing this for now. Will revisit later

PR #389, opened 2023-07-16
manifests: Remove control-plane label from namespace

oshoval:
Following https://github.com/kubevirt/cluster-network-addons-operator/pull/1145/files#r796372158
We don't need this label on KMP, and it require busy work on CNAO (see linked issue below).
Hence remove it from the namespace.

Note: the 2nd commit is manual removing from the files, where `make generate` didn't remove from.

See first bullet at https://github.com/kubevirt/cluster-network-addons-operator/issues/1567

**Release note**:
```release-note
None
```

### DIFF
--- a/config/default/manager/manager.yaml
+++ b/config/default/manager/manager.yaml
@@ -3,7 +3,6 @@ kind: Namespace
 metadata:
   labels:
     pod-security.kubernetes.io/enforce: "restricted"
-    control-plane: mac-controller-manager
     # in case mutatepods is set to opt-out mode,
     # make sure that KubeMacPool pods are also opted-out
     # to prevent dead-lock.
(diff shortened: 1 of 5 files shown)


RamLavi:
Can you run it manually to see that the label doesn't "return" unannounced by kustomize?

oshoval:
> Can you run it manually to see that the label doesn't "return" unannounced by kustomize?

Good point,
tried, it doesn't

we should also include running make generate on the unit-test if we don't already

RamLavi:
> > Can you run it manually to see that the label doesn't "return" unannounced by kustomize?
> 
> Good point, tried, it doesn't

good.
/lgtm
/approve

> 
> we should also include running make generate on the unit-test if we don't already

not sure, but it's a good idea

oshoval:
Thanks, once there will be a release we can update a little CNAO code (for now we can just remove the code on the 1st bullet)

no rush

PR #394, opened 2023-09-27
check hashicorp: Add hashicorp whitelist

oshoval:
Validates that no hashicorp are used beside the whitelisted ones
(because they shifted to BSL license)

Atm no hashicorp is used, but lets protect in case someone will add.
Based on the whitelist that CNAO uses.

Need to validate git actions runs it on next PRs, as this is the first one.
(git actions is enabled)

**Release note**:

```release-note
None
```

### DIFF
--- a/.github/workflows
+++ b/.github/workflows
@@ -1,0 +1,10 @@
+name: Check HashiCorp Modules
+on: [push, pull_request]
+jobs:
+  check_modules:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


RamLavi (approved):
/lgtm
/approve

oshoval:
It seems git actions isnt configured correctly, as it had to run it already even before it is merged
if you can take a look please ?

PR #397, opened 2023-10-24
Fix duplicate check with different mac format

fossedihelm:
A macAddress can be specified with different separators:
 - ":" e.g. "02:00:00:00:ff:ff"
 - "-" e.g. "02-00-00-00-ff-ff"
 - "." e.g. "0200.0000.ffff"

Currently, the duplicate mac check works only if the two compared mac address use the same separator.
The duplicate check is made looking at the existence of key of the macMap. The keys of the mapMap are the macAddress string.

With this commit, a `macKey` struct is introduced. The latter stores the normalized string of the macAddress and can be used as unique key and identifier of a macAddress.

Fixes https://issues.redhat.com/browse/CNV-30974

**Release note**:

```release-note
Fix duplicate check with different mac address separator
```

### DIFF
--- a/pkg/pool-manager/macentry_test.go
+++ b/pkg/pool-manager/macentry_test.go
@@ -21,33 +21,33 @@ var _ = Describe("mac-entry", func() {
 		newTimestamp := now.Add(time.Second)
 		BeforeEach(func() {
 			poolManager.waitTime = waitTimeSeconds
-			poolManager.macPoolMap = map[string]macEntry{
-				"02:00:00:00:00:00": macEntry{
+			poolManager.macPoolMap = map[macKey]macEntry{
+				NewMacKey("02:00:00:00:00:00"): macEntry{
 					instanceName:         "vm/default/vm0",
 					macInstanceKey:       "validInterface",
 					transactionTimestamp: &currentTimestamp,
 				},
-				"02:00:00:00:00:01": macEntry{
+				NewMacKey("02:00:00:00:00:01"): macEntry{
 					instanceName:         "vm/ns0/vm1",
 					macInstanceKey:       "staleInterface",
 					transactionTimestamp: &staleTimestamp,
 				},
-				"02:00:00:00:00:02": macEntry{
+				NewMacKey("02:00:00:00:00:02"): macEntry{
 					instanceName:         "vm/ns2/vm2",
 					macInstanceKey:       "validInterface",
 					transactionTimestamp: &currentTimestamp,
 				},
-				"02:00:00:00:00:03": macEntry{
+				NewMacKey("02:00:00:00:00:03"): macEntry{
 					instanceName:         "vm/ns3-4/vm3-4",
 					macInstanceKey:       "staleInterface",
 					transactionTimestamp: &staleTimestamp,
 				},
-				"02:00:00:00:00:04": macEntry{
+				NewMacKey("02:00:00:00:00:04"): macEntry{
 					instanceName:         "vm/ns3-4/vm3-4",
 					macInstanceKey:       "validInterface",
 					transactionTimestamp: &currentTimestamp,
 				},
-				"02:00:00:00:00:05": macEntry{
+				NewMacKey("02:00:00:00:00:05"): macEntry{
 					instanceName:         "vm/default/vm0",
 					macInstanceKey:       "validInterface",
 					transactionTimestamp: nil,
@@ -61,7 +61,7 @@ var _ = Describe("mac-entry", func() {
 		}
 		DescribeTable("and performing hasExpiredTransaction on macPoolMap entry",
 			func(i *hasExpiredTransactionParams) {
-				entry := poolManager.macPoolMap[i.macAddress]
+				entry := poolManager.macPoolMap[NewMacKey(i.macAddress)]
 				isStale, err := entry.hasExpiredTransaction(poolManager.waitTime)
 				Expect(err).ToNot(HaveOccurred(), "hasExpiredTransaction should not return error")
 				Expect(isStale).To(Equal(i.shouldSucceed), fmt.Sprintf("mac entry %s staleness status is not as expected", i.macAddress))
@@ -104,7 +104,7 @@ var _ = Describe("mac-entry", func() {
 		}
 		DescribeTable("and performing hasPendingTransaction on macPoolMap entry",
 			func(i *hasPendingTransactionParams) {
-				entry := poolManager.macPoolMap[i.macAddress]
+				entry := poolManager.macPoolMap[NewMacKey(i.macAddress)]
 				hasPendingTransaction := entry.hasPendingTransaction()
 				Expect(hasPendingTransaction).To(Equal(i.shouldSucceed), fmt.Sprintf("mac entry %s update required status is not as expected", i.macAddress))
 			},
@@ -147,7 +147,7 @@ var _ = Describe("mac-entry", func() {
 		}
 		DescribeTable("and performing hasReadyTransaction on macPoolMap",
 			func(i *hasReadyTransactionParams) {
-				entry := poolManager.macPoolMap[i.macAddress]
+				entry := poolManager.macPoolMap[NewMacKey(i.macAddress)]
 				isUpdated := entry.hasReadyTransaction(i.latestPersistedTimestamp)
 				Expect(isUpdated).To(Equal(i.expectedEntryReadiness), fmt.Sprintf("should get expected readiness on mac %s", i.macAddress))
 			},
--- a/pkg/pool-manager/mackey.go
+++ b/pkg/pool-manager/mackey.go
@@ -1,0 +1,22 @@
+package pool_manager
+
+import "net"
+
+type macKey struct {
+	normalizedMacAddress string
+}
+
+// NewMacKey creates a macKey struct containing a mac string
+// that can be used for different mac string format comparison.
+// It uses net.ParseMAC function to parse the given macAddress
+// and then stores its String() value.
+// The given macAddress MUST have already been validated.
+func NewMacKey(macAddress string) macKey {
+	// we can ignore the error here since the string value has already been validated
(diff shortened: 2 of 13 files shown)


maiqueb (changes requested):
This PR is required but IMO it doesn't do enough since there are other supported types of format for MAC addresses:
- canonical (`-` separator, which your PR adds support for)
- hexstring (`:` separator, the only thing currently accepted by kubemacpool)
- dot notation (`.` separator, in 4 char segments - e.g. `0000.5e00.5301`)
- hexadecimal 

Maybe the correct thing to do is to create a MAC address - `net.HardwareAddr` - using `net.ParseMAC` and then get the string from it (whatever that is) ?

fossedihelm:
> Maybe the correct thing to do is to create a MAC address - `net.HardwareAddr` - using `net.ParseMAC` and then get the string from it (whatever that is) ?

@maiqueb That's a great idea! IMHO we can also skip the `err` from ParseMAC function, since it has already been validated. WDYT?

maiqueb:
> > Maybe the correct thing to do is to create a MAC address - `net.HardwareAddr` - using `net.ParseMAC` and then get the string from it (whatever that is) ?
> 
> @maiqueb That's a great idea! IMHO we can also skip the `err` from ParseMAC function, since it has already been validated. WDYT?

Fine with me. I mean, there's nothing wrong in being redundant - checking against nil won't take long - but if you're sure that was already validated, all good. I just find it more readable / comforting to see the (ugly) golang pattern of `if err != nil ; return err` . It's up to the maintainer.

maiqueb (approved):
Looks good !

fossedihelm:
> Fine with me. I mean, there's nothing wrong in being redundant - checking against nil won't take long - but if you're sure that was already validated, all good. I just find it more readable / comforting to see the (ugly) golang pattern of `if err != nil ; return err` . It's up to the maintainer.

Yes, I generally agree! In this  specific case, handling the error means to create something like this:
```
func NewMacKey(macAddressDashes string) (macKey, error) {
	macAddress, err := net.ParseMAC(macAddressDashes)
        if err != nil{
            return macKey{}, err
        }
	return macKey{normalizedMacAddress: macAddress.String()}, nil
}
```
This is not a problem at all, but this will not allow us to create inline entry in the macMap.
And the latter is very useful in tests

maiqueb:
> > Fine with me. I mean, there's nothing wrong in being redundant - checking against nil won't take long - but if you're sure that was already validated, all good. I just find it more readable / comforting to see the (ugly) golang pattern of `if err != nil ; return err` . It's up to the maintainer.
> 
> Yes, I generally agree! In this specific case, handling the error means to create something like this:
> 
> ```
> func NewMacKey(macAddressDashes string) (macKey, error) {
> 	macAddress, err := net.ParseMAC(macAddressDashes)
>         if err != nil{
>             return macKey{}, err
>         }
> 	return macKey{normalizedMacAddress: macAddress.String()}, nil
> }
> ```
> 
> This is not a problem at all, but this will not allow us to create inline entry in the macMap. And the latter is very useful in tests, in addition to the fact that we should clearly always handle the return error.

Maybe it's best to add a comment to the `NewMacKey` func indicating it's inputs **must** be already validated MAC addresses ? And indicate all it does is parse from N mac address formats to "our" definition of canonical format.

This is just a nit. Consider adding it if someone else asks you to change the code.

fossedihelm:
> Maybe it's best to add a comment to the `NewMacKey` func indicating it's inputs **must** be already validated MAC addresses ? And indicate all it does is parse from N mac address formats to "our" definition of canonical format.
> 
> This is just a nit. Consider adding it if someone else asks you to change the code.

I put this just before the ParseMAC func:
```
// we can ignore the error here since the string value has already been validated
```
but yes, I can insert the description above the func! Thanks, will do!

maiqueb (approved):

pkg/pool-manager/pool.go
```diff
@@ -79,7 +79,26 @@ type macEntry struct {
 	transactionTimestamp *time.Time
 }
 
-type macMap map[string]macEntry
+type macKey struct {
```
RamLavi, line 82: I think this one deserves a separate file. Please move this to a new file `macKey.go`, and add the appropriate `macKey_tests.go` with unit tests for it (nothing fancy, I think this is enough:
```
var _ = Describe("MacKey", func() {
	DescribeTable("String should return the normalized MAC address",
		func(input string, expected string) {
			macKey := macKey{normalizedMacAddress: input}
			Expect(macKey.String()).To(Equal(expected))
		},
		Entry("colon-separated input", "00:00:5e:00:53:01", "00:00:5e:00:53:01"),
		Entry("hyphen-separated input", "00-00-5e-00-53-01", "00:00:5e:00:53:01"),
		Entry("period-separated input", "0000.5e00.5301", "00:00:5e:00:53:01"),
	)
})
```
)

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -431,7 +431,7 @@ func (p *PoolManager) revertAllocationOnVm(vmName string, allocations map[string
 
 	log.V(1).Info("Revert vm allocation", "vmName", vmName, "allocations", allocations)
 	for _, macAddress := range allocations {
-		delete(p.macPoolMap, macAddress)
+		p.macPoolMap.removeMacEntry(macAddress)
```
RamLavi, line 434: ditto, I would move to the preliminary commit

RamLavi (changes requested):
Hey @fossedihelm, thanks for your contribution!
Please see comments added.

RamLavi:
/hold

let's merge only after stable branches are introduced to KMP

fossedihelm:
Thanks @RamLavi for the review!
I am working on the relevant failing e2e tests. Thank you!

pkg/pool-manager/macentry_test.go
```diff
@@ -21,33 +21,33 @@ var _ = Describe("mac-entry", func() {
 		newTimestamp := now.Add(time.Second)
 		BeforeEach(func() {
```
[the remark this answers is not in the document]
fossedihelm: Done!

pkg/pool-manager/pod_pool.go
```diff
@@ -304,7 +304,7 @@ func (p *PoolManager) revertAllocationOnPod(podFullName string, allocations map[
 
 	log.V(1).Info("Revert vm allocation", "podFullName", podFullName, "allocations", allocations)
 	for _, macAddress := range allocations {
-		delete(p.macPoolMap, macAddress)
```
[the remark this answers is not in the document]
fossedihelm: Done!

pkg/pool-manager/pool.go
fossedihelm: Nice idea! Thanks! Done

fossedihelm:
[Changes](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/2479b07cc20aa2c5f3347e5fc9d413852b6f2510..447fd6637834eb509bf739eb7f398a004641f8f5):

- Addressed comments
- Fixed failing e2e test:
--  there were a mixed comparison between `iface.MacAddress` and `macKey.String()`
-- the first one was upper case, the latter in lower case.
- Added `MarshalJSON() ([]byte, error)`  to macMap struct to allow logger to correctly print the map

pkg/pool-manager/macpoolmap.go
```diff
@@ -98,7 +98,7 @@ func (m *macMap) alignMacEntryAccordingToVmInterface(macAddress, instanceFullNam
 	logger := log.WithName("alignMacEntryAccordingToVmInterface")
 	macEntry, _ := m.findByMacAddress(macAddress)
 	for _, iface := range vmInterfaces {
-		if iface.MacAddress == macAddress {
+		if NewMacKey(iface.MacAddress).normalizedMacAddress == macAddress {
```
RamLavi, line 101: can we use the `String` method?
```suggestion
		if NewMacKey(iface.MacAddress).String() == macAddress {
```

I think it works, see [here](https://go.dev/play/p/c0PoTSrbmF1)

pkg/pool-manager/mackey.go
```diff
@@ -0,0 +1,22 @@
...
+// It uses net.ParseMAC function to parse the given macAddress
+// and then stores its String() value.
+// The given macAddress MUST have already been validated.
+func NewMacKey(macAddress string) macKey {
+	// we can ignore the error here since the string value has already been validated
+	hwMacAddress, _ := net.ParseMAC(macAddress)
+	return macKey{normalizedMacAddress: hwMacAddress.String()}
+}
+
+func (m *macKey) String() string {
```
RamLavi, line 20: Since we are not changing the value of the object, there is no need to use a pointer to the object
```suggestion
func (m macKey) String() string {
```

This has an added value of being able to call the `String()` right after creating the object: `NewMacKey(MacAddress).String()`

RamLavi (changes requested):
thanks for the changes! 
Big like for the Marshal function added.

Added one more comment.

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -528,7 +528,7 @@ func (p *PoolManager) recoverVmFromCluster(macAddress string) (*kubevirt.Virtual
 			return nil
 		}
 
-		if iface.MacAddress != "" && iface.MacAddress == macAddress {
+		if iface.MacAddress != "" && NewMacKey(iface.MacAddress).normalizedMacAddress == macAddress {
```
RamLavi, line 531:
```suggestion
		if iface.MacAddress != "" && NewMacKey(iface.MacAddress).String() == macAddress {
```

RamLavi (changes requested):
last one :)

RamLavi (approved):
/lgtm
/approve

@fossedihelm does this change need to be backported? 
If not, then I'll remove the old after the CVE storm passes and I add the appropriate stable branches...

PR #426, opened 2024-04-15
cert-manager: Cache only kubemacpool secrets

qinqon:
The webhook secrets and services at kubemacpool are living under kubemacpool namespace, this change limit cert-manager reconciler to that so it will not cache secrets at all namespaces.

Related jira https://issues.redhat.com/browse/CNV-38129

**Release note**:

```release-note
Limit cert-manager to kubemacpool namespace
```


RamLavi (approved):
thanks!
/lgtm
/approve

PR #427, opened 2024-05-01
go.mod: Update go version to 1.21

RamLavi:
The current version 1.19 is causing `make generate` to fail on format issue.
Updating to 1.21.
Also updating the `deepcopy-gen` flags as the older ones are no longer supported.

**Release note**:

```release-note
NONE
```


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/k8snetworkplumbingwg/kubemacpool
 
-go 1.19
+go 1.21.9
```
oshoval, line 3: Thanks, did you run make vendor ?
asking because iirc a `toolchain` would be added once you do
but on the other hand i guess you did

RamLavi: did not. will do now

RamLavi: FYI I ran it now, but the `toolchain` thing was not added

oshoval: maybe ye, anyhow once you ran we are good
(worth to add make vendor on CI unit tests on a followup)

oshoval (approved):

go.mod
oshoval: btw seems it is from 1.21
https://go.dev/doc/toolchain
not sure why it is not added, but we will figure in the future, doesn't matter as long as everything works

PR #431, opened 2024-07-03
pool-manager: Prevent panic if VM template is nil

RamLavi:
This PR fixes a panic that occurs if a VM is created/updated with no template.
Although this field is mandatory, kubemacpool should not panic if it happens.

**Release note**:

```release-note
Fix panic over nil template
```


qinqon (approved):
/lgtm
/approve

PR #432, opened 2024-07-03
cluster: Allow kubevirtci tag overriding

oshoval:
It is helpful either manually or for example
when used from CNAO, to use same tag of the cluster that CNAO deployed.
Otherwise a kubevirtci tag mismatch can cause failures once running 
KMP e2e tests from CNAO.

**Release note**:

```release-note
None
```

### DIFF
--- a/cluster/cluster.sh
+++ b/cluster/cluster.sh
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.25'}
-export KUBEVIRTCI_TAG='2211021552-8cca8c0'
+export KUBEVIRTCI_TAG=${KUBEVIRTCI_TAG:-2211021552-8cca8c0}
 
 KUBEVIRTCI_REPO='https://github.com/kubevirt/kubevirtci.git'
 # The CLUSTER_PATH var is used in cluster folder and points to the _kubevirtci where the cluster is deployed from.
--- a/hack/bump-kubevirtci.sh
+++ b/hack/bump-kubevirtci.sh
@@ -3,6 +3,6 @@
 KUBEVIRTCI_TAG=$(curl -L -Ss https://storage.googleapis.com/kubevirt-prow/release/kubevirt/kubevirtci/latest)
 [[ ${#KUBEVIRTCI_TAG} != "18" ]] && echo "error getting KUBEVIRTCI_TAG" && exit 1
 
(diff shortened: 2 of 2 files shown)


oshoval:
/cc @RamLavi 
we need this (+ release once it is merged please)
so we can bump kubevirtci on CNAO
otherwise if the cluster kubevirtci tag CNAO deployed is different then the kubevirtci helper scripts that KMP uses
it can product failures.
with this small change, the tag CNAO uses will be also used whenever needed when CNAO runs KMP e2e tests

oshoval:
need first to adapt bump-kubevirtci.sh

qinqon (approved):
/lgtm
/approve

oshoval:
Thanks
Can you please create a patch release 0.43.1 on main so CNAO consume it ?
@phoracek @RamLavi @qinqon

It will be indeed nice when we implement mechanism to allow consuming non core related changes
by CNAO, i will work on it when i have time.

RamLavi:
@oshoval you need to backport this to `release-0.43` first no?

RamLavi:
> @oshoval you need to backport this to `release-0.43` first no?

oh my bad, we are already after CF on that stable branch.
This change will only affect U/S,  am I correct? 
And only the U/S dev env (i.e. the lanes etc..) right?

I still wonder why you would want to backport it to that stable branch

oshoval:
> > @oshoval you need to backport this to `release-0.43` first no?
> 
> oh my bad, we are already after CF on that stable branch. This change will only affect U/S, am I correct? And only the U/S dev env (i.e. the lanes etc..) right? If the answer is yes to both of the above then I assume that backporting this change is safe enough to do without getting QE to approve it and all that, but I still wonder why you would want to backport it to that stable branch - It's going GA in a few days
> 
> Perhaps you meant you want to issue a new `main` release (which will be minor `0.44.0` IIUC)? - in that case I'm all forward

this change doesn't affect KMP at all (as it only allow external overriding the default value)
we just need a main release so CNAO can consume it (it will help CNAO, not KMP)
ye we can use 0.44.0 if 0.43.0 is stable, in any case since CNAO main watches KMP main, the release must be on main
Thanks

RamLavi:
> can you please create new minor 0.44.0 on main ?

https://github.com/k8snetworkplumbingwg/kubemacpool/releases/tag/v0.44.0

oshoval:
> > can you please create new minor 0.44.0 on main ?
> 
> https://github.com/k8snetworkplumbingwg/kubemacpool/releases/tag/v0.44.0

Thanks !

PR #435, opened 2024-08-15
rbac: Move to use private ServiceAccount

RamLavi:
Currently kubemacpool pods are using the default SA. Therefore extra permissions are being granted to the default SA.
It would be better if they will use a dedicated SA.

**Release note**:

```release-note
NONE
```


qinqon (approved):
/lgtm
/approve

PR #441, opened 2024-11-25
Enable Multi-Platform Support for kubemacpool Image Builds

ashokpariya0:
These changes enable building and pushing kubemacpool container images for multiple platforms (amd64, s390x, arm64) from a single Dockerfile. Enhanced multi-platform support in the build process by adding a PLATFORMS argument in the Makefile for amd64, s390x, and arm64 architectures. Multi-platform build support is provided for both Docker and Podman container runtimes.

The upstream kubemacpool image ([link](https://quay.io/repository/kubevirt/kubemacpool?tab=tags&tag=latest)) currently supports only the amd64 architecture. The proposed changes in this PR enable multi-platform support for the kubemacpool image, broadening its compatibility to include additional architectures.

Multiplatform Build Instruction: https://gist.github.com/ashokpariya0/33cf22401d17153b7f91b40295397ccd

Multi-platform build support is enabled for both Docker and Podman container runtimes, and the necessary changes have been made accordingly.
The PLATFORMS variable defines the target platforms for building the manager image, allowing support for multiple architectures.

To use this feature, you need to:

For Docker:

Have access to docker buildx. More information: [[Docker Buildx Documentation](https://docs.docker.com/build/building/multi-platform/)](https://docs.docker.com/build/building/multi-platform/)
Ensure that BuildKit is enabled. More information: [[Docker BuildKit Enhancements](https://docs.docker.com/develop/develop-images/build_enhancements/)](https://docs.docker.com/develop/develop-images/build_enhancements/)
For Podman:

No additional dependencies are required to build multi-platform images. Podman supports this feature natively.

**Release note**:

```release-note
None
```


oshoval:
cc @RamLavi

Makefile
```diff
@@ -6,6 +6,16 @@ IMAGE_GIT_TAG ?= $(shell git describe --abbrev=8 --tags)
...
 TLS_SETTING := $(if $(filter $(OCI_BIN),podman),--tls-verify=false,)
+PLATFORMS ?= linux/amd64,linux/s390x
+# Set the platforms for building a multi-platform supported image.
+# Example:
+# PLATFORMS ?= linux/amd64,linux/arm64,linux/s390x
+# Alternatively, you can export the PLATFORMS variable like this:
+# export PLATFORMS=linux/arm64,linux/s390x,linux/amd64
+ARCH := $(shell uname -m | sed 's/x86_64/amd64/')
+DOCKER_BUILDER ?= kubemacpool-docker-builder
+KUBEMACPOOL_IMAGE_TAGGED_1 := ${REGISTRY}/${IMG}:${IMAGE_TAG}
```
RamLavi, line 17: please change `KUBEMACPOOL_IMAGE_TAGGED_1` --> `KUBEMACPOOL_IMAGE_TAGGED`
and `KUBEMACPOOL_IMAGE_TAGGED_2` --> `KUBEMACPOOL_IMAGE_GIT_TAGGED`
the current naming is not good, but it is still better than meaningless enumeration.

Makefile
```diff
@@ -83,18 +93,23 @@ generate: generate-go generate-deploy generate-test generate-external
...
+else
+	$(error Unsupported OCI_BIN value: $(OCI_BIN))
+endif
 
 # Push the docker image
 docker-push:
-	$(OCI_BIN) push ${TLS_SETTING} ${REGISTRY}/${IMG}:${IMAGE_TAG}
-	$(OCI_BIN) tag ${REGISTRY}/${IMG}:${IMAGE_TAG} ${REGISTRY}/${IMG}:${IMAGE_GIT_TAG}
-	$(OCI_BIN) push ${TLS_SETTING} ${REGISTRY}/${IMG}:${IMAGE_GIT_TAG}
+ifeq ($(OCI_BIN),podman)
```
RamLavi, line 108: why is there no docker counterpart? 
In any case, if we don't want to support pushing with docker, we should be explicit, failing the operation.

build/Dockerfile
```diff
@@ -1,4 +1,21 @@
+ARG BUILD_ARCH=amd64
```
RamLavi, line 1: please separate into 2 commits:
1. moving repo go build to container (+1 btw)
2. Adding multi arch thing..

hack/build-kubemacpool-podman.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+if [ -z "$ARCH" ] || [ -z "$PLATFORMS" ] || [ -z "$KUBEMACPOOL_IMAGE_TAGGED_1" ]; then
```
RamLavi, line 3: I'm missing KUBEMACPOOL_IMAGE_TAGGED_2 in the podman script. Is this intentional? 
I would want both tag to be there. having both `latest` and sha-specific [tags](https://quay.io/repository/kubevirt/kubemacpool?tab=tags&tag=latest) allow us to be flexible picking the desired image when deployed from other operators like CNAO.

Makefile
```diff
@@ -83,18 +93,23 @@ generate: generate-go generate-deploy generate-test generate-external
...
 	./hack/check.sh
 
-manager: $(GO)
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GO) build -o $(BIN_DIR)/manager github.com/k8snetworkplumbingwg/kubemacpool/cmd/manager
-
 # Build the docker image
-container: manager
-	$(OCI_BIN) build build/ -t ${REGISTRY}/${IMG}:${IMAGE_TAG}
+container:
+ifeq ($(OCI_BIN),podman)
```
RamLavi, line 98: I prefer not over-complicating the Makefile with these conditions. 
please call to a `build-multiarch-kubemacpool` script that deals with both podman and docker cases.

hack/build-kubemacpool-podman.sh
```diff
@@ -0,0 +1,25 @@
...
+podman rmi "${KUBEMACPOOL_IMAGE_TAGGED_2}" || true
+
+podman manifest create "${KUBEMACPOOL_IMAGE_TAGGED_1}"
+
+for platform in "${PLATFORM_LIST[@]}"; do
+    podman build \
+        --no-cache \
+        --build-arg BUILD_ARCH="$ARCH" \
+        --platform "$platform" \
+        --manifest "${KUBEMACPOOL_IMAGE_TAGGED_1}" \
```
RamLavi, line 23: again what happened to KUBEMACPOOL_IMAGE_TAGGED_2 tagging?

RamLavi (changes requested):
Hey @ashokpariya0 , thank you for this change.
I added some comments, please take a look

oshoval:
Please note that imo we should not merge any new PRs about cross compilation until
https://github.com/kubevirt/cluster-network-addons-operator/pull/1923#issuecomment-2509657197 issue is solved

1. We must make sure we have local machines that can run it (for me personally since the new machine isn't accessible right now, the old machine will take time to upgrade)
Alternative, we need to keep old docker support for now
2. CNAO merged PR might have some problems (some errors that are mentioned, need to check if they are important)
    and also not sure how all the lanes run.

Even if this effort is important, we should not rush it until it is done correctly please.

ashokpariya0:
> kubevirt/cluster-network-addons-operator#1923 (comment)
> Hey @ashokpariya0 , thank you for this change. I added some comments, please take a look

Okay, noted! I'll make the necessary updates.

Makefile
ashokpariya0: Done.

build/Dockerfile
ashokpariya0: Done.

Makefile
```diff
@@ -6,6 +6,16 @@ IMAGE_GIT_TAG ?= $(shell git describe --abbrev=8 --tags)
 IMG ?= $(REPO)/kubemacpool
 OCI_BIN ?= $(shell if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi)
 TLS_SETTING := $(if $(filter $(OCI_BIN),podman),--tls-verify=false,)
+PLATFORMS ?= linux/amd64,linux/s390x
```
oshoval, line 9: imo need to check benchmark before doing both
or best to just use the auto detect and have optional `all` flag as raised on CNAO please

build/Dockerfile
```diff
@@ -1,4 +1,21 @@
...
+FROM --platform=linux/${BUILD_ARCH} quay.io/centos/centos:stream9 AS builder
+RUN dnf install -y tar gzip jq && dnf clean all
+RUN ARCH=$(uname -m | sed 's/x86_64/amd64/') && \
+    GO_VERSION=$(curl -L -s "https://go.dev/dl/?mode=json" | jq -r '.[0].version') && \
```
oshoval, line 5: better to get go version from go.mod please
easier to maintain

hack/build-kubemacpool-docker.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+if [ -z "$ARCH" ] || [ -z "$PLATFORMS" ] || [ -z "$KUBEMACPOOL_IMAGE_TAGGED" ] || [ -z "$KUBEMACPOOL_IMAGE_GIT_TAGGED" ]; then
+    echo "Error: ARCH, PLATFORMS, KUBEMACPOOL_IMAGE_TAGGED, and KUBEMACPOOL_IMAGE_GIT_TAGGED must be set."
+    exit 1
+fi
+
+IFS=',' read -r -a PLATFORM_LIST <<< "$PLATFORMS"
+
+BUILD_ARGS="--no-cache --build-arg BUILD_ARCH=$ARCH -f build/Dockerfile -t $KUBEMACPOOL_IMAGE_TAGGED -t $KUBEMACPOOL_IMAGE_GIT_TAGGED . --push"
```
oshoval, line 10: why no cache?

hack/build-kubemacpool-podman.sh
```diff
@@ -0,0 +1,25 @@
...
+podman manifest rm "${KUBEMACPOOL_IMAGE_TAGGED}" 2>/dev/null || echo "Manifest ${KUBEMACPOOL_IMAGE_TAGGED} not found, skipping removal."
+podman manifest rm "${KUBEMACPOOL_IMAGE_GIT_TAGGED}" 2>/dev/null || echo "Manifest ${KUBEMACPOOL_IMAGE_GIT_TAGGED} not found, skipping removal."
+podman rmi "${KUBEMACPOOL_IMAGE_TAGGED}" 2>/dev/null || echo "Image ${KUBEMACPOOL_IMAGE_TAGGED} not found, skipping removal."
+podman rmi "${KUBEMACPOOL_IMAGE_GIT_TAGGED}" 2>/dev/null || echo "Image ${KUBEMACPOOL_IMAGE_GIT_TAGGED} not found, skipping removal."
+
+podman manifest create "${KUBEMACPOOL_IMAGE_TAGGED}"
+
+for platform in "${PLATFORM_LIST[@]}"; do
+    podman build \
+        --no-cache \
```
oshoval, line 20: why no cache?

hack/init-buildx.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+check_buildx() {
+  export DOCKER_CLI_EXPERIMENTAL=enabled
+
+  if ! docker buildx > /dev/null 2>&1; then
+     mkdir -p ~/.docker/cli-plugins
+     BUILDX_VERSION=$(curl -s https://api.github.com/repos/docker/buildx/releases/latest | jq -r .tag_name)
+     curl -L https://github.com/docker/buildx/releases/download/"${BUILDX_VERSION}"/buildx-"${BUILDX_VERSION}".linux-amd64 --output ~/.docker/cli-plugins/docker-buildx
```
oshoval, line 9: is it fine it has hardcoded `linux-amd64` ?

oshoval (commented):
some comments (fast pass)
thanks

hack/build-kubemacpool-podman.sh
ashokpariya0: Yes, both tags are properly handled.
For multi-architecture support, we create a single manifest and tag it with both the git_sha_tag and the latest tag before pushing. This process, managed in the Makefile, ensures that both tags are properly handled.

Makefile
ashokpariya0: Done.

hack/build-kubemacpool-podman.sh
ashokpariya0: As mentioned earlier, this script creates only a single Podman manifest(this how multi-platform build works with podman), while the tagging with `git_sha` is handled in the Makefile. so both tags are properly handled.

hack/build-kubemacpool-podman.sh
ashokpariya0: It seems --no-cache is recommend approach. Without --no-cache, the build is faster because it reuses previous layers. But this can sometimes cause outdated or incorrect dependencies to be used. Using --no-cache forces a fresh build, making sure all steps in the Dockerfile run again without using old layers. Since we already remove the old manifest before creating a new one, I think --no-cache is the best choice. However, I'm open to other suggestions.

oshoval: imho we can omit it please and use caching, docker caching seems too fundamental to have bugs beside corner cases
we use caching so far and it is fine iiuc
speed is important (of course unless we find in the future it is buggy and then can report / workaround that)

ashokpariya0: Okay sure.

Makefile
ashokpariya0: Done, Added auto detach and all flag option in latest commit.

hack/build-kubemacpool-docker.sh
ashokpariya0: Removed.

hack/init-buildx.sh
ashokpariya0: good catch, fixed it.

build/Dockerfile
ashokpariya0: Done.

ashokpariya0:
@RamLavi @oshoval I have addressed all the review comments and responded to the questions. Please take a look.

build/Dockerfile
```diff
@@ -1,4 +1,30 @@
...
+RUN dnf install -y wget && dnf clean all
+COPY go.mod .
+COPY go.sum .
+RUN GO_VERSION=$(sed -En 's/^go +(.*)$/\1/p' go.mod) && \
+    wget https://dl.google.com/go/go${GO_VERSION}.${BUILDOS}-${BUILDARCH}.tar.gz && \
+    tar -C /usr/local -xzf go${GO_VERSION}.${BUILDOS}-${BUILDARCH}.tar.gz && \
+    rm go${GO_VERSION}.${BUILDOS}-${BUILDARCH}.tar.gz
+
+ENV PATH=$PATH:/usr/local/go/bin
+COPY . .
```
oshoval, line 23: please consider adding a `RUN go mod download` before the copy, to improve layer building

build/Dockerfile
```diff
@@ -1,4 +1,30 @@
...
+
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -o build/_output/bin/manager github.com/k8snetworkplumbingwg/kubemacpool/cmd/manager
```
oshoval, line 25: is it just me or it builds remote sources which is not correct ?
i would expect something like
`go build -o build/_output/bin/manager ./cmd/manager`

hack/build-kubemacpool-docker.sh
```diff
@@ -0,0 +1,18 @@
...
+IFS=',' read -r -a PLATFORM_LIST <<< "$PLATFORMS"
+
+BUILD_ARGS="--build-arg BUILD_ARCH=$ARCH -f build/Dockerfile -t $KUBEMACPOOL_IMAGE_TAGGED -t $KUBEMACPOOL_IMAGE_GIT_TAGGED . --push"
+
+if [ ${#PLATFORM_LIST[@]} -eq 1 ]; then
+    docker build --platform "$PLATFORMS" $BUILD_ARGS
+else
+    ./hack/init-buildx.sh "$DOCKER_BUILDER"
+    docker buildx build --platform "$PLATFORMS" $BUILD_ARGS
+    docker buildx rm "$DOCKER_BUILDER" || true
```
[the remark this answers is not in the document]
ashokpariya0: Done.

build/Dockerfile
ashokpariya0: Done.

ashokpariya0:
@RamLavi @oshoval  All the review comments have been addressed. Please take a look.

build/Dockerfile
oshoval: thanks,
imho it is better to be explicit with what we build
`./cmd/manager`

ashokpariya0:
@RamLavi @oshoval I have addressed all the review comments. Please have a look.
cc @phoracek @ormergi

Makefile
RamLavi: Let me rephrase: in the existing code (https://github.com/k8snetworkplumbingwg/kubemacpool/blob/main/Makefile#L94-L98) you can either do docker/podman push/tag using the OCI_BIN env var.
Now however, you can only do it with podman:
```
ifeq ($(OCI_BIN),podman)
    push,tag,etc...
endif
...
```

my original question is - why are you limiting it like this?

build/Dockerfile
RamLavi: You did so but the order of the commits is wrong. 
Please separate to 2 commits. The first commit only moves the repo go build to a container, setting the ground to the main commit. The second does the main thing this PR is meant to do which is adding the multi arch

Makefile
RamLavi: Please change it in the appropriate commit. There should be no existence of KUBEMACPOOL_IMAGE_TAGGED_1 during any commit

RamLavi (changes requested):
please reorder the PR so that it is easier to review

Makefile
ashokpariya0: For building multi-platform images using docker buildx build, the build and push operations are combined into a single command, Therefore, the separate push and tag logic, which currently applies only to podman, isn't needed for docker when using buildx.

oshoval: theoretically it is possible to split even for docker single arch / multi arch ?
(i see here `--push` flag is used so i wonder)
https://github.com/k8snetworkplumbingwg/kubemacpool/pull/441/files#diff-8a6127c87c24bc27e5595386a152b920da0266e3b6d185a3e632cab440abf09bR10-R18

not saying we need to, first trying to understand

ashokpariya0: Currently it's not possible for multi arch images:
Below is the reason:

With docker buildx, the build result is stored only in the build cache by default. There are two options to handle the result, which I've already tested: --push and --load.

The --push flag pushes the built image to a container registry.
The --load flag loads the resulting image into the local Docker daemon. However, the --load option currently only supports single-platform images or simpler images. It does not work with multi-platform builds (i.e., manifest lists).
When using the --load flag with a multi-platform build, you'll encounter the following error:

docker exporter does not currently support exporting manifest lists

oshoval: thanks for the explanation
theoretically for single arch, it is possible then right?
it is nice to have, as it will allow to just build instead build + push, but i am not sure it worths the effort, and all the other repos have it combined already
moreover we encourage podman over docker

it is Ram decision here

ashokpariya0: Correct, but with single arch we can directly use docker build , for multiplatform we need to use docker buildx build and we have to do like this

ashokpariya0:
@RamLavi @oshoval I ran into some issues while reordering the commits, which led to a bit of confusion and extra time spent resolving it. To avoid further complications, I’ve decided to raise a new PR with the commits separated as you suggested.

The first commit now only addresses moving the go build process to a container, setting the groundwork for the main changes. The second commit focuses on the primary task of this PR, which is adding multi-architecture support.

Apologies for the inconvenience, and thank you for your understanding! You can find the new PR here: [#444](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/444).

oshoval:
all good thanks

Makefile
RamLavi: I see now. I don't like the asymmetry of it, but having build+push to the local repo isn't such a bad thing. 
Thanks for the explanation

PR #447, opened 2024-12-30
pool: Randomize mac

RamLavi:
Currently, the MAC allocation strategy relied on sequential assignment.
The system kept track of the last allocated MAC and assigned the next sequential MAC unless it was already in use.
This approach can cause issues in specific use cases where multi-clusters are merged into one, increasing the likelihood of MAC conflict.
Changing the algorithm to be random to ensure more robust MAC allocation for dynamic and distributed environments.

**Release note**:

```release-note
NONE
```


RamLavi:
/hold
depends on #448

RamLavi:
> /hold until https://github.com/k8snetworkplumbingwg/kubemacpool/pull/547 is merged

Change: [Rebased](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/bb6f1e94fe14a748e586022e5b3492b9235604b5..7c7b5de87644a34fcee964f6f695d86ed3f22647)

/hold cancel

qinqon (changes requested):
I am worry that choosing a random limit it limits the amounts of MACs kubemacpool can reserve, I understand the mac range offset is never used.

Do you know if at least kubemacpool can send a warning if a VM comes live with a collision mac ?

RamLavi:
> > > Do you know if at least kubemacpool can send a warning if a VM comes live with a collision mac ?
> > 
> > 
> > Not sure I'm following so just correct me if I'm not catching your train of thoughts, but KMP handles the VM regardless if it's alive(running) or not. If KMP tries to allocate a MAC from the range and it stumbles upon a used MAC, it will just try the next until it finds a free one. And if the user explicitly tries to use a specific MAC and it is used then it will block the VM creation/update via webhook.
> 
> Ok so in the scenario where a VM with a pre assigned mac on a backup startup on the cluster, if that mac is already assigned at other VM the vm create will fail at kubemacpool webhook, that's it ?
> 
> If this is the case then we good too.

yes. This is standard KMP work. 
The only thing we're changing is what is the first MAC used for allocation:
Before this PR:  it was the first MAC in the range
After: Some random MAC in the range.

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/7c7b5de87644a34fcee964f6f695d86ed3f22647..005cc57d6b293a10ed0367f3f4af5fcfb0d99003): Review fixes

qinqon (approved):
/lgtm
/approve

PR #453, opened 2025-01-14
e2e: Change deprecated vm Running API

RamLavi:
This PR fixed e2e infra that uses old kubevirt VM Running API.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/tests.go
+++ b/tests/tests.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/pointer"
+	"k8s.io/utils/ptr"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 	"kubevirt.io/client-go/kubecli"
 
@@ -91,11 +92,10 @@ func removeTestNamespaces() {
 		Should(BeTrue(), "Namespace %s haven't been deleted within the given timeout", TestNamespace)
 }
 
-func CreateVmObject(namespace string, running bool, interfaces []kubevirtv1.Interface, networks []kubevirtv1.Network) *kubevirtv1.VirtualMachine {
+func CreateVmObject(namespace string, interfaces []kubevirtv1.Interface, networks []kubevirtv1.Network) *kubevirtv1.VirtualMachine {
 	vm := getVMCirros()
 	vm.Name = randName("testvm")
 	vm.Namespace = namespace
-	vm.Spec.Running = &running
 	vm.Spec.Template.Spec.Domain.Devices.Interfaces = interfaces
 	vm.Spec.Template.Spec.Networks = networks
 
@@ -537,7 +537,7 @@ func getVMCirros() *kubevirtv1.VirtualMachine {
 			},
 		},
 		Spec: kubevirtv1.VirtualMachineSpec{
-			Running: pointer.Bool(false),
+			RunStrategy: ptr.To(kubevirtv1.RunStrategyHalted),
 			Template: &kubevirtv1.VirtualMachineInstanceTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: map[string]string{
--- a/tests/virtual_machines_test.go
+++ b/tests/virtual_machines_test.go
@@ -102,7 +102,7 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 				BeforeEach(func() {
 					By("Creating a vm with dry run mode")
 					var err error
-					vm := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br", "")}, []kubevirtv1.Network{newNetwork("br")})
+					vm := CreateVmObject(TestNamespace, []kubevirtv1.Interface{newInterface("br", "")}, []kubevirtv1.Network{newNetwork("br")})
 					createOptions := &client.CreateOptions{}
 					client.DryRunAll.ApplyToCreate(createOptions)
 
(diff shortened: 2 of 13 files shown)


tests/virtual_machines_test.go
```diff
@@ -973,18 +973,18 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 					macAddress := "02:00:ff:ff:ff:ff"
 
 					By(fmt.Sprintf("Adding a vm with a Mac Address %s in the managed namespace", macAddress))
-					vm1 := CreateVmObject(TestNamespace, false, []kubevirtv1.Interface{newInterface("br1", macAddress)}, []kubevirtv1.Network{newNetwork("br1")})
+					vm1 := CreateVmObject(TestNamespace, kubevirtv1.RunStrategyHalted, []kubevirtv1.Interface{newInterface("br1", macAddress)}, []kubevirtv1.Network{newNetwork("br1")})
 					vm1, err = testClient.VirtClient.VirtualMachine(vm1.Namespace).Create(context.TODO(), vm1)
-					Expect(err).ToNot(HaveOccurred())
+					Expect(err).ToNot(HaveOccurred(), "1")
```
orelmisan, line 978: What is the meaning of "1"?
Is it a leftover?

tests/virtual_machines_test.go
```diff
@@ -973,18 +973,18 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
 
 					err = addLabelsToNamespace(OtherTestNamespace, map[string]string{vmNamespaceOptInLabel: "ignore"})
 					Expect(err).ToNot(HaveOccurred(), "should be able to add the namespace labels")
 
 					By(fmt.Sprintf("Adding a vm with the same Mac Address %s on the unmanaged namespace", macAddress))
-					conflictingVM = CreateVmObject(OtherTestNamespace, false, []kubevirtv1.Interface{newInterface("br1", macAddress)}, []kubevirtv1.Network{newNetwork("br1")})
+					conflictingVM = CreateVmObject(OtherTestNamespace, kubevirtv1.RunStrategyHalted, []kubevirtv1.Interface{newInterface("br1", macAddress)}, []kubevirtv1.Network{newNetwork("br1")})
 					conflictingVM, err = testClient.VirtClient.VirtualMachine(conflictingVM.Namespace).Create(context.TODO(), conflictingVM)
-					Expect(err).ToNot(HaveOccurred())
+					Expect(err).ToNot(HaveOccurred(), "2")
```
orelmisan, line 987: Same.

tests/tests.go
```diff
@@ -91,11 +91,11 @@ func removeTestNamespaces() {
 		Should(BeTrue(), "Namespace %s haven't been deleted within the given timeout", TestNamespace)
 }
 
-func CreateVmObject(namespace string, running bool, interfaces []kubevirtv1.Interface, networks []kubevirtv1.Network) *kubevirtv1.VirtualMachine {
+func CreateVmObject(namespace string, strategy kubevirtv1.VirtualMachineRunStrategy, interfaces []kubevirtv1.Interface, networks []kubevirtv1.Network) *kubevirtv1.VirtualMachine {
```
orelmisan, line 94: I haven't noticed a call to this function with a `strategy` value different from `RunStrategyHalted`.
Please consider dropping the parameter.

orelmisan (changes requested):
Thank you for the PR @RamLavi.

tests/virtual_machines_test.go
RamLavi: leftover :) thanks!

tests/tests.go
RamLavi: DONE

tests/virtual_machines_test.go
RamLavi: Removed

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/5e724b8e385769c942612d2104b9253b12db2330..64fd6c36ad075064024c5106a5a42475e17a81f5): Address @orelmisan 's review

orelmisan:
@RamLavi it seems like unrelated changes had leaked into the PR.

orelmisan (approved):

tests/tests.go
```diff
@@ -91,11 +92,10 @@ func removeTestNamespaces() {
 		Should(BeTrue(), "Namespace %s haven't been deleted within the given timeout", TestNamespace)
 }
 
-func CreateVmObject(namespace string, running bool, interfaces []kubevirtv1.Interface, networks []kubevirtv1.Network) *kubevirtv1.VirtualMachine {
+func CreateVmObject(namespace string, interfaces []kubevirtv1.Interface, networks []kubevirtv1.Network) *kubevirtv1.VirtualMachine {
```
ormergi, line 95: So all tests use a "stopped" VM? I wonder if it make sense to test running VMs..
In any case its out of this PR scope.

tests/tests.go
```diff
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/pointer"
+	"k8s.io/utils/ptr"
```
ormergi, line 28: Please note the existing "k8s.io/utils/pointer" module provide the same functionality as "k8s.io/utils/pointer", adding the latter adds a lot of noise to the PR.

Adding ptr module w/o removing the pointer one make the codebase inconsistent,
The next dev may get confuse, arbitrarily pick one of them and make the code inconsistency worse.

If you intend to replace one module with another, I think it will be better doing that on a dedicated PR.

ormergi (commented):
Overall, LGTM

Please see my comment about the module changes

tests/tests.go
RamLavi: > Please note the existing "k8s.io/utils/pointer" module provide the same functionality as "k8s.io/utils/pointer",

existing "k8s.io/utils/pointer" is about converting simple data to standard params (int64, etc..)
"k8s.io/utils/ptr" is different entirely and is about converting values to pointers (To).
AFAIK They don't share function names, so there shouldn't be a problem. 

Can you elaborate?

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/1cf6fc0158e56171f56cdb42fb28982f79eb2ac1..41342ef6256fd73c05860abab1a2a2d20d68a662): rebase

tests/tests.go
ormergi: "k8s.io/utils/pointer" enable converting types to pointers, for example `pointer.Bool(myboolean), ponter.String(mystring)` .
 "k8s.io/utils/ptr" enable the same functionality implemented using generics for example: `ptr.To(mystring)` returns `*string`
In practice "k8s.io/utils/pointer"  uses  "k8s.io/utils/ptr".

See this example for replacing one with the other in k8s/k8s https://github.com/kubernetes/kubernetes/issues/130408 

WDYT?

RamLavi: ah.. then I'm all forward on doing it, but on a follow-up PR

ormergi (approved):
LGTM 

For next time, deferring non mandatory Go module changes can make PR review easier :slightly_smiling_face:

Issue #452, opened 2025-01-08
Flaky tier1: Virtual Machines Check the client When all the VMs are serviced by default (opt-out mode) and creating two VMs with same mac, one on unmanaged (opted out) namespace should detect duplicate macs gauge after restarting kubemacpool, and then mitigating that


RamLavi:
**What happened**:
Fails [here](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1973/pull-e2e-cnao-kubemacpool-functests-release-0.97/1876915297772900352)

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

RamLavi:
and [here](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1978/pull-e2e-cluster-network-addons-operator-kubemacpool-functests/1878404002324942848)

RamLavi:
I can't reproduce this locally. We will have to wait until the PR to fix logs is merged.

PR #454, opened 2025-01-15
e2e: Fix logging during test failure

RamLavi:
logs are not printed correctly as shown on https://github.com/k8snetworkplumbingwg/kubemacpool/issues/452
This PR is fixing that

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -63,7 +63,8 @@ func getPodContainerLogs(podName, containerName string) (string, error) {
 }
 
 func KubemacPoolFailedFunction(message string, callerSkip ...int) {
-	podList, err := testClient.VirtClient.CoreV1().Pods(managerNamespace).List(context.TODO(), metav1.ListOptions{})
+	podList, err := testClient.VirtClient.CoreV1().Pods(managerNamespace).List(context.TODO(),
+		metav1.ListOptions{LabelSelector: "app=kubemacpool"})
 	if err != nil {
 		fmt.Println(err)
 		Fail(message, callerSkip...)
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -1,8 +1,10 @@
 package tests
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


RamLavi:
@oshoval will help us debug the flake we saw on CNAO. PTAL when you have the time

tests/tests_suite_test.go
```diff
@@ -35,26 +37,59 @@ var _ = AfterSuite(func() {
...
+	})
+	podLogs, err := req.Stream(context.TODO())
+	if err != nil {
+		return "", err
+	}
+	defer func(podLogs io.ReadCloser) {
+		err := podLogs.Close()
+		if err != nil {
+			panic(err)
+		}
```
oshoval, line 52: do we need panic here ?
if we can refrain from it, it will be better please (maybe just log instead)

tests/tests_suite_test.go
```diff
@@ -35,26 +37,59 @@ var _ = AfterSuite(func() {
...
+	}(podLogs)
+
+	buf := new(bytes.Buffer)
+	_, err = io.Copy(buf, podLogs)
```
oshoval, line 56: why to copy it first to buf and just directly use `podLogs.String()` ?

tests/tests_suite_test.go
```diff
@@ -35,26 +37,59 @@ var _ = AfterSuite(func() {
...
+		for i := range pod.Spec.Containers {
+			containerName := pod.Spec.Containers[i].Name
+			podLogs, err := getPodLogs(pod.Name, containerName)
+			if err != nil {
+				fmt.Println(err)
+				Fail(message, callerSkip...)
+			}
+
+			fmt.Printf("\nPod container %q Logs:\n%s \n", containerName, podLogs)
+		}
```
oshoval, line 92: please consider to extract to a func

oshoval (commented):
Thanks

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/2cc0e78bd5874370876a2025019dd50f060f7c4a..83771fd37bc9d7b9e41241544ad6631003146445): Rebase

tests/tests_suite_test.go
RamLavi: DONE

tests/tests_suite_test.go
RamLavi: podLogs is an io stream interface. This means it provides a way to read data sequentially, not as a complete in-memory string. Copying it to the buffer is a way to safely move the data to a string.

tests/tests_suite_test.go
RamLavi: including the for loop?

oshoval: yep i think so, this is the core of the helper

RamLavi: DONE

oshoval (approved):

PR #505, opened 2025-03-12
[release-0.41] go.mod, Makefile: Set go version to 1.20

RamLavi:
This PR sets the go.mod version to 1.20

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ KUSTOMIZE := GOFLAGS=-mod=mod $(GO) run sigs.k8s.io/kustomize/kustomize/v4@v4.5.
 CONTROLLER_GEN := GOFLAGS=-mod=mod $(GO) run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.0
 GOFMT := GOFLAGS=-mod=mod $(GO)fmt
 VET := GOFLAGS=-mod=mod $(GO) vet
-DEEPCOPY_GEN := GOFLAGS=-mod=mod $(GO) install k8s.io/code-generator/cmd/deepcopy-gen@latest
+DEEPCOPY_GEN := GOFLAGS=-mod=mod $(GO) install k8s.io/code-generator/cmd/deepcopy-gen@v0.29.15
 GO_VERSION = $(shell hack/go-version.sh)
 
 E2E_TEST_EXTRA_ARGS ?=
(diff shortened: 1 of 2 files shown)


oshoval:
/lgtm

if you want you can add in this PR or later the guard of makefile for the go toolchain

oshoval:
/lgtm

see please comment above

RamLavi:
> /lgtm
> 
> if you want you can add in this PR or later the guard of makefile for the go toolchain

Regardless of the guard poc on CNAO, the more I think about it, it seems to me that it's something that needs to come from the CI&release's side, cause they know both ends, and don't need to statically set the go version. Doing it on our side is too hard-coded. I will talk to them about introducing such a tool

oshoval:
> > /lgtm
> > if you want you can add in this PR or later the guard of makefile for the go toolchain
> 
> Regardless of the guard poc on CNAO, the more I think about it, it seems to me that it's something that needs to come from the CI&release's side, cause they know both ends, and don't need to statically set the go version. Doing it on our side is too hard-coded. I will talk to them about introducing such a tool

once there is a stable release, and when we know the version, it is fine and required imho to limit it on our side, it doesnt change a lot / doesnt change at all on stable branches
It is static per specific stable branch so it is great imo

there are links with this info that we can create a script already / ask them for such script
but in any case the place that enforce it must be make vendor, which is the first gating imo

RamLavi:
> > > /lgtm
> > > if you want you can add in this PR or later the guard of makefile for the go toolchain
> > 
> > 
> > Regardless of the guard poc on CNAO, the more I think about it, it seems to me that it's something that needs to come from the CI&release's side, cause they know both ends, and don't need to statically set the go version. Doing it on our side is too hard-coded. I will talk to them about introducing such a tool
> 
> once there is a stable release, and when we know the version, it is fine and required imho to limit it on our side, it doesnt change a lot / doesnt change at all on stable branches It is static per specific stable branch so it is great imo
> 
> there are links with this info that we can create a script already / ask them for such script but in any case the place that enforce it must be make vendor, which is the first gating imo

I'll issue a separate PR for it once I have more answers from ci&release team.

oshoval:
np, note please that imo this is an example of case where it is good to start with something already and later improve
upon needs

PR #541, opened 2025-05-15
Add NetworkPolicies

ormergi:
In a scenario where a cluster networking has been hardened and futures
a default deny-all network policy that affects kubemacpool namespace,
kubemacpool will not be able to operate properly.

This PR introduce network-policies for kubemacpool allowing kubemacpool to operate,
complying with restrictions in form of deny all network-policy.

In order to have coverage for the introduced network-policies, and additional deny-all network-policies is installed as part of the cluster-sync flow.
Having a deny-all network-policy installed, allow the introduced network-policies to be tested in a development and test environments, locally and CI.

**Release note**:

```release-note
config: Introduce network-policies to enable kubemacpool operate in restricted env
```

### DIFF
--- a/cluster/sync.sh
+++ b/cluster/sync.sh
@@ -49,6 +49,9 @@ done
 
 ./cluster/kubectl.sh apply -f $config_dir/kubemacpool.yaml
 
+# Ensure the project network-polices are valid by installing an additional deny-all network policy affecting kubemacpool pods
+./hack/install-deny-all-net-pol.sh
+
 pods_ready_wait() {
   if [[ "$KUBEVIRT_PROVIDER" != external ]]; then
     echo "Waiting for non-kubemacpool containers to be ready ..."
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -92,6 +92,10 @@ func dumpKubemacpoolLogs(failureCount int) {
 	if err := logEndpoints(managerNamespace, names.WEBHOOK_SERVICE, failureCount); err != nil {
 		fmt.Println(err)
 	}
+
+	if err := logNetworkPolicies(managerNamespace, failureCount); err != nil {
+		fmt.Println(err)
+	}
 }
 
 func logPodContainersLogs(podName string, containers []corev1.Container, failureCount int) error {
@@ -175,5 +179,24 @@ func logPods(podsNamespace string, failureCount int) error {
 			return err
 		}
 	}
+	return nil
+}
+
+func logNetworkPolicies(namespace string, failureCount int) error {
+	npList, err := testClient.VirtClient.NetworkingV1().NetworkPolicies(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	bytes, err := json.MarshalIndent(*npList, "", "  ")
+	if err != nil {
+		return err
+	}
+
+	err = reporter.LogToFile(fmt.Sprintf("network-policies"), string(bytes), artifactDir, failureCount)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -28,7 +28,12 @@ resources:
 - rbac/rbac_role.yaml
 - rbac/rbac_role_binding.yaml
 - manager/manager.yaml
-#- ../rbac/auth_proxy_service.yaml
+# provide network-policy in case the env has restrictive network-policy in place
+- network-policy/allow-egress-dns.yaml
+- network-policy/allow-egress-kube-apiserver.yaml
+- network-policy/allow-ingress-to-kmp-service.yaml
+
+  #- ../rbac/auth_proxy_service.yaml
 #- ../rbac/auth_proxy_role.yaml
 #- ../rbac/auth_proxy_role_binding.yaml
   # If you want your controller-manager to expose the /metrics
--- a/config/default/network-policy/allow-egress-dns.yaml
+++ b/config/default/network-policy/allow-egress-dns.yaml
@@ -1,0 +1,24 @@
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-egress-to-dns
+spec:
+  podSelector:
+    matchLabels:
+      app: kubemacpool
+  policyTypes:
+  - Egress
+  egress:
+    - to:
+      - namespaceSelector:
+          matchLabels:
+            kubernetes.io/metadata.name: kube-system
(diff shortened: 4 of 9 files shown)


tests/tests_suite_test.go
```diff
@@ -55,6 +60,30 @@ var _ = JustAfterEach(func() {
 	}
 })
 
+// installTestsNetworkPolicy install NetworkPolicy delay all ingress / egress traffic in the given namespace.
+// Installing deny all network policy provides coverage for kubemacpool network-policy in e2e tests.
+func installTestsNetworkPolicy(name, ns string) error {
+	np := &netv1.NetworkPolicy{
+		ObjectMeta: metav1.ObjectMeta{Name: name},
+		Spec: netv1.NetworkPolicySpec{
+			PodSelector: metav1.LabelSelector{},
```
RamLavi, line 69: Can you limit is to only the kubemcapool namespace?

RamLavi (commented):
hmmm, if we are adding the "deny-all" on the e2e Suite, then it would probably collide with the one you'll issue on CNAO, when running it as tier1 lane on CNAO.
IOW, when running KMP tier1 on CNAO, you would end up with 2 "deny-all" policies. 

Is there an option to just include it as part of the cluster-up script?

tests/tests_suite_test.go
ormergi: The network-policy is create in kubemacpool namespaces and doesnt affect other namespaces.

ormergi:
> hmmm, if we are adding the "deny-all" on the e2e Suite, then it would probably collide with the one you'll issue on CNAO, when running it as tier1 lane on CNAO. IOW, when running KMP tier1 on CNAO, you would end up with 2 "deny-all" policies.
> 
> Is there an option to just include it as part of the cluster-up script?

I am looking at this again and it seems that creating the network-policy (NP) else where is not feasable or make the process more cumbersome, hear me out:
- cluster-up: 
  Not feasible because kubemacpool is not exist yet.
- cluster-sync: 
  Doesnt seem like a good fit to me because the NP should be created for the tests in order to verify KMP NPs,  
  I would rather avoid adding test related stuff to the cluster-sync flow.
- functest: 
  Seem like a good candidate to accommodate the deny-all NP.
  Although it not much of different then having the tests create the NP programmatically.
  Its even worse in functests because it require learning KMP namespace which is something that is already given 
  in the test suite code.

As for colliding with other futuristic deny-all NP, we can introduce some flag to the test suite to controll the deployment of the NP.
This flag can be then controlled by the Makefile functest target, allowing other CI env test KMP with or w/o the deny all NP.

WDYT?

RamLavi:
> > hmmm, if we are adding the "deny-all" on the e2e Suite, then it would probably collide with the one you'll issue on CNAO, when running it as tier1 lane on CNAO. IOW, when running KMP tier1 on CNAO, you would end up with 2 "deny-all" policies.
> > Is there an option to just include it as part of the cluster-up script?
> 
> I am looking at this again and it seems that creating the network-policy (NP) else where is not feasable or make the process more cumbersome, hear me out:
> 
> * cluster-up:
>   Not feasible because kubemacpool is not exist yet.
> * cluster-sync:
>   Doesnt seem like a good fit to me because the NP should be created for the tests in order to verify KMP NPs,
>   I would rather avoid adding test related stuff to the cluster-sync flow.
> * functest:
>   Seem like a good candidate to accommodate the deny-all NP.
>   Although it not much of different then having the tests create the NP programmatically.
>   Its even worse in functests because it require learning KMP namespace which is something that is already given
>   in the test suite code.
> 
> As for colliding with other futuristic deny-all NP, we can introduce some flag to the test suite to controll the deployment of the NP. This flag can be then controlled by the Makefile functest target, allowing other CI env test KMP with or w/o the deny all NP.
> 
> WDYT?

ah yes, I think you're right. Let's introduce a CI env to be able to control if we want to deploy the deny-all behavior

tests/tests_suite_test.go
```diff
@@ -41,10 +43,13 @@ var _ = BeforeSuite(func() {
...
+
+	Expect(installTestsNetworkPolicy(testsNetworkPolicyName, managerNamespace)).To(Succeed())
 })
 
 var _ = AfterSuite(func() {
 	removeTestNamespaces()
+	Expect(uninstallTestsNetworkPolicy(testsNetworkPolicyName, managerNamespace)).To(Succeed())
 })
 
 var _ = JustAfterEach(func() {
```
RamLavi, line 55: Can you please add in `dumpKubemacpoolLogs` func, a section logging the NetworkPolicies? It would come in handy when investigating such issues in the future.
You can use `logEndpoints` as a reference.

RamLavi (commented):
Overall looks good!
Added one comment regarding logging.

Also, can you explain on the first commit message `config: Provide network-policies manifests`, that you used the `mac-controller-manager` label to target the webhooks and `app: kubemacpool` to target both cert-manager and kubemabpool-manager pods?

ormergi:
[Changes](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/930510d282c6b9c00e8adfa74fcec6dbecd05a09..f5d13623b6486b34ea39549735b3e932b3fb608f):
- Install deny-all network-policy as part of cluster-sync flow.
  Following offline discussion with @RamLavi, we concluded installing the deny-all NP as part of sync flow is the best fit 
  for this project and project that consume the e2e tests (CNAO).
- Resolve https://github.com/k8snetworkplumbingwg/kubemacpool/pull/541#pullrequestreview-2848785498

@RamLavi PTAL

ormergi:
[Changes](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/930510d282c6b9c00e8adfa74fcec6dbecd05a09..e1bbe07f35ba1bda519fb16613cb58f6623b7a0e):
- Install deny-all network-policy as part of cluster-sync flow.
  Following offline discussion with @RamLavi, we concluded installing the deny-all NP as part of sync flow is the best fit 
  for this project and project that consume the e2e tests (CNAO).
- Resolve https://github.com/k8snetworkplumbingwg/kubemacpool/pull/541#pullrequestreview-2848785498 and https://github.com/k8snetworkplumbingwg/kubemacpool/pull/541#discussion_r2094412037

@RamLavi PTAL

cluster/sync.sh
```diff
@@ -73,3 +76,5 @@ pods_ready_wait() {
 }
 
 pods_ready_wait
+
```
RamLavi, line 79: can you remove the extra new lines?

RamLavi (commented):
looks great, added a small comment.

cluster/sync.sh
ormergi: Fixed

PR #544, opened 2025-05-29
Rm managed nps from release manifests

ormergi:
Previously #541, NPs were added to the project release manifest, including NPs for allowing access to cluster services (API and DNS).

Access to cluster API and DNS is fundamental requirement for any controller to work, allowing access to cluster services should be done by the same entity that introduce such restrictions.
Thus this PR removes the mentioned NPs from the project release manifest.

In addition, this PR simplify how network restrictions are simulated in tests:
- Attach default-deny-all along with allow access to cluster API and DNS, to the project test manifest.
- Rm hack/install-deny-all-net-pol.sh script.

This PR includes addition tiny improvement: use named port in the allow-ingress-to-metrics endpoint.

**Release note**:

```release-note
NONE
```


config/test/network-policy/allow-egress-dns.yaml
```diff
@@ -0,0 +1,25 @@
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-egress-to-dns
+  namespace: kubemacpool-system
```
RamLavi, line 6: just to keep it aligned with the others: use `namespace: system`, kustomize adds the prefix later anyways.
Same for the other yamls you added

RamLavi (approved):
/lgtm
/approve

/hold
let's unhold after we see that the probes situation is solved on CNAO

ormergi:
@RamLavi  I verified this PR on local env, all e2e tests pass using the release manifests as well.

PR #545, opened 2025-06-03
pod-pool: skip if pod has no multus network

qinqon:
This change make the pod initial looping faster by first checking if pods has no multus network (that's going to be 99% of the cases).

**Release note**:

```release-note
make kubemacpool readiness faster by checking first multus network annotation existence.
```


RamLavi (approved):
/lgtm
/approve

PR #547, opened 2025-06-22
pool-manager: Refactor full ip pool e2e tests to unit test

RamLavi:
The scenario of full mac pool is not easy to recreate in a cluster with
a big range, and it could be easily replaced by a unit test.
This PR is 
- removing e2e tests that check same KMP functionality with full ip pool scenario.
- introducing a unit test to check full ip pool scenario.
- removing the range patch reducing the MAC range in tests manifest to the original size

**Release note**:

```release-note
NONE
```

### DIFF
--- a/config/test/kubemacpool.yaml
+++ b/config/test/kubemacpool.yaml
@@ -109,7 +109,7 @@ subjects:
 ---
 apiVersion: v1
 data:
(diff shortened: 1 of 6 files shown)


ormergi (approved):
Great work! 

The third commit diff is so nice 

LGTM

PR #566, opened 2025-09-11
Support day2 range changes

RamLavi:
This PR adds a controller that watches the kubemacpool-mac-range configmap (where the ranges are set, and are extracted as env vars on the kubemacpool manager pod) and safely updates the new ranges.

In case of invalid range params inserted on day2, the controller will log the errors and will ignore the new suggested range. 

**Release note**:

```release-note
Support day2 range changes
```

### DIFF
--- a/pkg/controller/configmap/configmap_controller.go
+++ b/pkg/controller/configmap/configmap_controller.go
@@ -24,10 +24,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	"github.com/go-logr/logr"
 
 	poolmanager "github.com/k8snetworkplumbingwg/kubemacpool/pkg/pool-manager"
 )
@@ -44,6 +47,7 @@ type ConfigMapReconciler struct {
 	Scheme             *runtime.Scheme
 	PoolManager        PoolManager
 	ConfigMapNamespace string
+	EventRecorder      record.EventRecorder
 }
 
 // RangeConfig represents MAC address range configuration parsed from ConfigMap
@@ -53,6 +57,8 @@ type RangeConfig struct {
 }
 
 //+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=pods,verbs=get;list
+//+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
 
 // Add creates a new ConfigMap Controller and adds it to the Manager with the given poolManager.
 func Add(mgr manager.Manager, poolManager *poolmanager.PoolManager) error {
@@ -64,6 +70,7 @@ func Add(mgr manager.Manager, poolManager *poolmanager.PoolManager) error {
 		Scheme:             mgr.GetScheme(),
 		PoolManager:        poolManager,
 		ConfigMapNamespace: podNamespace,
+		EventRecorder:      mgr.GetEventRecorderFor("configmap-controller"),
 	}
 
 	return ctrl.NewControllerManagedBy(mgr).
@@ -94,6 +101,10 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if err != nil {
 		logger.Error(err, "Failed to extract ranges from ConfigMap",
 			"data", configMap.Data)
+
+		r.recordEvent(ctx, logger, corev1.EventTypeWarning, "InvalidRange",
+			"Invalid MAC range in ConfigMap, ignoring new range values: %v", err)
+
 		return ctrl.Result{}, fmt.Errorf("invalid ConfigMap data: %v", err)
 	}
 
@@ -121,6 +132,11 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		logger.Error(err, "Failed to update MAC address ranges in PoolManager",
 			"rangeStart", rangeConfig.Start.String(),
 			"rangeEnd", rangeConfig.End.String())
+
+		r.recordEvent(ctx, logger, corev1.EventTypeWarning, "UpdateFailed",
+			"Failed to apply new MAC ranges %s - %s: %v",
+			rangeConfig.Start.String(), rangeConfig.End.String(), err)
+
 		return ctrl.Result{}, fmt.Errorf("failed to apply new ranges: %v", err)
 	}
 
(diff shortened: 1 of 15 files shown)


RamLavi:
> Mostly nits.
> 
> One question though: how long it takes for kubemacpool to come up after a "restart" ? 

Linearly depends on the amount of VMs on the cluster. In cluster with the MAX amount of VMs allowed (10K) we measured ~80s. 

>Asking because we might be OK with not having an informer watching that particular config map - we could mount the config map into the kubemacpool pod, which would be rolled out whenever the config map data changes. IIUC it would have some downtime though - your option is probably safer.

AFAIK configmap mounted as volume does not cause pod rollout when configmap changes - it only changes the file's content after a while. 
Using this way it also possible - but watching the configmap directly is much more clean IMO.

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/e8068462185ab970fce3b1aff74364bf827517ab..d211657e298c681e0bce269e90df155ebc629d76): @maiqueb 's review fixes

RamLavi:
Known flake

/test pull-kubemacpool-e2e-k8s

pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,175 @@
...
+	if !ok {
+		return nil, fmt.Errorf("RANGE_START not found in ConfigMap")
+	}
+
+	endStr, ok := cm.Data["RANGE_END"]
+	if !ok {
+		return nil, fmt.Errorf("RANGE_END not found in ConfigMap")
+	}
+
+	// Parse and validate MAC addresses
```
qinqon, line 158: Is it worth it to add constants for RANGE_START and RANGE_END ? I see them again somewhere in the code

pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,175 @@
...
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Create predicate to only watch our specific ConfigMap
+	configMapPredicate := predicate.NewPredicateFuncs(func(object client.Object) bool {
+		return object.GetName() == names.MAC_RANGE_CONFIGMAP && object.GetNamespace() == r.ConfigMapNamespace
+	})
+
```
qinqon, line 137: Even if we do this the informer's cache is going to be filled in with all the config maps, we should limit it with cache filters at controller-runtime.

Here is an generic example how to integrate at main

```golang
package main

import (
	"context"
	"os"

	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apimachinery/pkg/runtime/schema"
	"k8s.io/apimachinery/pkg/fields"
	"k8s.io/client-go/kubernetes"
	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
	_ "k8s.io/client-go/plugin/pkg/client/auth"

	ctrl "sigs.k8s.io/controller-runtime"
	"sigs.k8s.io/controller-runtime/pkg/cache"
	"sigs.k8s.io/controller-runtime/pkg/client"
	"sigs.k8s.io/controller-runtime/pkg/log/zap"
)

var (
	scheme   = runtime.NewScheme()
)

func init() {
	_ = clientgoscheme.AddToScheme(scheme)
	// Add other APIs to the scheme if needed
	// _ = myv1alpha1.AddToScheme(scheme)
}

func main() {
	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))

	// Define the namespace and name you want to filter for
	const namespace = "my-filtered-namespace"
	const configMapName = "my-special-configmap"

	// Create a new field selector to filter by name
	nameFieldSelector := fields.OneTermEqualSelector("metadata.name", configMapName)

	// Configure the cache options with ByObject
	cacheOptions := cache.Options{
		// Namespaces can be used for a list of namespaces to watch,
		// or use cache.MultiNamespacedCacheBuilder for more granular control.
		// For this example, we use a single namespace.
		// NOTE: ByObject.Namespaces overrides this.
		DefaultNamespaces: map[string]cache.Config{
			namespace: {}, // Empty config to indicate we want to cache this namespace
		},
		ByObject: map[client.Object]cache.ByObject{
			// Filter ConfigMaps by namespace and name.
			&corev1.ConfigMap{}: {
				// We can define the list of namespaces here, which is more powerful.
				// In this case, we only want to watch the one namespace.
				Namespaces: map[string]cache.Config{
					namespace: {},
				},
				// Use the Field selector to filter by the name
				Field: nameFieldSelector,
			},
			// You can add more objects to filter here.
			// &corev1.Pod{}: {
			//     Namespaces: map[string]cache.Config{
			//         namespace: {},
			//     },
			//     Field: fields.OneTermEqualSelector("metadata.ownerReferences", "my-owner-reference-uid"),
			// },
		},
	}

	// Create a new manager with the configured cache
	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
		Scheme: scheme,
		Cache:  cacheOptions,
	})
	if err != nil {
		ctrl.Log.Error(err, "unable to start manager")
		os.Exit(1)
	}

	// Now you can build your controller and reconciler, and it will use the filtered cache.
	// For example, to set up a controller that watches ConfigMaps:
	// err = (&MyReconciler{
	// 	Client: mgr.GetClient(),
	// 	Scheme: mgr.GetScheme(),
	// }).SetupWithManager(mgr)
	// if err != nil {
	// 	ctrl.Log.Error(err, "unable to create controller", "controller", "MyController")
	// 	os.Exit(1)
	// }

	// Start the manager
	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
		ctrl.Log.Error(err, "problem running manager")
		os.Exit(1)
	}
}
```

pkg/pool-manager/pool.go
```diff
@@ -184,7 +176,7 @@ func checkRange(startMac, endMac net.HardwareAddr) error {
 		}
 	}
 
-	return fmt.Errorf("Invalid range. rangeStart: %s rangeEnd: %s", startMac.String(), endMac.String())
+	return fmt.Errorf("invalid range. rangeStart: %s rangeEnd: %s", startMac.String(), endMac.String())
```
qinqon, line 179: Boy scout mode: maybe instead of iterating manually we can use the bytes.Compare 
```golang
bytes.Compare(startMac, endMac)
```

pkg/pool-manager/pool.go
```diff
@@ -405,3 +401,78 @@ func (p *PoolManager) getOptMode(mutatingWebhookConfigName, webhookName string)
...
+func (p *PoolManager) GetCurrentRanges() (start, end net.HardwareAddr) {
+	p.rangeMutex.RLock()
+	defer p.rangeMutex.RUnlock()
+
+	// Return copies to avoid external modification
+	start = make(net.HardwareAddr, len(p.rangeStart))
+	end = make(net.HardwareAddr, len(p.rangeEnd))
+	copy(start, p.rangeStart)
+	copy(end, p.rangeEnd)
+
```
qinqon, line 459: Do we really need to copy it ? from what I see in the code we only use this at reconciler and we directly compare it as string, so maybe we just need to return the string representation of these macs instead of copy them.

pkg/pool-manager/pool.go
```diff
@@ -405,3 +401,78 @@ func (p *PoolManager) getOptMode(mutatingWebhookConfigName, webhookName string)
...
+// getCurrentMAC returns the current MAC address with proper locking
+func (p *PoolManager) getCurrentMAC() net.HardwareAddr {
+	p.rangeMutex.RLock()
+	defer p.rangeMutex.RUnlock()
+
+	// Return a copy to avoid external modification
+	currentMac := make(net.HardwareAddr, len(p.currentMac))
+	copy(currentMac, p.currentMac)
+
```
qinqon, line 471: Ditto: maybe we just need to return the string representation

qinqon (changes requested):
Nothing big

pkg/controller/configmap/configmap_controller.go
RamLavi: ACK, thanks!

pkg/controller/configmap/configmap_controller.go
RamLavi: Let's do it on a follow up PR

pkg/pool-manager/pool.go
RamLavi: I agree. fixed

pkg/pool-manager/pool.go
RamLavi: yep..

RamLavi (commented):
@qinqon pushed the fixes, PTAL

qinqon (commented):
/lgtm

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/566/commits/768c0dfb75b20c13faec7876f7fa7a7fd7226c35): added events on the kmp manager pod in case of success/failure to update range.

@qinqon PTAL

pkg/controller/configmap/configmap_controller.go
maiqueb: nice suggestion.

PR #611, opened 2026-03-29
tests: Add dry run VMI check after cert recovery

RamLavi:
**What this PR does / why we need it**:
The certificate recovery tests (certificates_test.go) verify that a deleted secret/caBundle gets regenerated, but do not wait for the webhook server to actually reload and serve the new certificate. 

Due to kubelet secret volume propagation delay (~60s), subsequent tests can hit "x509: certificate signed by unknown authority" if they call the webhook before the new cert is in place.

This PR adds a probe in checkCertLibraryRecovery that creates a VM via the webhook and retries until it succeeds, ensuring the new certificate is fully operational before the test completes.

**Special notes for your reviewer**:

**Release note**:
```release-note
NONE
```

PR #612, opened 2026-03-29
Swap pod MAC collision webhook block with collision detection

RamLavi:
This PR introduces a pod MAC collision detection controller, that will replace the current pod  MAC collision rejection webhook behavior.
It introduces a controller for pod collisions, and also handles cross-pod-vmi coliision cases.
In order to keep consistency between the cross-collision objects, a secondary watch is added, along with e2e that cover all above scenarios.

**Release note**:
```release-note
Remove pod create/update rejection behavior when colliding MAC
Add pod MAC collision detection controller
```

### DIFF
--- a/pkg/controller/maccollision/cache.go
+++ b/pkg/controller/maccollision/cache.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package vmicollision
+package maccollision
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -23,7 +23,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-var cacheLog = logf.Log.WithName("VMICollision Cache")
+var cacheLog = logf.Log.WithName("MACCollision Cache")
 
 const (
 	// MacAddressIndexName is the index name for MAC address lookups
--- a/pkg/controller/maccollision/cache.go
+++ b/pkg/controller/maccollision/cache.go
@@ -17,10 +17,15 @@ limitations under the License.
 package maccollision
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	networkv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	netutils "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/utils"
 )
 
 var cacheLog = logf.Log.WithName("MACCollision Cache")
@@ -28,6 +33,9 @@ var cacheLog = logf.Log.WithName("MACCollision Cache")
 const (
 	// MacAddressIndexName is the index name for MAC address lookups
 	MacAddressIndexName = "status.interfaces.mac"
+
+	// PodMacAddressIndexName is the index name for Pod MAC address lookups.
+	PodMacAddressIndexName = "annotations.networks.mac"
 )
 
 // StripVMIForCollisionDetection keeps only:
@@ -80,4 +88,74 @@ func IndexVMIByMAC(obj client.Object) []string {
 	}
 
 	return macs
+}
+
+// StripPodForCollisionDetection keeps only fields needed for collision detection,
+// reducing memory for cached Pod objects.
+func StripPodForCollisionDetection(obj interface{}) (interface{}, error) {
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		return obj, nil
+	}
+
+	stripped := &corev1.Pod{
+		TypeMeta: pod.TypeMeta,
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        pod.Name,
+			Namespace:   pod.Namespace,
+			UID:         pod.UID,
+			Labels:      pod.Labels,
+			Annotations: pod.Annotations,
+		},
+		Status: corev1.PodStatus{
+			Phase: pod.Status.Phase,
+		},
+	}
+
+	return stripped, nil
+}
+
+// IndexPodByMAC returns all requested MAC addresses from a Pod's multus
+// network-attachment annotation for indexing.
+// Returns nil if multus has not yet processed the Pod (no network-status annotation).
+func IndexPodByMAC(obj client.Object) []string {
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		return nil
+	}
+
+	if _, hasStatus := pod.Annotations[networkv1.NetworkStatusAnnot]; !hasStatus {
+		return nil
+	}
+
+	networks, err := netutils.ParsePodNetworkAnnotation(pod)
+	if err != nil {
+		return nil
+	}
+
+	seen := sets.New[string]()
+	var macs []string
+	for _, net := range networks {
+		if net.MacRequest == "" {
+			continue
+		}
+		normalizedMAC, err := NormalizeMacAddress(net.MacRequest)
+		if err != nil {
+			cacheLog.Error(err, "failed to normalize MAC address", "mac", net.MacRequest, "pod", pod.Name, "namespace", pod.Namespace)
+			continue
+		}
+		if seen.Has(normalizedMAC) {
+			continue
+		}
+		seen.Insert(normalizedMAC)
+		macs = append(macs, normalizedMAC)
+	}
+
+	return macs
+}
+
+// IsKubevirtOwned returns true if the Pod is a virt-launcher Pod.
+// These Pods' MACs are already tracked through the VMI collision controller.
+func IsKubevirtOwned(pod *corev1.Pod) bool {
+	return pod.Labels[kubevirtv1.AppLabel] == "virt-launcher"
 }
--- a/pkg/controller/maccollision/pod_candidates.go
+++ b/pkg/controller/maccollision/pod_candidates.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2025 The KubeMacPool Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package maccollision
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func listRunningPodsWithMAC(ctx context.Context, k8sClient client.Client, normalizedMAC string) ([]*corev1.Pod, error) {
+	podList := &corev1.PodList{}
+	if err := k8sClient.List(ctx, podList, client.MatchingFields{PodMacAddressIndexName: normalizedMAC}); err != nil {
+		return nil, errors.Wrap(err, "failed to list Pods by MAC")
+	}
+
+	var runningPods []*corev1.Pod
+	for i := range podList.Items {
+		p := &podList.Items[i]
+		if p.Status.Phase != corev1.PodRunning {
+			continue
+		}
+		runningPods = append(runningPods, p)
+	}
+
+	return runningPods, nil
+}
+
+func excludePodUID(pods []*corev1.Pod, excludeUID types.UID) []*corev1.Pod {
+	var filtered []*corev1.Pod
+	for _, p := range pods {
+		if p.UID == excludeUID {
+			continue
+		}
+		filtered = append(filtered, p)
+	}
+	return filtered
+}
+
+func filterCollisionCandidatePods(pods []*corev1.Pod, reconciledNamespace string, poolManager PoolManagerInterface, logger logr.Logger) []*corev1.Pod {
+	managedNamespaces := sets.New[string](reconciledNamespace)
+
+	var filtered []*corev1.Pod
+	for _, p := range pods {
+		if IsKubevirtOwned(p) {
+			logger.V(1).Info("Filtering out kubevirt-owned pod", "pod", p.Name, "namespace", p.Namespace)
+			continue
+		}
+
+		if managedNamespaces.Has(p.Namespace) {
+			filtered = append(filtered, p)
+			continue
+		}
+
+		managed, err := poolManager.IsPodManaged(p.Namespace)
+		if err != nil {
+			logger.Error(err, "Failed to check if pod namespace is managed, including in collisions",
+				"namespace", p.Namespace, "pod", p.Name)
+			filtered = append(filtered, p)
+			managedNamespaces.Insert(p.Namespace)
+			continue
+		}
+
+		if !managed {
+			logger.V(1).Info("Filtering out pod from unmanaged namespace",
+				"namespace", p.Namespace, "pod", p.Name)
+			continue
+		}
+
+		managedNamespaces.Insert(p.Namespace)
+		filtered = append(filtered, p)
+	}
+
+	return filtered
+}
--- a/pkg/controller/maccollision/pod_controller.go
+++ b/pkg/controller/maccollision/pod_controller.go
@@ -88,8 +88,45 @@ func SetupPodControllerWithManager(mgr manager.Manager, poolManager *pool_manage
 		return fmt.Errorf("failed to watch Pods: %w", err)
 	}
 
+	if r.poolManager.IsKubevirtEnabled() {
+		err = c.Watch(
+			source.Kind(
+				mgr.GetCache(),
+				&kubevirtv1.VirtualMachineInstance{},
+				handler.TypedEnqueueRequestsFromMapFunc(r.mapVMIToPods),
+				collisionRelevantChanges(),
+			),
+		)
+		if err != nil {
+			return fmt.Errorf("failed to watch VMIs for Pod cross-type collisions: %w", err)
+		}
+	}
+
 	podLog.Info("Successfully registered MAC collision Pod controller")
 	return nil
+}
+
+func (r *PodReconciler) mapVMIToPods(ctx context.Context, vmi *kubevirtv1.VirtualMachineInstance) []reconcile.Request {
+	macs := IndexVMIByMAC(vmi)
+
+	var requests []reconcile.Request
+	seen := map[types.NamespacedName]struct{}{}
+	for _, mac := range macs {
+		var podList corev1.PodList
+		if err := r.List(ctx, &podList, client.MatchingFields{PodMacAddressIndexName: mac}); err != nil {
+			podLog.Error(err, "failed to list Pods by MAC for cross-type enqueue", "mac", mac)
+			continue
+		}
+		for i := range podList.Items {
+			key := types.NamespacedName{Namespace: podList.Items[i].Namespace, Name: podList.Items[i].Name}
+			if _, ok := seen[key]; ok {
+				continue
+			}
+			seen[key] = struct{}{}
+			requests = append(requests, reconcile.Request{NamespacedName: key})
+		}
+	}
+	return requests
 }
 
 // Reconcile handles Pod reconciliation for collision detection.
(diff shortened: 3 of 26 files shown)


RamLavi:
/hold

still running local tests to make sure we don't introduce any flakes

pkg/controller/maccollision/vmi_controller.go
```diff
@@ -294,21 +337,43 @@ func (r *VMICollisionReconciler) filterOutUnmanagedNamespaces(collisionCandidate
...
+				"namespace", p.Namespace, "pod", p.Name)
+			continue
+		}
+
+		runningPods = append(runningPods, p)
 	}
-	return collisions
+
+	return runningPods, nil
 }
```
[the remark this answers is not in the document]
RamLavi: [gemini-code-assist](https://github.com/apps/gemini-code-assist) I fixed. please re-review

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/ed5d8ee794cedfcdc0423a7e811547ebc4eb03a0..b2f79e325e5bdeb29ca2b497541795e56ba8d2ef): Address gemini review and lint issues,

RamLavi:
Change: Refactor commit message

RamLavi:
/hold 
until https://github.com/k8snetworkplumbingwg/kubemacpool/pull/611 is merged

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/af00f3d5344d980ec2d8333c0f86c66adb3c5524..c55432d2308d29844df46f5ff39b035bf92a6415): use dedicated opt in function to avoid opt in/out race flakes

RamLavi:
FYI e2e lane ran 200 times on local with no issues.

pkg/controller/maccollision/vmi_controller.go
```diff
@@ -294,21 +282,16 @@ func (r *VMICollisionReconciler) filterOutUnmanagedNamespaces(collisionCandidate
...
-			refs[i] = vmiObjectRef(v.Namespace, v.Name)
-		}
-		collisions[mac] = refs
+func (r *VMIReconciler) findRunningPodsWithMAC(ctx context.Context, normalizedMAC string, logger logr.Logger) ([]*corev1.Pod, error) {
+	runningPods, err := listRunningPodsWithMAC(ctx, r.Client, normalizedMAC)
+	if err != nil {
+		return nil, err
 	}
-	return collisions
+	return filterCollisionCandidatePods(runningPods, "", r.poolManager, logger), nil
```
qinqon, line 290: Passing `""` as `reconciledNamespace` means the VMI's own namespace is *not* seeded into the managed-namespaces cache inside `filterCollisionCandidatePods`, so every Pod's namespace will trigger an `IsPodManaged()` call — including Pods in the same namespace as the VMI being reconciled (which is already known to be managed).

Consider accepting the VMI namespace as a parameter and forwarding it here:
```suggestion
	return filterCollisionCandidatePods(runningPods, vmiNamespace, r.poolManager, logger), nil
```
This would require adding `vmiNamespace string` to the method signature and passing `vmi.Namespace` from `checkMACCollisions`.

pkg/controller/maccollision/pod_candidates.go
```diff
@@ -0,0 +1,79 @@
+package maccollision
```
qinqon, line 17: Missing the Apache 2.0 license header that all other files in the package have. Same applies to `vmi_candidates.go`.

pkg/controller/maccollision/predicates.go
```diff
@@ -101,3 +104,32 @@ func getTargetMigrationUID(vmi *kubevirtv1.VirtualMachineInstance) string {
 	}
 	return ""
 }
+
+func podCollisionRelevantChanges() predicate.TypedPredicate[*corev1.Pod] {
+	return predicate.TypedFuncs[*corev1.Pod]{
+		CreateFunc: func(e event.TypedCreateEvent[*corev1.Pod]) bool {
+			return true
```
qinqon: nit: Returning `true` for every Pod create means *all* Pod creations (including pods without multus annotations) will trigger a reconcile. The reconciler handles this gracefully (early return when no MACs found), but in clusters with many non-multus pods this could be noisy.

Consider filtering here for pods that have the `k8s.v1.cni.cncf.io/networks` annotation:
```go
CreateFunc: func(e event.TypedCreateEvent[*corev1.Pod]) bool {
    _, hasNetworks := e.Object.GetAnnotations()[networkv1.NetworkAttachmentAnnot]
    return hasNetworks
},
```
Not a blocker — the reconciler is already safe without this.

qinqon (changes requested):

ormergi (commented):
Thanks for posting

This change seem is quite big changing the project fundamental way of detecting conflicts or preventing them in the fist place via the webhook.
Although there is no existing mechanism yet, please consider gating this functionality to allow fallback in case of major issues.

It seems the new controller that handle pods, relay on the fact Kubevirt pods are labeled with `kubevirt.io=virt-launcher`.
I am not sure its formal API that could be eliminated in the future, we should verify if its legit for our use-case.
Alternatively the controller could look for owner-reference of the parent VirtualMachine object. To me it sounds more stable.

RamLavi:
> Thanks for posting
> 
> This change seem is quite big changing the project fundamental way of detecting conflicts or preventing them in the fist place via the webhook. Although there is no existing mechanism yet, please consider gating this functionality to allow fallback in case of major issues.

I agree that it's a big change.
Regarding gating, I'm afraid the time schedule that was forced upon us makes it that we must deliver this (meaning the VMI collision) as GA, otherwise decentralized migration will be broken on arival.
Note that this particualr change is only a complimentary change to the already existing changes on the VMI side. The community was already informed and we got no NACKs.

> 
> It seems the new controller that handle pods, relay on the fact Kubevirt pods are labeled with `kubevirt.io=virt-launcher`. I am not sure its formal API that could be eliminated in the future, we should verify if its legit for our use-case. Alternatively the controller could look for owner-reference of the parent VirtualMachine object. To me it sounds more stable.

That's a valid concern. @omisan - the kubemacpoool pod controller assumes that a virt-launcher pod is a pod that has the `kubevirt.io=virt-launcher` label, same as kubevirt/kubevirt code uses that. Is that a good assumption, or should we use something else, like owner reference?

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/c55432d2308d29844df46f5ff39b035bf92a6415..85c637401767706783f1257d348523bde954ff01): Rebase

pkg/controller/maccollision/vmi_controller.go
RamLavi: I don’t think passing vmi.Namespace here is correct.
In this path we know the namespace is VMI-managed, but filterCollisionCandidatePods() intentionally applies Pod-managed filtering. Those are different checks in kubemacpool (IsVirtualMachineManaged() vs IsPodManaged() - they have different labels..), so seeding the cache with vmi.Namespace would implicitly (and wrongly) assume that a VM-managed namespace is also pod-managed.

pkg/controller/maccollision/pod_candidates.go
RamLavi: DONE

pkg/controller/maccollision/predicates.go
RamLavi: you're right, DONE

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/85c637401767706783f1257d348523bde954ff01..7e46676db386320d7a7a28b726363817cad4300f): Address @qinqon 's comments

/hold cancel 

https://github.com/k8snetworkplumbingwg/kubemacpool/pull/611 merged, rebased on top of it.

pkg/controller/maccollision/pod_controller.go
```diff
@@ -98,19 +105,147 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
...
+			continue
+		}
+
+		isManaged, err := r.poolManager.IsPodManaged(p.Namespace)
+		if err != nil {
+			logger.Error(err, "Failed to check if pod namespace is managed, including in collisions",
+				"namespace", p.Namespace, "pod", p.Name)
+			filtered = append(filtered, p)
+			managedNamespaces[p.Namespace] = struct{}{}
+			continue
```
ormergi: This doesnt look safe, it may miss conflicting pods/vms. 
WDYT?

RamLavi: hmm I think the risk here is the opposite: this path is fail-open, so on IsPodManaged() error the `continue` means that we **include** the candidate rather than skip it. That can lead to temporary false positives, but it should not miss real conflicts. I chose that intentionally for candidate objects, so a transient managed-namespace lookup failure would not hide a collision. For the reconciled object itself we still fail and requeue on namespace-check errors.

pkg/controller/maccollision/cache.go
```diff
@@ -81,3 +88,73 @@ func IndexVMIByMAC(obj client.Object) []string {
...
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		return nil
+	}
+
+	if _, hasStatus := pod.Annotations[networkv1.NetworkStatusAnnot]; !hasStatus {
+		return nil
+	}
+
+	networks, err := netutils.ParsePodNetworkAnnotation(pod)
```
qinqon, line 131: There is apimachinery type for sets "k8s.io/apimachinery/pkg/util/sets", we can use it

pkg/controller/maccollision/cache.go
```diff
@@ -81,3 +88,73 @@ func IndexVMIByMAC(obj client.Object) []string {
...
+			continue
+		}
+		if _, dup := seen[normalizedMAC]; dup {
+			continue
+		}
+		seen[normalizedMAC] = struct{}{}
+		macs = append(macs, normalizedMAC)
+	}
+
+	return macs
```
qinqon, line 154: Instead of this maybe we can configure the controller runtime manager cache to skip pods with this label, so they informer directly ignore them instead of filter them up

pkg/controller/maccollision/cache.go
```diff
@@ -81,3 +88,73 @@ func IndexVMIByMAC(obj client.Object) []string {
...
+		return nil
+	}
+
+	seen := map[string]struct{}{}
+	var macs []string
+	for _, net := range networks {
+		if net.MacRequest == "" {
+			continue
+		}
+		normalizedMAC, err := NormalizeMacAddress(net.MacRequest)
```
qinqon, line 142: I think this is not true, two pods having the same mac but at different multus network is ok, i think we should index by network name too

pkg/controller/maccollision/cache.go
```diff
@@ -81,3 +88,73 @@ func IndexVMIByMAC(obj client.Object) []string {
 
 	return macs
```
qinqon, line 90: Maybe we can just use metav1.PartialObjectMetadata and retrieve Phase when needed

pkg/controller/maccollision/pod_controller.go
```diff
@@ -0,0 +1,116 @@
...
+		logger.V(1).Info("Pod is virt-launcher, skipping (tracked via VMI controller)")
+		return reconcile.Result{}, nil
+	}
+
+	isManaged, err := r.poolManager.IsPodManaged(pod.Namespace)
+	if err != nil {
+		logger.Error(err, "Failed to check if namespace is managed")
+		return reconcile.Result{}, errors.Wrapf(err, "failed to check if namespace %s is managed", pod.Namespace)
+	}
+	if !isManaged {
```
qinqon, line 160: This should be done using controller runtime filtering mechanism

pkg/controller/maccollision/pod_controller.go
```diff
@@ -0,0 +1,251 @@
...
+		logger.V(1).Info("Pod namespace not managed by kubemacpool, skipping")
+		return reconcile.Result{}, nil
+	}
```
qinqon, line 163: As mention we have to filter by network too, also you can use the apimachinery set for this kind of things

pkg/controller/maccollision/pod_controller.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+
+		collidingPods = r.filterPodsForCollision(collidingPods, pod.Namespace, logger)
+
+		if len(collidingPods) > 0 {
+			var refs []pool_manager.ObjectReference
+			for _, p := range collidingPods {
+				refs = append(refs, podObjectRef(p.Namespace, p.Name))
+			}
+
```
qinqon, line 208: This should be done at reconcile cache filtering level, so we never reconcile VM pods

pkg/controller/maccollision/predicates.go
```diff
@@ -101,3 +104,33 @@ func getTargetMigrationUID(vmi *kubevirtv1.VirtualMachineInstance) string {
 	}
 	return ""
 }
+
+func podCollisionRelevantChanges() predicate.TypedPredicate[*corev1.Pod] {
+	return predicate.TypedFuncs[*corev1.Pod]{
+		CreateFunc: func(e event.TypedCreateEvent[*corev1.Pod]) bool {
+			_, hasNetworks := e.Object.GetAnnotations()[networkv1.NetworkAttachmentAnnot]
+			return hasNetworks
```
qinqon, line 112: Should we skip if they have multus annotation but just for default pod network ? for example at ovn-kuberntes it has one entry but is for cluster default network, still Reconcile will kick in

qinqon (changes requested):
First pass

qinqon:
@RamLavi I see that 99% of what we do with VMIs is more or less the same thing that we do with pods + multus annotations, maybe we should just have one pod's controller that trait VMs as pods and add some extra stuff that reads VMIM for cross cluster live migration ? It will simplify kubemacpool a lot.

RamLavi:
> @RamLavi I see that 99% of what we do with VMIs is more or less the same thing that we do with pods + multus annotations, maybe we should just have one pod's controller that trait VMs as pods and add some extra stuff that reads VMIM for cross cluster live migration ? It will simplify kubemacpool a lot.

This is an interesting thought as a follow up. We need to make sure all VMI use cases can be safely checked by using pods. Or maybe even fetch the VMIs in case the pod is a virt-launcher pod.. Anyways it needs a thorough thought.

pkg/controller/maccollision/cache.go
RamLavi: Well, I think this is what KMP currently does on the webhook. 
We only use the network name in order to exclude self rejections..
hrmm, changing that is out of this PR scope I'm afraid

ormergi:
Please consider having single controller watching pods and VMs, so that it has single entrypoint that acts according to the request object.
This approach eliminates the need to synchronize between the two distinct controllers reducing point of failures and races, while enabling modular and separated handler entities for each type.

pkg/controller/maccollision/cache.go
RamLavi: I'm not sure it's a good idea, cause it's a manager-wide change that also affects the old pod controller.

I'm not 100% sure we can't do that, but I prefer to do it on a follow up PR, so that we can take all things into account.

pkg/controller/maccollision/cache.go
RamLavi: The old webhook rejected same-MAC-different-network. 
I'm not saying it's the most accurate thing to do, but we shouldn't change the behavior on this scope. 
This PR is only about moving the MAC collision logic outcome from webhook reject to a allow-but-issue-event+alert...

pkg/controller/maccollision/cache.go
RamLavi: We need to check phase on _every_ pod reconcile, and we need to act differently according to it (**store** MAC collision in case of  Running and **releasing** otherwise), so I'm afraid that field is crucial for the controller..

pkg/controller/maccollision/pod_controller.go
RamLavi: We can do that, but I kinda feel that predicates should work fast. This check however is anything but fast (IsPodManaged does a webhook config lookup + namespace Get). 
In Reconcile, it runs in parallel worker goroutines.

The real solution to this is adding a watch for namespace & webhook opt mode changes. But that's yet another tech debt that is out of this PR's scope...

pkg/controller/maccollision/pod_controller.go
RamLavi: Same as comment [above](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/612#discussion_r3086223587)

pkg/controller/maccollision/pod_controller.go
RamLavi: same as [above](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/612#discussion_r3086205637)

pkg/controller/maccollision/predicates.go
RamLavi: The predicate already checks for `k8s.v1.cni.cncf.io/networks` (=network request annotation), not `k8s.v1.cni.cncf.io/network-status` (=status annotation). 
IIUC - pods that only use the cluster default network don't have the networks request annotation - that annotation is specifically for requesting secondary network attachments. So in an OVN-Kubernetes setup, pods with only the default network wouldn't pass the predicate and wouldn't trigger a reconcile.

Right?

pkg/controller/maccollision/cache.go
RamLavi: DONE both here and on `filterCollisionCandidatePods()`

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/7e46676db386320d7a7a28b726363817cad4300f..1fc9bef45626ba2809e10d30e339e54b8c5f484f): use sets

RamLavi:
> Please consider having single controller watching pods and VMs, so that it has single entrypoint that acts according to the request object. This approach eliminates the need to synchronize between the two distinct controllers reducing point of failures and races, while enabling modular and separated handler entities for each type.

This approach of using one controller catching both pod and VMI object types may work (each reconcile checks if there is maybe a pod or a VMI with the ns and name, and treating both just in case, to avoid missing dedupped events) MAY work, but I feel the traditional way is more bullet proof.

Also, the approach suggested by @qinqon of using one controller watching **only pods**, and using existing cache to index virt-launcher pods their vmis - could work. We could also use VMIM informers to have insights on CCLM to avoid false-fires.. 


But, in the end - I am not sure it's better. 
The complexity added by having 2 controllers with sub-watches, is a pure controller-runtime infra complexity -which is well-tested, declarative, lives in the framework - not in our kubemacpool core logic.  
Moving to one controller would inevitably make the kubemacpool core logic more complex, cause we would have to move the logic of "is this pod a vmi?" (and more..) - to our unified controller logic.

So splitting the 2 cases seems like a good "separate of concerns" approach to me.

qinqon:
> > Please consider having single controller watching pods and VMs, so that it has single entrypoint that acts according to the request object. This approach eliminates the need to synchronize between the two distinct controllers reducing point of failures and races, while enabling modular and separated handler entities for each type.
> 
> This approach of using one controller catching both pod and VMI object types may work (each reconcile checks if there is maybe a pod or a VMI with the ns and name, and treating both just in case, to avoid missing dedupped events) MAY work, but I feel the traditional way is more bullet proof.
> 
> Also, the approach suggested by @qinqon of using one controller watching **only pods**, and using existing cache to index virt-launcher pods their vmis - could work. We could also use VMIM informers to have insights on CCLM to avoid false-fires..
> 
> But, in the end - I am not sure it's better. The complexity added by having 2 controllers with sub-watches, is a pure controller-runtime infra complexity - not kubemacpool logic complexity. Moving to one controller will inevitably make the kubemacpool pure logic more complex, cause we would have to move the logic of "is this pod a vmi?" and more - to our unified controller logic.
> 
> So splitting the 2 cases seems like a good "separate of concerns" approach to me.

Well I don't think this is the case, it merits a PoC to see how complexity looks, and how much pure kubevirt logic we really needs, in fact what we need to list is what data the pure pod is not giving us to have kubevirt support, that's the important part.

Anyhow not for this PR.

/lgtm
/approve

orelmisan:
> > Thanks for posting
> > This change seem is quite big changing the project fundamental way of detecting conflicts or preventing them in the fist place via the webhook. Although there is no existing mechanism yet, please consider gating this functionality to allow fallback in case of major issues.
> 
> I agree that it's a big change. Regarding gating, I'm afraid the time schedule that was forced upon us makes it that we must deliver this (meaning the VMI collision) as GA, otherwise decentralized migration will be broken on arival. Note that this particualr change is only a complimentary change to the already existing changes on the VMI side. The community was already informed and we got no NACKs.
> 
> > It seems the new controller that handle pods, relay on the fact Kubevirt pods are labeled with `kubevirt.io=virt-launcher`. I am not sure its formal API that could be eliminated in the future, we should verify if its legit for our use-case. Alternatively the controller could look for owner-reference of the parent VirtualMachine object. To me it sounds more stable.
> 
> That's a valid concern. @Omisan - the kubemacpoool pod controller assumes that a virt-launcher pod is a pod that has the `kubevirt.io=virt-launcher` label, same as kubevirt/kubevirt code uses that. Is that a good assumption, or should we use something else, like owner reference?

Disclaimer: I haven't reviewed the PR.
It would make sense to identify virt-launcher pods using the same mechanism KubeVirt uses [1], as it is unlikely to change in an unexpected way.
Keep in mind that it can be deprecated in the future.

https://github.com/kubevirt/kubevirt/blob/621e571727540bb0c62d87d1bddc9e378328095c/pkg/virt-api/webhooks/validating-webhook/admitters/pod-eviction-admitter.go#L94

PR #682, opened 2026-08-18
Bump go 1.26

ormergi:
 This PR bumps go to 1.26

Along many improvements, newer Go version support additional TLS groups for post-quantum cryptography (PQC) readiness [1].
PQC readiness is crucial for our d/s product (Openshift).

[1] https://go.dev/doc/go1.26#:~:text=crypto/tls,post-quantum%20key%20exchanges

**Release note**:

```release-note
vendor: bump go to 1.26
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,9 +1,9 @@
 module github.com/k8snetworkplumbingwg/kubemacpool
 
-// allowed_go 1.25
-go 1.25.0
+// allowed_go 1.26
+go 1.26.0
 
-toolchain go1.25.12
+toolchain go1.26.6
 
 require (
 	github.com/go-logr/logr v1.4.3


go.mod
```diff
@@ -1,9 +1,9 @@
 module github.com/k8snetworkplumbingwg/kubemacpool
 
-// allowed_go 1.25
-go 1.25.0
+// allowed_go 1.26
+go 1.26.0
 
-toolchain go1.25.12
+toolchain go1.26.6
```
oshoval, line 6: please make sure D/S uses >= .6
otherwise it will break it

ormergi: Mote that the minimal version is set to 1.26 
D/S  products could specify their own toolchain for overriding what we use here (1.26.6)

oshoval: D/S is using GO_TOOLCHAIN=local
and since go.mod has `toolchain`, the D/S local toolchain must be >= 1.26.6 else it will break

oshoval:
/lgtm cancel

please see the comment above, this can break D/S, depends on the available GO version

go.mod
oshoval: D/S has go version go1.26.5 (Red Hat 1.26.5-2.el9_8) linux/amd64
the PR will break it

you can use .5 if you want

oshoval: checking deeper with you offline,
seems `go 1.26.0` is the ones matter, and the go toolchain is just a suggestion, depends on GOTOOLCHAIN,

so you can proceed once Ram is back,
worst case something fail if we are wrong and we will fix

go.mod
```diff
@@ -1,9 +1,9 @@
 module github.com/k8snetworkplumbingwg/kubemacpool
 
-// allowed_go 1.25
-go 1.25.0
+// allowed_go 1.26
+go 1.26.0
```
oshoval, line 4: worth to ask @RamLavi 
if we prefer to have the go directive aligned to the toolchain or not

ormergi:
From U/S prespective, the project Go verision should not be bound by D/S product build process.
This PR sets minmal go version to 1.26, and also toolchain to use 1.26.6.
The tool chain can be overridden by D/S products with the standard GOTOOLCHAIN var.

From what I tested, using older Z version doesnt break:
 ```
$ go version
go version go1.26.5 linux/amd64
$ head 10 go.mod
head: cannot open '10' for reading: No such file or directory
==> go.mod <==
module github.com/kubevirt/ipam-extensions

go 1.26

toolchain go1.26.6

$ GOTOOLCHAIN=local go build cmd/manager/main.go
$ echo $?
0

oshoval:
> From U/S prespective, the project Go verision should not be bound by D/S product build process. This PR sets minmal go version to 1.26, and also toolchain to use 1.26.6. The tool chain can be overridden by D/S products with the standard GOTOOLCHAIN var.
> 
> From what I tested, using older Z version doesnt break:
> 
> ```
> $ go version
> go version go1.26.5 linux/amd64
> $ head 10 go.mod
> head: cannot open '10' for reading: No such file or directory
> ==> go.mod <==
> module github.com/kubevirt/ipam-extensions
> 
> go 1.26
> 
> toolchain go1.26.6
> 
> $ GOTOOLCHAIN=local go build cmd/manager/main.go
> $ echo $?
> 0
> ```

this pass but
if the go directive itself is 1.26.6, GOTOOLCHAIN=local, and go 1.26.5 is installed it will fail,
and we need to obey this, anyhow this PR is safe atm from that aspect
D/S uses GOTOOLCHAIN=local

having the go toolchain higher on upstream is the WA that allows upstream to diverge from D/S

ormergi:
> this pass but if the go directive itself is 1.26.6, GOTOOLCHAIN=local, and go 1.26.5 is installed it will fail, and we need to obey this, anyhow this PR is safe atm from that aspect D/S uses GOTOOLCHAIN=local
> 
> having the go toolchain higher on upstream is the WA that allows upstream to diverge from D/S

Yes this is the intention, set min version with go directive, set toolchain to use spesific Z version.
D/S product could either spsify the desired toolchain or use spesific Go version and set GOTOOLCHAIN=local.
We are on the same page.

go.mod
RamLavi: Prefer not to align them.

`go 1.26.0` is the minimum version (the hard floor, including for D/S with `GOTOOLCHAIN=local`). `toolchain go1.26.6` is the upstream build pin that `hack/install-go.sh` and the Dockerfile install.

RamLavi (approved):
/lgtm
/approve

oshoval:
https://github.com/k8snetworkplumbingwg/kubemacpool/blob/main/hack/go-version.sh

aint this one need to be changed to take toolchain line if exists [else fall back to go directive, if desired]

but maybe since it affects only the go mod tidy, then not sure,
but it is prone for mistakes, as now there are 2 variants all over,
also other repos / branches have different logic

ormergi:
@oshoval I checked it and it seems this PR changes doesnt break anything because the script that use go-version.sh look for major and minor versions only. 
The script in question is hack/check-go-version.sh, called by make target "check-go-version".

IMO, if anything should be done, I think we should simplify those checks to minimum and use native Go tooling for verifying Go version used and what the project support. 
Such as specifying Go's GOTOOLCHAIN var for anchoring the toolchain the project should work with.
And use 'go version' directly for fetching the used Go version.

PR #690, opened 2026-08-31
Align VMI MAC collision detection with allocated interfaces

RamLavi:

KubeMacPool does not allocate unmanaged pod-network interfaces
(l2bridge / bridge / passt on `pod: {}`), including primary UDN.
Collision detection is not L2-domain-aware: it treats MACs as
cluster-wide. As a result, Isolated PUDNs may reuse the same 
OVN IP-derived MAC (`0a:58` + IPv4) on different L2 domains; that 
is not a dataplane collision.

On a given CUDN/PUDN L2, OVN-K already owns uniqueness: IPAM for
auto-assigned MACs, and (on primary layer2 with preconfigured UDN
addresses) a per-network MAC registry that rejects duplicate user
MACs. KubeMacPool must stay out of that L2 rather than emit a
second, topology-agnostic alert.

This PR therefore aligns collision detection with allocation:
masquerade or Multus only. Status MACs are joined to spec by name,
so the VMI cache keeps spec networks and interfaces. Unmanaged
primary UDN MACs are not alerted on.

https://redhat.atlassian.net/browse/CNV-95971

- Unplug of the last managed NIC while the VMI stays Running is
  handled in a follow-up commit: that VMI is dropped from collision
  tracking because peers are not reconciled on that change.
- E2E uses `bridge` + `pod: {}` with a shared MAC as the unmanaged
  pod-network stand-in (`IsInterfaceSupported == false`). kubevirtci
  has no UDN CRDs, passt feature gate, or l2bridge binding plugin.

**Release note**:

```release-note
MAC collision detection now ignores interfaces KubeMacPool does not allocate, such as isolated primary UDN pod-network MACs.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MAC collision detection to consider only valid, supported network interfaces.
  * Prevented unmanaged, unsupported, unnamed, or invalid pod-network interfaces from triggering false collision alerts.
  * Continued detecting collisions on supported masquerade pod-network and Multus interfaces.
  * Cleared collision tracking when a virtual machine no longer has managed network interfaces.
  * Preserved relevant network and interface details during collision processing while excluding unrelated configuration.


RamLavi:
/retest

unrelated infra flake

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -73,7 +73,7 @@ func (p *PoolManager) AllocateVirtualMachineMac(virtualMachine *kubevirt.Virtual
 	copyVM := virtualMachine.DeepCopy()
 	newAllocations := map[string]string{}
 	for idx, iface := range copyVM.Spec.Template.Spec.Domain.Devices.Interfaces {
-		if iface.Masquerade == nil && networks[iface.Name].Multus == nil {
+		if !IsInterfaceSupported(iface, networks) {
 			logger.Info("mac address can be set only for interface of type masquerade and slirp on the pod network")
```
ormergi, line 77: The log message is outdated: slirp has been deprecated long time ago, and I think the message should reflect MAC can be assigned for Multus networks.

Here and in other occurrences.

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -331,7 +331,8 @@ func (p *PoolManager) initVirtualMachineMap() error {
 
 func (p *PoolManager) initMacMapFromCluster(parentLogger logr.Logger) error {
 	err := p.forEachManagedVmInterfaceInClusterRunFunction(func(vmFullName string, iface kubevirt.Interface, networks map[string]kubevirt.Network) error {
-		if !validateInterfaceSupported(iface, networks) {
+		if !IsInterfaceSupported(iface, networks) {
```
ormergi: please see https://github.com/k8snetworkplumbingwg/kubemacpool/pull/690/changes/02bf3ead8ae32a16fc1ccb06988437b79b65ff4f#r4004164889

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -599,12 +600,10 @@ func (p *PoolManager) getvmInstance(vmFullName string) (*kubevirt.VirtualMachine
...
 }
 
-func validateInterfaceSupported(iface kubevirt.Interface, networks map[string]kubevirt.Network) bool {
-	if iface.Masquerade == nil && networks[iface.Name].Multus == nil {
-		log.Info("mac address can be set only for interface of type masquerade and slirp on the pod network")
-		return false
-	}
-	return true
+// IsInterfaceSupported reports whether KubeMacPool allocates a MAC for the
+// interface: masquerade on the pod network, or any Multus attachment.
```
ormergi: nit: report may imply an info is published for consumers (e.g.:event, file,log), it merely return true/false

pkg/controller/maccollision/cache.go
```diff
@@ -39,15 +41,20 @@ const (
...
 	if !ok {
 		return obj, nil
 	}
 
-	// Create minimal VMI with only fields needed for collision detection
+	networks := make([]kubevirtv1.Network, len(vmi.Spec.Networks))
+	copy(networks, vmi.Spec.Networks)
+	interfaces := make([]kubevirtv1.Interface, len(vmi.Spec.Domain.Devices.Interfaces))
+	copy(interfaces, vmi.Spec.Domain.Devices.Interfaces)
+
```
ormergi, line 57: Why coping interface/network is needed?
Can we assign them as any other filed?
The reconciler should get a copy of the related VMI anyways, so it should be safe.

pkg/controller/maccollision/cache.go
```diff
@@ -66,30 +81,81 @@ func StripVMIForCollisionDetection(obj interface{}) (interface{}, error) {
 	return stripped, nil
 }
 
-// IndexVMIByMAC returns all MAC addresses from a VMI's status for indexing.
-// A VMI with multiple interfaces will be indexed under each MAC address.
-// This enables O(1) lookups of all VMIs that have a given MAC address.
+// IndexVMIByMAC returns MACs from a VMI that KubeMacPool would allocate.
```
ormergi: nit: this is no longer Index's it also filters, please rename the function to reflect that (maybe collectVMIMACs).

Please avoid specifying the network provider names (ovn-k), it can change in the future and imply the project is biased for spesific network providers.

So much for o(1) lookups?

pkg/controller/maccollision/cache.go
```diff
@@ -66,30 +81,81 @@ func StripVMIForCollisionDetection(obj interface{}) (interface{}, error) {
...
-	macs := []string{}
-	for _, iface := range vmi.Status.Interfaces {
-		if iface.MAC != "" {
-			normalizedMAC, err := NormalizeMacAddress(iface.MAC)
-			if err != nil {
-				cacheLog.Error(err, "failed to normalize MAC address", "mac", iface.MAC, "vmi", vmi.Name, "namespace", vmi.Namespace)
-				continue
-			}
-			macs = append(macs, normalizedMAC)
+	return managedMACsFromVMI(vmi)
```
ormergi: This function body doesnt make much sesne to me: 
The type check should be done at the caller so this function has less things to do.
Now, all it does is type check and call another helper.
Please consider inline managedMACsFromVMI(), or inline IndexVMIByMAC() - do the type check at the caller and call managedMACsFromVMI()

pkg/controller/maccollision/cache.go
```diff
@@ -66,30 +81,81 @@ func StripVMIForCollisionDetection(obj interface{}) (interface{}, error) {
...
+}
+
+// managedMACsFromVMI returns normalized MACs from status interfaces that join
+// to a spec interface KubeMacPool allocates for.
+func managedMACsFromVMI(vmi *kubevirtv1.VirtualMachineInstance) []string {
+	networks := networksByName(vmi.Spec.Networks)
```
ormergi: nit: specNetworks?

pkg/controller/maccollision/cache.go
```diff
@@ -66,30 +81,81 @@ func StripVMIForCollisionDetection(obj interface{}) (interface{}, error) {
...
+		if !ok {
+			continue
 		}
+		macs = append(macs, mac)
 	}
-
 	return macs
 }
 
+func networksByName(networks []kubevirtv1.Network) map[string]kubevirtv1.Network {
```
ormergi: nit: BTW the "indexed" naming fits here - indexSpecNetworksByName

pkg/controller/maccollision/cache.go
```diff
@@ -66,30 +81,81 @@ func StripVMIForCollisionDetection(obj interface{}) (interface{}, error) {
...
+}
+
+func specInterfacesByName(ifaces []kubevirtv1.Interface) map[string]kubevirtv1.Interface {
+	byName := make(map[string]kubevirtv1.Interface, len(ifaces))
+	for _, iface := range ifaces {
+		byName[iface.Name] = iface
+	}
+	return byName
+}
+
```
ormergi: nit:
Please consider inline these functions for simplification, all they do is index given slice, its fine to keep the logic at the caller, nothing special here.

Or convert these function to single generic function:
```
type goo struct{Name string; Value int}
	
s := []goo{
		{Name: "foo", Value: 1 },
		{Name: "boo", Value: 2 },
		{Name: "goo", Value: 3 },
}

inexedGooByName := foo(s, func(e goo) string { return e.Name })

...

func index[K string, V any](eles []V, keyFn func(ele V) K) map[K]V {
	indexed := make(map[K]V, len(eles))
	for _, e := range eles {
		indexed[keyFn(e)] = e
	}
	return indexed
}
```

pkg/controller/maccollision/cache_test.go
```diff
@@ -140,34 +170,88 @@ var _ = Describe("IndexVMIByMAC", func() {
...
 			},
 		}
 
 		macs := maccollision.IndexVMIByMAC(vmi)
-		Expect(macs).To(HaveLen(1))
-		Expect(macs).To(ContainElement("02:00:00:00:00:01"))
+		Expect(macs).To(BeEmpty())
+	})
+
+	It("should skip unnamed status interfaces", func() {
```
ormergi, line 248: Wouldn't that make the detector miss spoofed MACs? 
For example:
- Given VM A assigned with MAC A, VM B assigned with mac B.
- VM A stated with MAC A, later on the guest changes the MAC to B.
- VM A status present MAC B as unnamed interface entry.
- Traffic destined to VM B (MAC B) alternate between VM A (spoofing MAC B) and VM B (assigned with MAC B)

WDYT?

tests/vmi_test.go
```diff
@@ -179,6 +179,96 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 				})
 			})
 
+			Context("and VMIs use unmanaged pod-network interfaces", func() {
```
ormergi, line 182: Should we add test for covering a scenario when when the guest A changes the MAC to an already assigned one (by VM B)?

ormergi (commented):
First round 

- The terminology around an interface that should not raise collision is confusing, sometimes it referred to as "un-name interface", "unmanaged interface", "unspported interface" in various forms (GODOC, function names, etc..). 
Lets pick one and stick with it. I vote for "managed" naming.
- Some components (GODOC, tests, names) mentions project spesific items such as `l2bridge` of Kubevirt and P-UDN, UDN, CUDN of ON-Kubenretes, implying support for spesific projects. Where in practice it could work with others.
I think we should avoid mentioning project spesific naming, it could refered to as an example though.
- Please consider making unit test declarative as possible; so that the expected output is specified and asserted against the actual state. 
Instead of asserting spesific fields of a type (such as VMI). Well.. where possible.
Use ConsistOf instead of using Len+Contains matchers. Where possible.
- context.TODO is deprecated.

pkg/controller/maccollision/cache.go
RamLavi: The reconciler copy does not help here. This is a cache transform: if we assign those slices, the stored VMI aliases the informer’s backing array. 
We copy spec networks and interfaces so the cache has its own. Status already shares - we can copy that too if you want it consistent.

pkg/pool-manager/virtualmachine_pool.go
RamLavi: fixed (on my next push)

pkg/controller/maccollision/cache_test.go
RamLavi: OK so I verified what you said with the kubevirt/kubveirt code and you are right about the KubeVirt status shape.

If the guest rewrites the MAC on a Multus (or masquerade) NIC, libvirt keeps the allocated MAC on the named status entry. Guest-agent then reports the new MAC as a separate iface with empty Name (ifacesStatusFromGuestAgent: match by MAC, otherwise append guest-only). 
KubeVirt covers that in reports masquerade and bridge interfaces with their MAC changed in the guest.

We would not alert on that guest MAC. That is the same skip as other guest-only NICs (nested ovn-k8s-mp0, extra bridges, and so on). We cannot tell spoof from those, but that is not a regression from the behavior of kubemacpool, where we only relied on VM.Spec anyways.

This PR tracks MACs on interfaces KMP allocates, joined by spec name. Guest-rewritten MACs were never on the spec/domain NIC. Catching them would mean indexing unnamed guest-agent rows again, which is the false-positive we are fixing.

I would keep skipping unnamed. Spoofing belongs to port security / macspoofchk, not this alert.

tests/vmi_test.go
RamLavi: Same case as the unnamed guest-agent MAC [above](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/690#discussion_r4024049577). KubeVirt keeps the allocated MAC on the named status iface and puts the guest-rewritten one on a separate empty-Name entry. We skip that on purpose.

I would not add an e2e for it. It needs a guest agent and an in-guest MAC change, which this suite does not do, and it would lock in “do not alert on spoof” as a product test. The unit test already covers skipping unnamed status MACs. Spoofing is still port security / macspoofchk.

pkg/controller/maccollision/cache.go
ormergi: Thanks for explaining, make sense 

> Status already shares - we can copy that too if you want it consistent.

What do you mean by already share?

pkg/controller/maccollision/cache_test.go
ormergi: OK so no regression whatsoever as kubemacpool account for MAC it alocates only; MAC assinged on spec.interfaces
Thanks for explaining.

pkg/pool-manager/virtualmachine_pool.go
```diff
@@ -599,12 +600,10 @@ func (p *PoolManager) getvmInstance(vmFullName string) (*kubevirt.VirtualMachine
...
+func IsInterfaceSupported(iface kubevirt.Interface, networks map[string]kubevirt.Network) bool {
```
ormergi: nit: two items:
- GODOC: report may imply an info is published for consumers (e.g.:event, file,log), it merely return true/false
- name: how about `isManagedNetworkInterface`?

pkg/pool-manager/pool_test.go
```diff
@@ -124,6 +124,31 @@ var _ = Describe("Pool", func() {
 			Entry("Invalid address: FF:FF:00:00:00:00, the first octet is not 02, 06, 0A or 0E", "FF:FF:00:00:00:00", true),
 		)
 
+		DescribeTable("IsInterfaceSupported", func(iface kubevirt.Interface, networks map[string]kubevirt.Network, expected bool) {
```
ormergi: nit: you can pass the matcher (BeTrue/BeFalse) as is instead of the "expected" bool

pkg/controller/maccollision/cache.go
```diff
@@ -66,30 +81,81 @@ func StripVMIForCollisionDetection(obj interface{}) (interface{}, error) {
 	return stripped, nil
 }
 
-// IndexVMIByMAC returns all MAC addresses from a VMI's status for indexing.
-// A VMI with multiple interfaces will be indexed under each MAC address.
-// This enables O(1) lookups of all VMIs that have a given MAC address.
+// IndexVMIByMAC returns MACs from a VMI that KubeMacPool would allocate.
+// Unmanaged pod-network interfaces (for example OVN-generated MACs on primary
+// UDN) are omitted so they are not treated as cluster-wide collisions.
 func IndexVMIByMAC(obj client.Object) []string {
```
ormergi: nit: this function is no longer just indexs, now it also filters. Consider renaming to better reflect what it does, maybe `collectIMACs`.

Please avoid specifying the network provider names (ovn-k), it can change in the future and imply the project is biased for spesific network providers.

So much for o(1) lookups?

pkg/controller/maccollision/cache.go
```diff
@@ -66,30 +81,81 @@ func StripVMIForCollisionDetection(obj interface{}) (interface{}, error) {
...
-	for _, iface := range vmi.Status.Interfaces {
-		if iface.MAC != "" {
-			normalizedMAC, err := NormalizeMacAddress(iface.MAC)
-			if err != nil {
-				cacheLog.Error(err, "failed to normalize MAC address", "mac", iface.MAC, "vmi", vmi.Name, "namespace", vmi.Namespace)
-				continue
-			}
-			macs = append(macs, normalizedMAC)
+	return managedMACsFromVMI(vmi)
+}
```
ormergi: Following those changes this function body doesnt make much sense to me:
I think type checks should be done at the caller, leaving this function focus on one thing.
Now, what left for this function to do is  type check and call for another helper.
Please consider inline `managedMACsFromVMI()`.
Or inline `IndexVMIByMAC()`: perform the type check at the caller, and call `managedMACsFromVMI()` directly.

pkg/controller/maccollision/cache.go
ormergi: nit:
```suggestion
	specNetworks := networksByName(vmi.Spec.Networks)
```

pkg/controller/maccollision/cache.go
```diff
@@ -66,30 +81,81 @@ func StripVMIForCollisionDetection(obj interface{}) (interface{}, error) {
...
+	return byName
+}
+
+func specInterfacesByName(ifaces []kubevirtv1.Interface) map[string]kubevirtv1.Interface {
+	byName := make(map[string]kubevirtv1.Interface, len(ifaces))
+	for _, iface := range ifaces {
+		byName[iface.Name] = iface
+	}
+	return byName
+}
```
ormergi: nit:
Please consider inline these functions for simplification, all they do is index given slice, its fine to keep the logic at the caller, nothing special here.

Or convert these function to single generic function:
```
type goo struct{Name string; Value int}
	
s := []goo{
		{Name: "foo", Value: 1 },
		{Name: "boo", Value: 2 },
		{Name: "goo", Value: 3 },
}

inexedGooByName := foo(s, func(e goo) string { return e.Name })

...

func index[K string, V any](eles []V, keyFn func(ele V) K) map[K]V {
	indexed := make(map[K]V, len(eles))
	for _, e := range eles {
		indexed[keyFn(e)] = e
	}
	return indexed
}
```

pkg/controller/maccollision/cache_test.go
```diff
@@ -111,6 +111,40 @@ var _ = Describe("StripVMIForCollisionDetection", func() {
...
+
+		stripped := result.(*kubevirtv1.VirtualMachineInstance)
+		Expect(stripped.Spec.Hostname).To(BeEmpty())
+		Expect(stripped.Spec.Domain.Devices.Disks).To(BeEmpty())
+		Expect(stripped.Spec.Networks).To(HaveLen(2))
+		Expect(stripped.Spec.Networks[0].Name).To(Equal("pod"))
+		Expect(stripped.Spec.Networks[1].Multus.NetworkName).To(Equal("nad"))
+		Expect(stripped.Spec.Domain.Devices.Interfaces).To(HaveLen(2))
+		Expect(stripped.Spec.Domain.Devices.Interfaces[0].Binding.Name).To(Equal("l2bridge"))
+		Expect(stripped.Spec.Domain.Devices.Interfaces[1].Bridge).NotTo(BeNil())
```
ormergi: nit: Consider changing the asserts to declerative one, can be done by defining the expected VMI object.

pkg/controller/maccollision/cache_test.go
```diff
@@ -140,34 +170,88 @@ var _ = Describe("IndexVMIByMAC", func() {
...
 	It("should normalize MAC addresses to lowercase", func() {
+		vmi := vmiWithMasqueradeStatusIfaces([]kubevirtv1.VirtualMachineInstanceNetworkInterface{
+			{Name: "eth0", MAC: "02:00:00:00:00:AA"},
+			{Name: "eth1", MAC: "02:00:00:00:00:BB"},
+		})
+
+		macs := maccollision.IndexVMIByMAC(vmi)
+		Expect(macs).To(HaveLen(2))
+		Expect(macs).To(ContainElement("02:00:00:00:00:aa"))
+		Expect(macs).To(ContainElement("02:00:00:00:00:bb"))
```
ormergi: nit: Consider changing the assert to be mode declerative, by defining the expected MAC addresses slice and use the ConsistOf/Equal matcher.
Here and in other similar occurrences.

pkg/controller/maccollision/cache_test.go
ormergi: nit:
```suggestion
	It("should skip non-managed interfaces (ifaces who present in status but not in spec)", func() {
```

pkg/controller/maccollision/cache_test.go
```diff
@@ -140,34 +170,88 @@ var _ = Describe("IndexVMIByMAC", func() {
...
 
 		macs := maccollision.IndexVMIByMAC(vmi)
-		Expect(macs).To(HaveLen(2))
-		Expect(macs).To(ContainElement("02:00:00:00:00:aa"))
-		Expect(macs).To(ContainElement("02:00:00:00:00:bb"))
+		Expect(macs).To(ConsistOf("02:00:00:00:00:01"))
 	})
 
-	It("should skip interfaces without MAC addresses", func() {
+	It("should skip l2bridge binding on the pod network", func() {
```
ormergi: nit:
```suggestion
	It("should skip non masquerade binding ifaces attached to pod network", func() {
```

pkg/controller/maccollision/vmi_controller.go
```diff
@@ -247,18 +247,11 @@ func (r *VMIReconciler) filterVMIsWithMAC(ctx context.Context, normalizedMAC str
...
-				logger.Error(err, "failed to normalize MAC address", "mac", iface.MAC, "vmi", vmi.Name, "namespace", vmi.Namespace)
-				continue
-			}
-			macs[normalizedMAC] = struct{}{}
-		}
+	for _, mac := range managedMACsFromVMI(vmi) {
+		macs[mac] = struct{}{}
 	}
 	return macs
 }
```
ormergi: This method seem redundent, it does not r/w from the reciver object, also it duplicates `managedMACSet`.
In general I think such simple operations as indexing can be done at the caller, the prodcut variable can self explain what's being done.

pkg/controller/maccollision/vmi_controller_test.go
```diff
@@ -94,16 +94,98 @@ func withPhase(phase kubevirtv1.VirtualMachineInstancePhase) vmiOption {
...
+}
+
+func withUnnamedStatusMAC(mac string) vmiOption {
+	return func(vmi *kubevirtv1.VirtualMachineInstance) {
+		vmi.Status.Interfaces = append(vmi.Status.Interfaces, kubevirtv1.VirtualMachineInstanceNetworkInterface{
+			MAC: mac,
+		})
+	}
+}
+
```
ormergi: nit: Those helpers seem to be used later in this file and distracting the flow of reviewing. Consider moving them closer and below to where they used (similar to the common convention in the repo)

pkg/controller/maccollision/cache.go
RamLavi: > Thanks for explaining, make sense
> 
> > Status already shares - we can copy that too if you want it consistent.
> 
> What do you mean by already share?

I mean - not copy, but share the slice.

pkg/controller/maccollision/cache.go
```diff
@@ -66,30 +81,81 @@ func StripVMIForCollisionDetection(obj interface{}) (interface{}, error) {
...
 	return macs
 }
 
+func networksByName(networks []kubevirtv1.Network) map[string]kubevirtv1.Network {
+	byName := make(map[string]kubevirtv1.Network, len(networks))
+	for _, network := range networks {
+		byName[network.Name] = network
+	}
+	return byName
+}
```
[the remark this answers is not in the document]
RamLavi: Leaving them. specInterfaceNamedInStatus is the empty-Name skip (guest-agent unmatched NICs). networksByName / specInterfacesByName keep supportedMACsFromVMI readable. withUnnamedStatusMAC matches the other VMI test options. One caller is fine when the name is the documentation.

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/cc6f34c8fb626fdd4817b01b5bd5001874158046..b5923dd03bcec28a8e94096966c70421cc295a0e): address @ormergi's review:
- Commit 1 (pool-manager):
virtualmachine_pool.go: skip logs say masquerade-or-Multus (drop slirp)
pool_test.go: table entry title plugin binding on pod network

- Commit 2 (maccollision align):
cache.go / predicates.go / vmi_controller.go: managed* → supported* (filter unchanged)
cache_test.go: ConsistOf; unsupported / plugin-binding titles
predicates_test.go: supported/unsupported titles and helper names
vmi_controller_test.go: withPluginBindingPodNetworkMAC; unsupported titles
message: drop PUDN/CUDN/OVN-K registry; keep cache/join/filter bullets

- Commit 3 (clear tracking):
vmi_controller.go: log/comment supported (still removeVMIFromAllCollisions)
vmi_controller_test.go: unplug test title last supported NIC

- Commit 4 (e2e):
tests/vmi_test.go: unsupported wording; context.Background() on the new Creates
subject: unmanaged → unsupported

pkg/controller/maccollision/cache.go
ormergi: Whatever approach we take I think it should be followed for all fields in the stripped object should be treated the same.

From what I know controller runtime internals (cache, listWatch mechanism) account for working with a deep-copy of the subject object, so assigning slices/maps should be safe.
Seems like the concern about slices could be true for object pointers.

pkg/controller/maccollision/cache.go
ormergi, line 84: The current implementation is hard to follow due to the multiple addition helpers and how the logic is divided between them (also raised on https://github.com/k8snetworkplumbingwg/kubemacpool/pull/690/changes/0a6d7052ddf1a1304e0065a2846f9247b7d8634e#r3897916331)
Also it seems that the logic is pivot MAC collection by status interfaces.

Consider inverting the logic, so that spec interfaces/networks are validated against the status:
- Fetch spec interfaces/networks
- Filter managed interfaces/networks (multus, or masqudade ones)
- Fetch status interfaces, filter by managed spec interfaces
  I think is the only place require indexing (the spec ifaces)
- Collect the MACs
- Normalize the collected MACs

Keep simple operations as indexing/filtering inline, proper variable naming would avoid compromising readability. Or use generics.

I think this approach could simplify the implementation make it easier to follow, reducing helper quantity, indexing / filtering operations,

pkg/controller/maccollision/vmi_controller_test.go
```diff
@@ -94,16 +104,98 @@ func withPhase(phase kubevirtv1.VirtualMachineInstancePhase) vmiOption {
...
```
ormergi: nit: please move those function closer and below the place they are used.

pkg/controller/maccollision/vmi_controller.go
```diff
@@ -183,9 +183,12 @@ func (r *VMIReconciler) checkMACCollisions(ctx context.Context, vmi *kubevirtv1.
 		return nil
 	}
 
-	macs := r.extractMACsFromVMI(vmi, logger)
+	macs := r.extractMACsFromVMI(vmi)
```
ormergi, line 186: The method `extractMACsFromVMI` seem redundant: it does not r/w from the receiver object (can be defined as regular function), and duplicates `supportedMACSet` lives in the same package.

It seems supportedMACsFromVMI() can be used here, could you use it?
Or at least resue `supportedMACSet`?

pkg/controller/maccollision/predicates.go
```diff
@@ -60,23 +57,13 @@ func collisionRelevantChanges() predicate.TypedPredicate[*kubevirtv1.VirtualMach
...
 func macAddressesChanged(oldVMI, newVMI *kubevirtv1.VirtualMachineInstance) bool {
-	oldMACs := extractMACAddresses(oldVMI)
-	newMACs := extractMACAddresses(newVMI)
-
-	return !maps.Equal(oldMACs, newMACs)
+	return !maps.Equal(supportedMACSet(oldVMI), supportedMACSet(newVMI))
 }
 
-func extractMACAddresses(vmi *kubevirtv1.VirtualMachineInstance) map[string]struct{} {
+func supportedMACSet(vmi *kubevirtv1.VirtualMachineInstance) map[string]struct{} {
```
ormergi, line 63: Why this function is necessary?

It seems supportedMACsFromVMI() can be used directly.
Is it for filtering out duplicate addresses? 
If so it can also be done by supportedMACsFromVMI().

ormergi (changes requested):
Thanks, next round

pkg/controller/maccollision/cache.go
RamLavi: ah I now understand what you mean. 

I’ll remove the partial slice copies and treat the spec fields consistently with the status fields. Since this transform only _reads_ the retained fields, and controller-runtime’s underlying client-go cache explicitly permits a TransformFunc to mutate its input in place, then retaining these references is safe and avoids unnecessary work in the cache insertion path: [TransformFunc documentation](https://github.com/kubernetes/client-go/blob/v0.34.1/tools/cache/delta_fifo.go#L148-L163).

pkg/controller/maccollision/cache.go
RamLavi: Agreed. Eligibility is determined by the VMI spec, while status only provides the observed MAC. I’ll invert the collection flow to select KubeMacPool-managed spec interfaces first and then match them to status by name. That will also remove the extra helper layering and let the reconciler and predicate share the same collection function.

pkg/controller/maccollision/vmi_controller.go
RamLavi: Done as part of the collection refactor in commit above. I removed extractMACsFromVMI; the reconciler now builds its set directly from supportedMACsFromVMI.

pkg/controller/maccollision/predicates.go
RamLavi: Done as part of the same refactor in commit above. The predicate now compares sets constructed directly from supportedMACsFromVMI, so the separate set helper is gone.

RamLavi:
[Change](https://github.com/k8snetworkplumbingwg/kubemacpool/compare/b5923dd03bcec28a8e94096966c70421cc295a0e..b3ad1f6af36b8d56cb1c813ec81b6760a126c1e7): address 2nd round comments

@ormergi PTAL

